PR #214, opened 2021-09-30
Flush caches for domain verification

### LINKED ISSUE #154: The Site cannot be verified on the onboarding process
### Describe the bug:
Captura de pantalla de 2021-08-13 16-15-14

### Steps to reproduce:
1- Install and activate the plugin
2- do the onbarding
3- Login with the Pitnerest account on onboarding
4- Verify Site on the onboarding
Same steps with 0.9.0 works as expected

#### Expected behavior
Verify site works ok

#### Actual behavior
error message is showing

nima-karimi:
Closes #154

To make sure that caching doesn't cause issues with the domain verification, the WordPress rewrite rules are flushed whenever the rule is added. Since the `verification_rewrite` method is run on every page load (on `init`), I've added a simple check that fetches the WP rewrite rules from DB and makes sure that the new rule doesn't exist before flushing the rules. This still adds an extra DB call to EVERY page load, which isn't ideal.

There's also the option of removing the meta tag and the verification HTML from the site once the domain is claimed. This should be the preferred way of handling domain verification because once the domain is claimed users no longer need those tags and URLs on their site. However, it might not be that easy for the user to claim another domain without disconnecting the plugin and removing data.

As an extra measure, the `home_url` WordPress function is now used to obtain the domain because we've had issues with `site_url` in other extensions:

> We've seen problems with other integrations where it is preferable to use home_url() to cover situations where merchant sites might have a more complicated setup with wp files being located somewhere different to where the front end of the site is accesible. e.g. woocommerce/google-listings-and-ads#798

(from https://github.com/woocommerce/pinterest-for-woocommerce/issues/154#issuecomment-920977632)

This PR also adds `no-cache` headers when loading the **pinterest-XXXXX.html** file. I originally wanted to add the no-cache headers to the site whenever the domain verification meta tag is added but then I realized it's always added so that's not an option. If we decide to hide the meta tag once the domain gets claimed then we can add the no-cache headers temporarily until then. Another option would be to check for the `user-agent` of the Pinterest bot that checks for verification meta tag and only reset cache if it's a match. However, I'm not sure if `user-agent` or any other identifying information is provided by Pinterest when they check the user's website for the verification meta tag.

1. Connect your Pinterest account and claim your domain

### DIFF
--- a/class-pinterest-for-woocommerce.php
+++ b/class-pinterest-for-woocommerce.php
@@ -710,7 +710,14 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 
 			if ( $filename ) {
 				$escaped = preg_quote( $filename, '/' );
-				add_rewrite_rule( '^' . $escaped . '$', 'index.php?pinterest_verification=true', 'top' );
+				$rewrite_regex = "^{$escaped}$";
+				add_rewrite_rule( $rewrite_regex, 'index.php?pinterest_verification=true', 'top' );
+
+				// Flush rewrite rules as we need to serve the URL for the pinterest-XXXXX.html file.
+				$rewrite_rules = get_option( 'rewrite_rules' );
+				if ( ! is_array( $rewrite_rules ) || ! array_key_exists( $rewrite_regex, $rewrite_rules ) ) {
+					flush_rewrite_rules();
+				}
 			}
 		}
 
--- a/class-pinterest-for-woocommerce.php
+++ b/class-pinterest-for-woocommerce.php
@@ -773,6 +773,7 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 				$verification_data = self::get_data( 'verification_data' );
 				$verification_code = $verification_data['verification_code'];
 
+				wc_nocache_headers();
 				header( 'Content-Type: text/html' );
 				echo '<!DOCTYPE html><html lang="en" xmlns="http://www.w3.org/1999/xhtml"><head><meta name="p:domain_verify" content="' . esc_attr( $verification_code ) . '"/><title></title></head><body>' . esc_html__( 'Pinterest for WooCommerce verification page', 'pinterest-for-woocommerce' ) . '</body></html>';
 				exit;
--- a/class-pinterest-for-woocommerce.php
+++ b/class-pinterest-for-woocommerce.php
@@ -236,10 +236,8 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 			add_action( 'pinterest_for_woocommerce_token_saved', array( $this, 'update_account_data' ) );
 			add_action( 'pinterest_for_woocommerce_token_saved', array( $this, 'set_default_settings' ) );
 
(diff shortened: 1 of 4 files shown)


nima-karimi:
Sorry, @jconroy! Assigning this to you since I'm not sure who I can ask to review this.

nima-karimi:
@ChristianAF Can you please check if this resolves the issue you mentioned in #154? I wasn't able to reproduce it but as @jconroy mentioned, it could be due to cahcing.

findingsimple:
No worries @nima-karimi , I'll just request a review from all of Ventures - as someone will no doubt be able to get to it before me

JPry:
@nima-karimi I'll take a look at this today

JPry:
@nima-karimi I just wanted to address this sentence to set your mind at ease: 

> This still adds an extra DB call to EVERY page load, which isn't ideal.

Fortunately, this isn't actually an *extra* call to the DB. WordPress will load the rewrite rules on every page load anyway, which can be seen in the `WP_Rewrite::wp_rewrite_rules()` method. That method is called during `WP::parse_request()`, and the `get_option()` function will cache what it pulls from the DB during the request in order to avoid fetching the same data twice.

JPry:
@nima-karimi Upon looking into this a bit and doing some testing, I don't think we even need to use rewrites at all. We're not ever going to have a variable structure for the domain verification request, and having some kind of variation is what necessitates the need for rewrites.

We're already hooking into the `parse_request` action with the `Pinterest_For_Woocommerce::verification_request()` method. We could adjust this method to just look at the incoming request directly without the need for rewrite rules and query variables. We could then entirely remove the `verification_rewrite()` and `verification_query_var()` methods, and the `verification_request()` method could be updated to look like this:

```php
		/**
		 * Hook the parse_request action and serve the html
		 *
		 * @param WP $wp Current WordPress environment instance.
		 */
		public function verification_request( $wp ) {
			$verification_data = self::get_data( 'verification_data' );
			if ( ! $verification_data || ! array_key_exists( 'filename', $verification_data ) ) {
				return;
			}

			// phpcs:ignore WordPress.Security.ValidatedSanitizedInput
			$request = trim( $wp->request ?? $_SERVER['PHP_SELF'] ?? '', '/' );
			if ( $verification_data['filename'] === $request ) {
				wc_nocache_headers();
				header( 'Content-Type: text/html' );
				?>
<!DOCTYPE html>
<html lang="en" xmlns="http://www.w3.org/1999/xhtml">
<head>
	<meta name="p:domain_verify" content="<?php echo esc_attr( $verification_data['verification_code'] ); ?>"/>
	<title></title>
</head>
<body><?php esc_html_e( 'Pinterest for WooCommerce verification page', 'pinterest-for-woocommerce' ); ?></body>
</html>
				<?php
				exit;
			}
		}
```

I'll commit what I did here and push it up so you can take a look at it directly. I think this gives us what we need for the domain verification without needing to worry about rewrites and whether they are updated or not.

nima-karimi:
Nice 💯 Works perfectly. Thanks, @JPry.

JPry (approved):

PR #216, opened 2021-10-01
Reduce minimum PHP version requirement to 7.3

### LINKED ISSUE #215: Reduce min php requirements to php 7.3
Decision internally to reduce the min requirements to php 7.3

tl;dr; was even though [php 7.3 stops being supported soon](https://www.php.net/supported-versions.php) there is still a decent % of users running 7.3 and reducing the requirement should help reduce any friction witj onboarding and adoption

nima-karimi:
Closes #215 

This PR reduces the minimum required PHP version to 7.3 (from 7.4). I couldn't find any incompatible code so most of the changes in this PR are basically replacing the 7.4 texts with 7.3. I also added a new `config > platform` field to `composer.json` so that it mocks the platform's PHP version and no incompatible packages can be installed even if the host PHP version is greater than 7.3.

Nothing specific to test here so I think it'd be good to just configure your local to use PHP 7.3 and do some basic smoke testing.

### Changelog note

> Dev: Reduce minimum PHP version requirement to 7.3

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,6 +11,7 @@
 	"prefer-stable": true,
 	"minimum-stability": "dev",
 	"require": {
+		"php": ">=7.3",
 		"composer/installers": "^1.7.0",
 		"automattic/jetpack-autoloader": "^2.10.1",
 		"defuse/php-encryption": "^2.2"
@@ -21,6 +22,11 @@
 	"autoload": {
 		"psr-4": {
 			"Automattic\\WooCommerce\\Pinterest\\": "src/"
+		}
+	},
+	"config": {
+		"platform": {
+			"php": "7.3.0"
 		}
 	},
 	"scripts": {
--- a/.github/disabled-workflows/php-coding-standards.yml
+++ b/.github/disabled-workflows/php-coding-standards.yml
@@ -45,7 +45,7 @@ jobs:
       - name: Set up PHP
         uses: shivammathur/setup-php@v2
         with:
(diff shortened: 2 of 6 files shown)


composer.json
```diff
@@ -23,6 +24,11 @@
 			"Automattic\\WooCommerce\\Pinterest\\": "src/"
 		}
 	},
+	"config": {
+		"platform": {
+			"php": "7.3.0"
```
JPry, line 29: In Google Listings and Ads, we have the latest version of PHP `7.3.27` as this value, rather than `7.3.0`. I don't know how much of a difference it makes in practice, but I *think* that we might want to specify a higher version for this value.

Do you have any more information about which option is better, if any?

JPry (approved):
Looks good overall, I just had one question about a line in the `composer.json` file. Other than that this is ready 🚀

composer.json
nima-karimi: That shouldn't make much of a difference in practice since most of the packages compatible with 7.3.27 are also compatible with 7.3.0. I think it's best to specify the minimum version that we support though, which is 7.3.0. Just for the sake of consistency with our docs. Because we don't mention the minor patch version anywhere.

nima-karimi: This config just simulates the PHP version when installing or upgrading packages so when for example a dev has PHP 8 installed and tries to add or update a package, Composer still checks for compatibility with PHP 7.3 when installing the package.

PR #217, opened 2021-10-05
Cancel ActionScheduler jobs on account disconnection

### LINKED ISSUE #166: Actions should be removed/cancelled on disconnect, deactivation & catalog disable
### Describe the bug:
There's a scheduled action (`WooCommerce > Status > Scheduled Actions`) for registering and generating the feed file (and maybe other feed stuff).

If the merchant disconnects from Pinterest, all feed-related code should stop (as I understand it). However, the action continues to run:

Screen Shot 2021-08-19 at 10 01 49 AM

Note also there are two actions in the table. Is this intended? Seems like a bug.

### Steps to reproduce:
1. Onboard and enable product sync.
2. View the `Pending` actions in `WooCommerce > Status > Scheduled Actions` and search for `pinterest`.
3. Disconnect from Pinterest.

#### Expected behavior
- The actions should be removed/cancelled if user does any of the following:
  - disconnects from Pinterest
  - deactivates or uninstalls Pinterest plugin
  - disables catalog sync.
- Ideally only one scheduled/recurring action. If two are needed, give them different hook names and make it clear how they relate.

In general, if catalog sync is not active, no code relating to it should run.

#### Actual behavior
- 2x recurring actions that look the same.
- Actions continue to run when user has disconnected their store from Pinterest.

nima-karimi:
Related to #166 

This PR cancels the feed generation and product sync jobs when the user disconnects their Pinterest account.

1. Connect your account and wait for the action scheduler jobs to be scheduled (by checking in `WooCommerce > Status > Action Scheduler`).
2. Disconnect your Pinterest account and make sure the jobs are canceled.

### DIFF
--- a/class-pinterest-for-woocommerce.php
+++ b/class-pinterest-for-woocommerce.php
@@ -578,6 +578,9 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 			self::save_setting( 'tracking_advertiser', null );
 			self::save_setting( 'tracking_tag', null );
 
+			// Cancel scheduled jobs
+			Pinterest\ProductSync::cancel_jobs();
+
 			// At this point we're disconnected.
 			return true;
 		}
--- a/src/ProductSync.php
+++ b/src/ProductSync.php
@@ -811,4 +811,15 @@ class ProductSync {
 			self::feed_reschedule();
 		}
 	}
+
+
+	/**
+	 * Cancels the scheduled product sync jobs.
+	 *
+	 * @return void
+	 */
+	public static function cancel_jobs() {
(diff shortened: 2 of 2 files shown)


ksere:
While this cleanup is needed, #166 is probably related to double schedulling occurring for other reasons, and should be fixed by #201

nima-karimi:
Hmm. That's probably why I couldn't reproduce it creating 2x or 3x scheduled jobs.

findingsimple (approved):
LGTM

findingsimple:
Added https://github.com/woocommerce/pinterest-for-woocommerce/commit/bc5ccbdc0e03c30e08763859ccdfc33ac6afd06f

PR #220, opened 2021-10-06
Add phpcs to composer dependencies

findingsimple:
`npm run lint:php` wasn't working for me (I'm guessing because I don't have phpcs installed globally).

This PR adds phpcs as a composer dev dependency.


JPry (approved):
LGTM 🚀

PR #224, opened 2021-10-13
Fix title breadcrumbs in WC 5.8

tomalec:

- Use ~`@woocommerce/wc-admin-settings`~ `@woocommerce/settings` imported via DEWP,
   instead of global `window.wcSettings` or settings store `getSetting`.
   Improves WC5.8 compatibility.
	  
   Note, that the package is not published and not added to dependencies.

   Works around https://github.com/woocommerce/woocommerce-admin/issues/7781
- It also updates DEWP to the latest version, it is not required for the fix.

1. Checkout the branch
2. `npm install`
3. `npm run build`
4. Open any Pinterest page, like [`/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Fonboarding`](https://gla1.test/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Fonboarding)
5. Check the page title, `document.title` in console

You should see 
```
'Onboarding Guide ‹ Pinterest ‹ Marketing ‹ WooCommerce ‹ gla1.test title — WooCommerce'
```
instead of 
```
'Onboarding Guide ‹ Pinterest ‹ Marketing ‹ ‹ gla1.test title — WooCommerce'
'Onboarding Guide ‹ Pinterest ‹ Marketing ‹ false ‹ gla1.test title — WooCommerce'
```

### Changelog note

> Fix|New|Dev: Add suggested changelog note here

### Todo
- [x] figure out why sometimes it returns `false`
- [x] Check if webpack config change is needed, if we update the WooDEWP 

### Additional notes

This PR introduces a dependency that is not mentioned in `package.json` which may break local unit tests in the future, but at least it somewhat indicates where the `wcSettings` global comes from. Still, behind DEWP mystery, but at least "somewhat".

### DIFF
--- a/assets/source/setup-guide/index.js
+++ b/assets/source/setup-guide/index.js
@@ -3,6 +3,9 @@
  */
 import { __ } from '@wordpress/i18n';
 import { addFilter } from '@wordpress/hooks';
+import { getSetting } from '@woocommerce/wc-admin-settings'; // eslint-disable-line import/no-unresolved
+// The above is an unpublished package, delivered with WC, we use Dependency Extraction Webpack Plugin to import it.
+// See https://github.com/woocommerce/woocommerce-admin/issues/7781
 
 /**
  * Internal dependencies
@@ -21,7 +24,7 @@ addFilter(
 	( pages ) => {
 		const navigationEnabled = !! window.wcAdminFeatures?.navigation;
 		const initialBreadcrumbs = [
-			[ '', wcSettings.woocommerceTranslation ],
+			[ '', getSetting( 'woocommerceTranslation' ) ],
 		];
 
 		/**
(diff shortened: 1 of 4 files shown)


findingsimple:
Thanks @tomalec !

tomalec:
The WC 5.6-5.8 compatible way to obtain `woocommerceTranslation` way is 
```js
const woocommerceTranslation = wc.wcSettings.getSetting('woocommerceTranslation') || wc.wcSettings.getSetting('admin')?.woocommerceTranslation;
```


It seems that `wc-settings` given by DEWP come from a different package than we anticipated in the beginning.

It's not [`wc-admin-settings`](https://github.com/woocommerce/woocommerce-admin/tree/a4ddc2a8cb8c204b9a7fc4f77029dc676a56cc31/packages/wc-admin-settings), but `@woocommerce/settings` which may or may not be https://www.npmjs.com/package/@woocommerce/settings it would require deeper digging as the links there are broken.

findingsimple:
@eason9487 would you mind giving this one a review too please?

eason9487 (approved):
Tested well. LGTM.

PR #226, opened 2021-10-18
Add support for WooRelease

JPry:
Closes #219

### Changes proposed in this pull request

This sets up the repository to be able to be released with the Woorelease tool.

### Detailed test instructions
1. Run the Woorelease script using this branch:
    ```
    wr simulate --product_version=1.0.0 https://github.com/woocommerce/pinterest-for-woocommerce/tree/feature/woorelease
    ```
2. The script should not produce any errors

### Changelog note

> Dev: Add support for releasing with Woorelease

PR #229, opened 2021-10-21
Add WPORG assets

findingsimple:
Adds the WPORG plugin assets/images and some additional text to the README.txt for the WPORG plugin page to be consistent with the WCCOM plugin page.

@JPry with GLA we use a ` .gitattributes` to exclude folders/files from the build but we don't have that here.  I'll leave that for you to address in https://github.com/woocommerce/pinterest-for-woocommerce/pull/226

Plugins assets and content from pcTzPy-8s-p2


JPry:
> with GLA we use a `.gitattributes` to exclude folders/files from the build but we don't have that here. I'll leave that for you to address in #226

This should already be covered. The build process uses `gulp`, and one of the changes in #226 was explicitly specifying the files to include rather than specifying the files to exclude.

PR #233, opened 2021-10-27
Fix npm vulnerabilities

tomalec:
Our packages deliver `132 vulnerabilities (2 low, 76 moderate, 50 high, 4 critical)`
and a number of packages are pretty outdated
image

This PR aims to address at least a part of it.

 Our lowest WordPress 5.6 theoretically supports https://github.com/WordPress/wordpress-develop/blob/5.6/package.json#L80 so we cannot bump our `@wordpress/` dependencies.
WooCommerce 5.3.2, uses WC-admin `2.2.7`. Now we need some excavation to find out the individual packages. But it seems we use already higher ones. Maybe we could consider bumping the lowest WC to L-2 = 5.6 now?

So we only can bump devDependencies (and probably not all)

I'm not a user of gulp flows, so I skipped it as well.
Personally, I'd replace `underscore` with native modern ES, but that's something for a bigger PR.

Bumping es/css lints would require adapting to new rules, so it's another candidate for a separate PR.

- Run `npm audit fix` (562d5a0)
- Add `repository` field to `package.json`, (1605d76)
to suppress warning during `npm install`.

##### Before
`npm install`
132 vulnerabilities (2 low, 76 moderate, 50 high, 4 critical)
`npm outdated`
image
##### After
found 60 vulnerabilities (1 low, 11 moderate, 44 high, 4 critical)

1. `npm install`

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,10 +4,10 @@
   "description": "Pinterest for WooCommerce",
   "version": "1.0.0",
   "main": "gulpfile.js",
-	"repository": {
-		"type": "git",
-		"url": "git@github.com:woocommerce/pinterest-for-woocommerce.git"
-	},
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:woocommerce/pinterest-for-woocommerce.git"
+  },
   "dependencies": {
     "@woocommerce/components": "^5.1.2",
     "@woocommerce/data": "^1.1.1",
--- a/package.json
+++ b/package.json
@@ -4,6 +4,10 @@
   "description": "Pinterest for WooCommerce",
   "version": "1.0.0",
   "main": "gulpfile.js",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -4,6 +4,10 @@
   "description": "Pinterest for WooCommerce",
   "version": "1.0.0",
   "main": "gulpfile.js",
+	"repository": {
+		"type": "git",
+		"url": "git@github.com:woocommerce/pinterest-for-woocommerce.git"
+	},
```
eason9487: 💅 
Mixed tab and space indentations.

eason9487 (approved):
LGTM and approving in advance. Left a comment for a minor issue that has mixed indentations.

PR #240, opened 2021-11-04
Feature/improve feedstate logic

### LINKED ISSUE #166: Actions should be removed/cancelled on disconnect, deactivation & catalog disable

### LINKED ISSUE #178: Do not store progress as translated strings in the database
While working on #158, it became relevant that there's strings that are stored in the database to mark things like progress. These are stored translated (and after changes implemented in #163) with HTML markup even.

One particular example is the progress status for the feed generation (Wrote %s out of %s to file). 

I'm opening up this issue by request of @haszari to reinforce this:

- Markup / UI text should be generated as late as possible. 
- State should not be cached unless there's a specific need. If stats/data needs to be stored  for later display in UI (e.g. num products in feed so far, feed generation time, feed file size), store in transients so it's clear it's derived, temporary data. Store raw values, not generated markup. Prefer one value per transient, rather than storing a large nested structure.

ksere:
Closes #178 
Closes #166 

Small-ish refactor of the `ProductSync` methods related to handling the status of product feed.
More specifically;

- Improves logic of how we set & retrieve the status of the product feed by refactoring `ProductSync::feed_job_status()` onto smaller get/set methods. 
- Simplifies/reduces stored data for feed's current status.
- Uses transients for saving the related data.
- Avoids storing progress as string in the database.  
- Moves deregistration methods to concurrent request
- Avoid unnecessary calls to reschedule.
- Re-implements cleaner `reschedule_if_expired()`
- Implements `reschedule_if_errored()` to restart job after a delay.
- Remove scheduled actions on plugin deactivation. 

#### Screenshots

No visible changes to user flows / interactions. 

1. Start with a clean plugin installation and a new Pinterest account.
2. Start the onboarding process and Connect to Pinterest.
3. Complete onboarding process
4. Verify that the product feed is being generated properly by visiting the Catalog tab of the plugin.
5. Verify that the product feed is being registered with Pinterest properly using Pinterest's dashboard.
5. Verify that the product feed is being regenerated after it expires (24h)
6. Verify that progress is not stored as translated strings as stated in #178
6. Verify that when Product Sync is disabled, the feed file is being deleted, the related transients are being removed and the scheduled actions are being removed. 
7. Verify that when the plugin is deactivated, scheduled tasks are removed.

### DIFF
--- a/src/API/FeedState.php
+++ b/src/API/FeedState.php
@@ -11,7 +11,6 @@ namespace Automattic\WooCommerce\Pinterest\API;
 use Automattic\WooCommerce\Pinterest as Pinterest;
 
 use \WP_REST_Server;
-use \WP_REST_Request;
 
 if ( ! defined( 'ABSPATH' ) ) {
 	exit;
--- a/src/API/FeedState.php
+++ b/src/API/FeedState.php
@@ -164,7 +164,16 @@ class FeedState extends VendorAPI {
 	 */
 	private function add_local_feed_state( $result ) {
 
-		$state      = Pinterest\ProductSync::feed_job_status() ?? array( 'status' => 'pending_config' );
+		$local_feed = Pinterest\ProductFeedStatus::get_local_feed();
+		$state      = Pinterest\ProductFeedStatus::get(
+			array(
+				'status',
+				'current_index',
+				'last_activity',
+				'error_message',
+				'product_count',
+			)
+		);
 		$extra_info = '';
 
 		switch ( $state['status'] ) {
@@ -178,9 +187,12 @@ class FeedState extends VendorAPI {
 				$status_label = esc_html__( 'Feed generation in progress.', 'pinterest-for-woocommerce' );
 				$extra_info   = sprintf(
 					/* Translators: %1$s Time string, %2$s text string indicating progress */
-					esc_html__( 'Last activity: %1$s ago - %2$s', 'pinterest-for-woocommerce' ),
+					esc_html__( 'Last activity: %1$s ago - Wrote %2$s out of %3$s products to %4$sfeed file%5$s.', 'pinterest-for-woocommerce' ),
 					human_time_diff( $state['last_activity'] ),
-					$state['progress']
+					$state['current_index'],
+					$state['product_count'],
+					'<a href="' . $local_feed['feed_url'] . '" target="_blank">',
+					'</a>',
 				);
 				break;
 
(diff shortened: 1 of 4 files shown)


src/API/FeedState.php
```diff
@@ -189,9 +192,12 @@ private function add_local_feed_state( $result ) {
 				$status_label = esc_html__( 'Up to date', 'pinterest-for-woocommerce' );
 				$extra_info   = sprintf(
 					/* Translators: %1$s Time string, %2$s text string indicating progress */
```
msaggiorato: Will need to update translators string here.

src/API/FeedState.php
```diff
@@ -178,9 +178,12 @@ private function add_local_feed_state( $result ) {
 				$status_label = esc_html__( 'Feed generation in progress.', 'pinterest-for-woocommerce' );
 				$extra_info   = sprintf(
 					/* Translators: %1$s Time string, %2$s text string indicating progress */
```
msaggiorato: Will need to update translators string here.

src/API/FeedState.php
```diff
@@ -212,7 +218,7 @@ private function add_local_feed_state( $result ) {
...
```
msaggiorato: Will need to update translators string here.

msaggiorato (changes requested):
I'd say this is good to go to QA, just need to update some translateable strings.

budzanowski:
This looks good to go.

budzanowski (approved):

msaggiorato (approved):

PR #246, opened 2021-11-09
Enable enhanced match support by default 

### LINKED ISSUE #191: Enhanced Match is not enabled by default
### Describe the bug:

After installing the plugin and going through the setup onboarding flow in a brand new site, I go to the settings page and I noticed that the Enhanced Match support is not enabled by default.

image

I think it would be better to have it enabled by default (a sensible default). If the merchants do not want it, they can disable it in settings after the onboarding flow. This is also the same as what @garymurray mentioned in a Slack message some time ago.

I am not sure if there is a downside to have it enabled by default. I think, if I am a merchant and I want to track conversion (which is already enabled by default), I would want to have Enhanced Match too.

### Steps to reproduce:

1. Install the plugin on a brand new site (so that there isn't any previous old Pinterest plugin data).
2. Go through the setup onboarding flow.
3. Go to the settings page.

#### Expected behavior

Enhanced Match is enabled by default just like other settings.

#### Actual behavior

Enhanced Match is not enabled by default, I have to check it and click on the "Save changes" button.

ianlin:
Closes #191.

This PR changes the default value of `enhanced_match_support` to `true` in the default settings so the user who installs the plugin for the first time will see the option `enhanced_match_support` is checked.

#### Screenshots

Screenshot 2021-11-09 at 09 58 26

1. Install the plugin on a brand new site (so that there isn't any previous old Pinterest plugin data).
2. Go through the setup onboarding flow.
3. Go to the settings page.

### Changelog entry

> Fix - Enable enhanced match by default

### DIFF
--- a/class-pinterest-for-woocommerce.php
+++ b/class-pinterest-for-woocommerce.php
@@ -95,7 +95,7 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 		 */
 		protected static $default_settings = array(
 			'track_conversions'      => true,
-			'enhanced_match_support' => false,
+			'enhanced_match_support' => true,
 			'save_to_pinterest'      => true,
 			'rich_pins_on_posts'     => true,
 			'rich_pins_on_products'  => true,


ecgan (approved):
Tested, working as expected, LGTM 👍

Issue #232, opened 2021-10-27
Implement new tracking events


tomalec:
There is a list of WooCommerce Tracking events we would like to trigger in the plugin, to be able to analyze how the plugin is being used.
A comprehensive list of all events is available at https://docs.google.com/spreadsheets/d/1u25mOcsODgcdPMy_rIagUAWHCfzsvJI1e8UPrJvYtVE/edit#gid=903716843

### Describe the solution you'd like 

- [x] Ensure all the dependencies and libraries needed are in place and in up-to-date versions.
- [ ] Implement individual events (each should be prefixed with `wcadmin_pfw_`): 
   - [x] `page_view`	n/a		When viewing wc react powered page. Default wc-admin event.
   - [x] ~`disconnected_account_link_click`~ `account_disconnect_button_click { context: 'settings' }` (see [below](https://github.com/woocommerce/pinterest-for-woocommerce/issues/232#issuecomment-1002184777))			When acocunt is disconnected rom the setting page https://d.pr/i/s7Oh7T  https://github.com/woocommerce/pinterest-for-woocommerce/pull/313
   - [x] `documentation_link_click`			When clicking documentation link e.g. https://d.pr/i/nSI71V, https://d.pr/i/COCw3G  and https://d.pr/i/plH2LQ 	https://github.com/woocommerce/pinterest-for-woocommerce/pull/261		
   - [x] `get_started_faq`			Clicking on getting started page faq item to collapse or expand it https://d.pr/i/CP6WPF - https://github.com/woocommerce/pinterest-for-woocommerce/pull/237
   - [x] `get_started_notice_link_click`			Clicking on a text link within the notice on the Get Started page https://d.pr/i/NWRC2W  https://github.com/woocommerce/pinterest-for-woocommerce/pull/299
   - [ ] `help_click`		**Not needed yet**	Triggered when the "Help" button is clicked.
   - [x] `account_connect_button_click`			https://d.pr/i/E3rtBZ - https://github.com/woocommerce/pinterest-for-woocommerce/pull/239
   - [x] `account_create_button_click`			https://d.pr/i/E3rtBZ  - https://github.com/woocommerce/pinterest-for-woocommerce/pull/239
   - [x] `modal_closed`			Triggered when a modal is closed https://github.com/woocommerce/pinterest-for-woocommerce/pull/290
   - ~[ ] `modal_content_link_click`			Triggered when clicking on a text link within the modal content~ currently there is no link in the modal.
[...]

PR #254, opened 2021-11-12
Tweak Tracks docs generation,

tomalec:
related to #232 

- sort events by name
- add source file lineno to the link
- add a line break after the end disclaimer
- support empty descriptions
- document links of Track event emitters
- added a plugin, to handle `@fires` descriptions better - support `.:,#\n`

0. Do some changes in `@events` descriptions
1. `npm run doc:tracking`
2. Review the `TRACKING.md`


woo-tracking-jsdoc/publish.js
```diff
@@ -40,34 +58,64 @@ exports.publish = function ( data ) {
...
+					);
+					const description = property.description.replace(
+						/\|/g,
+						'\\|'
+					);
+					mdResult += `\`${ property.name }\` | \`${ type }\` | ${ description }\n`;
+				} );
+			}
+
+			// Find all palces that fires the event.
```
eason9487: 💅 
Seems a typo?
```suggestion
			// Find all places that fires the event.
```

woo-tracking-jsdoc/publish.js
```diff
@@ -40,34 +58,64 @@ exports.publish = function ( data ) {
...
+					fires.startsWith( 'event:' + symbol.name )
+				)
+				if ( firesCurrent.length ) {
+					emitters.set(
+						emitter,
+						// Consider everything after the event name [and whitespace] an additional description.
+						// Sanitize the descriptions, trim empty ones.
+						firesCurrent.map( description => description.match( /^(\S+\s*)(.*)/ )[ 2 ])
+					);
+				}
```
eason9487: ❓ 
I'm trying to understand what formatting would appear as a multi-line description, but it seems that `#` and `.` symbols for `@fires` are part of its format. If the partial formatting is borrowed as a description, there may be more than one of those symbols in the description, and then resulting in unintended parsing results. For examples:
- Multiple sentences in a single line. Or, if there is anything after the first `.`, it will have the same result
    ```
    * @fires wcadmin_pfw_get_started_faq whenever the FAQ is toggled. The second sentence.
    ```
- Multi-line descriptions
    ```
    * @fires wcadmin_pfw_get_started_faq whenever the FAQ is toggled.
        The second sentence.
    ```
- Using `#`. (Looks like it follows the same parsing rule of `.`)
    ```
    * @fires wcadmin_pfw_get_started_faq whenever the #FAQ is toggled
    ```

These usages will cause different formats of strings `emitter.fires` and mismatch the `filter()` in line 91.

eason9487 (commented):
💅 
This file is not included in the check scope of JS lint. But I think it could still be adjusted the format to avoid distracted editor errors?

image

woo-tracking-jsdoc/publish.js
tomalec: Sorry, I should've mentioned that.
- To use `.` and `:` you need to use the full name path for the event, like:
   ```
   * @fires event:wcadmin_pfw_get_started_faq whenever the FAQ is toggled. The second sentence `{ foo: bar, baz: fiz}`.
   ```
   - Currently, this breaks `lint:js` as we use an outdated version of `eslint-plugin-jsdoc` without this fix https://github.com/gajus/eslint-plugin-jsdoc/issues/750. I was thinking of installing the updated version of `eslint-plugin-jsdoc` once we have the use case for that, as I hope `woocommerce/eslint…` may catch up in the meantime.
- I missed multiline support. I can add it by updating the regexp to `description.match( /^(\S+\s*)(.*)/ )[ 2 ]` - but it again has the `event:` problem  with dots in following lines. - I guess it's just a thing in `jsdoc` parser, as it really does not match catch the description, just considers everything a name :(	I may try to cover it by PR to jsdoc or separate JSDoc plugin, but that would definitely be a separate PR :scroll: 
- `#` and `,` are even more problematic, as I have found no easy workaround.

I think in this PR I may just fix the multiline support. Then, work on better parsing of `#,.:` in fires descriptions, and actually parsing first desriptions in a separate PR.

tomalec:
@eason9487 Thanks for the review, would you mind another round?

woo-tracking-jsdoc/publish.js
tomalec: I created a new plugin in https://github.com/woocommerce/pinterest-for-woocommerce/pull/254/commits/d42f3c52e4e1c34213da143acb5ae4519eabfcb1

Thanks to it, 
 - there is no need to use the full event name (which was incompatible with our bugged version of eslint)
 - we can use `.;#` in descriptions
 - we can use line breaks
 - there is a strict distinction between name and desc.


I think as we reached 2 plugins + template, it's a fair time to think about extracting this tool to a separate repo.

I was thinking of making a PR to jsdoc/jsdoc with the `@fires` fix/feature. But I think, that may be too much effort, given there is a jsdoc 4.x line coming, and I already spend much time on jsdoc instead of actual events.

eason9487 (approved):
Tested locally and works well. LGTM.

PR #260, opened 2021-11-15
Handle WC5.9 deprecation 

### LINKED ISSUE #259: Deprecated function log: Onboarding:should_show_task.
### Describe the bug:

In WC5.9 Onboarding:should_show_task() was deprecated. TaskList API should be used instead.

### Steps to reproduce:

1. Install and activate the plugin.
2. Go to Marketing -> Pinterest
3. Error should be in logs or directly on screen depending on logging settings.

image

#### Expected behavior

Menu opens without issues.

#### Actual behavior

Deprecation information.

budzanowski:
Closes #259 

Fixes WC59 deprecation logs for Onboarding::should_show_tasks().

Deprecation handling is performed through a simple static method of a Compat class. A class method was used instead of a plain file with a function to leverage PSR4 autoloading and namespaces.

1. Follow steps in #259 and confirm that the deprecation log is gone.

### Changelog entry

> Fix - Use Task List API to detect if we should show Pinterest onboarding tasks.

### DIFF
--- a/src/Compat.php
+++ b/src/Compat.php
@@ -1,0 +1,38 @@
+<?php
+/**
+ * Deprecated methods compatibility layer.
+ *
+ * @package Pinterest/Compat
+ */
+
+namespace Automattic\WooCommerce\Pinterest;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+use Automattic\WooCommerce\Admin\Features\Onboarding;
+use Automattic\WooCommerce\Admin\Features\OnboardingTasks\TaskLists;
+
+/**
+ * Helper class with functions that handle WordPress and WooCommerce deprecations.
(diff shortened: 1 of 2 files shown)


src/Compat.php
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Helper class with functions that handle WordPress and WooCommerce deprecations.
+ * Using helper class with static methods to help with autoloading.
+ */
+class Compat {
+	/**
+	 * Helper function to check if UI should show tasks.
+	 *
+	 * @since x.x.x
```
layoutd, line 25: Personal preference, I guess, but the `since` tag could also go with the `class` doc block. Although, with only one method, it makes sense here too.

layoutd (approved):
💯  Looks good, works well - removes the deprecation warning in 5.9

PR #263, opened 2021-11-16
Fix pull request template to make it work with WooRelease changelog generation

ecgan:

This PR fixes the `.github/pull_request_template.md` file: the "Changelog note" should be "Changelog entry", so that it works with WooRelease tool.

The new pull request template here is exactly the same as the one from Google Listings and Ads repo:

https://github.com/woocommerce/google-listings-and-ads/blob/43146b68ba19fb6fe3f9fb737d28b34931c08819/.github/pull_request_template.md

Once we have this PR merged, new PR should have the same template as Google Listings and Ads, and it should work well with WooRelease changelog generation.

### Changelog note

>

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,23 +1,36 @@
-<!-- You can erase any parts of this template not applicable to your pull request. -->
+### Changes proposed in this Pull Request:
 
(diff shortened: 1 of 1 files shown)


ianlin (approved):
Nice 👍

PR #261, opened 2021-11-15
Add Track Events to documentation links.

tomalec:
:warning:  This PR is based on https://github.com/woocommerce/pinterest-for-woocommerce/pull/254, for better documentation 

Implements part of #232 

### Changes proposed in this pull request

Add Track Events to documentation links.

Change `StepOverview` to accept more props for read more `Button` link.

### Detailed test instructions

1. Enable tracks logging, by `localStorage.setItem('debug', 'wc-admin*')`
2. Check the `wcadmin_pfw_documentation_link_click` events fired for the following links:
	- at [`/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Flanding`](https://wp581wc580.test/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Flanding)
		-  "Terms of Service." - `{ link_id: 'terms-of-service', context: 'welcome-section' }`
	- at [`/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Fonboarding`](https://wp581wc580.test/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Fonboarding)
		- "Pinterest advertising guidelines." - `{ link_id: 'ad-guidelines', context: 'setup-account' }`
		image

	- at [`/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Fonboarding&step=claim-website`](https://wp581wc580.test/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Fonboarding&step=claim-website)
		- "Read more" - `{ link_id: 'claim-website', context: 'claim-website' }` 
		image
	- at [`/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Fonboarding&step=setup-tracking`](https://wp581wc580.test/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Fonboarding&step=setup-tracking)
		-  "Ad Guidelines" - `{ link_id: 'ad-guidelines', context: 'wizard'|'settings' }`
		-  "Ad Data Terms" - `{ link_id: 'ad-data-terms', context: 'wizard'|'settings' }`
		-  "Read more" - `{ link_id: 'setup-tracking', context: 'wizard'|'settings' }`
		- "Pinterest Advertising Agreement" - `{ link_id: 'tos', context: 'wizard'|'settings' }`
	image

(no changelog, as we will add a single entry for a feature branch)

PR #266, opened 2021-11-17
Add documentation link props ~hook~ helper.

tomalec:
This PR aims to improve https://github.com/woocommerce/pinterest-for-woocommerce/pull/261

- Separate Documentation Link props creation to a helper. 
	Make the code more DRY, and assert consistency for all links.
- Set `rel:'noopener'` for external link Buttons. 
	To secure them in IE.

_Replace this with a good description of your changes & reasoning._

Double-check that, the changes introduced in https://github.com/woocommerce/pinterest-for-woocommerce/pull/261 works as before.

### Additional notes:
1. This PR exercises Component hooks described in p7H4VZ-2IH-p2. However, I created a helper, not a hook, as I don't really get why it would need to be a hook. I'd appreciate a comment from somebody more experienced in React.

(no changelog, as we will add a single entry for a feature branch)
### Changelog entry

### DIFF
--- a/assets/source/setup-guide/app/helpers/documentation-link-props.js
+++ b/assets/source/setup-guide/app/helpers/documentation-link-props.js
@@ -17,7 +17,9 @@ import { recordEvent } from '@woocommerce/tracks';
  * Creates properties for an external documentation link.
  * May take any other props to be extended and forwarded to a link element (`<a>`, `<Button isLink>`).
  *
- * Sets `target="_blank" rel="noopener"` and onClick handler that fires track event.
+ * Sets `target="_blank" rel="noopener"` and `onClick` handler that fires track event.
+ *
+ * Please be careful not to overwrite the `onClick` handler coincidently.
  *
  * @fires wcadmin_pfw_documentation_link_click on click, with given `linkId` and `context`.
  *
@@ -27,8 +29,10 @@ import { recordEvent } from '@woocommerce/tracks';
  * @param {string} props.context Forwarded to {@link wcadmin_pfw_documentation_link_click}
  * @param {string} [props.target='_blank']
  * @param {string} [props.rel='noopener']
- * @param {Function} [props.onClick]
+ * @param {Function} [props.onClick] onClick event handler to be decorated with firing Track event.
  * @param {...import('react').AnchorHTMLAttributes} props.props
+ *
+ * @return {{herf: string, target: string, rel: string, onClick: Function, props}} Documentation link props.
  */
 export default function documentationLinkProps( {
 	href,
--- a/assets/source/setup-guide/app/helpers/documentation-link-props.js
+++ b/assets/source/setup-guide/app/helpers/documentation-link-props.js
@@ -34,7 +34,7 @@ import { recordEvent } from '@woocommerce/tracks';
  *
  * @return {{herf: string, target: string, rel: string, onClick: Function, props}} Documentation link props.
  */
-export default function documentationLinkProps( {
+function documentationLinkProps( {
 	href,
 	linkId,
 	context,
@@ -62,3 +62,4 @@ export default function documentationLinkProps( {
 		},
 	};
 }
+export default documentationLinkProps;
--- a/assets/source/setup-guide/app/helpers/documentation-link-props.js
+++ b/assets/source/setup-guide/app/helpers/documentation-link-props.js
@@ -27,7 +27,7 @@ import { recordEvent } from '@woocommerce/tracks';
  * @param {string} props.context Forwarded to {@link wcadmin_pfw_documentation_link_click}
  * @param {string} [props.target='_blank']
  * @param {string} [props.rel='noreferrer']
- * @param {Function} props.onClick
+ * @param {Function} [props.onClick]
  * @param {...import('react').AnchorHTMLAttributes} props.props
  */
 export default function documentationLinkProps( {
--- a/assets/source/setup-guide/app/helpers/documentation-link-props.js
+++ b/assets/source/setup-guide/app/helpers/documentation-link-props.js
@@ -19,7 +19,7 @@ import { recordEvent } from '@woocommerce/tracks';
  *
  * Sets `target="_blank" rel="noreferrer"` and onClick handler that fires track event.
  *
- * @fires pfw_documentation_link_click on click, with given `linkId` and `context`.
+ * @fires wcadmin_pfw_documentation_link_click on click, with given `linkId` and `context`.
  *
  * @param {Object} props React props.
  * @param {string} props.href Href to used by link and in track event.
--- a/assets/source/setup-guide/app/helpers/documentation-link-props.js
+++ b/assets/source/setup-guide/app/helpers/documentation-link-props.js
@@ -1,0 +1,60 @@
+/**
+ * External dependencies
+ */
+import { recordEvent } from '@woocommerce/tracks';
+
+/**
+ * Clicking on an external documentation link.
+ *
+ * @event wcadmin_pfw_documentation_link_click
+ *
+ * @property {string} link_id Identifier of the link.
+ * @property {string} context What action was initiated.
+ * @property {string} href Href to which the user was navigated to.
+ */
+
+/**
+ * Creates properties for an external documentation link.
+ * May take any other props to be extended and forwarded to a link element (`<a>`, `<Button isLink>`).
+ *
+ * Sets `target="_blank" rel="noreferrer"` and onClick handler that fires track event.
+ *
+ * @fires pfw_documentation_link_click on click, with given `linkId` and `context`.
+ *
+ * @param {Object} props React props.
+ * @param {string} props.href Href to used by link and in track event.
+ * @param {string} props.linkId Forwarded to {@link wcadmin_pfw_documentation_link_click}
+ * @param {string} props.context Forwarded to {@link wcadmin_pfw_documentation_link_click}
+ * @param {string} [props.target='_blank']
+ * @param {string} [props.rel='noreferrer']
+ * @param {Function} props.onClick
+ * @param {...import('react').AnchorHTMLAttributes} props.props
+ */
+export default function documentationLinkProps( {
+	href,
+	linkId,
+	context,
+	target = '_blank',
+	rel = 'noreferrer',
+	onClick,
+	...props
+} ) {
+	return {
+		href,
+		target,
+		rel,
+		...props,
+		onClick: ( event ) => {
+			if ( onClick ) {
+				onClick( event );
+			}
+			if ( ! event.defaultPrevented ) {
+				recordEvent( 'pfw_documentation_link_click', {
+					link_id: linkId,
+					context,
+					href,
+				} );
+			}
+		},
+	};
+}
--- a/assets/source/setup-guide/app/helpers/documentation-link-props.js
+++ b/assets/source/setup-guide/app/helpers/documentation-link-props.js
@@ -17,7 +17,7 @@ import { recordEvent } from '@woocommerce/tracks';
  * Creates properties for an external documentation link.
  * May take any other props to be extended and forwarded to a link element (`<a>`, `<Button isLink>`).
  *
- * Sets `target="_blank" rel="noreferrer"` and onClick handler that fires track event.
+ * Sets `target="_blank" rel="noopener"` and onClick handler that fires track event.
  *
  * @fires wcadmin_pfw_documentation_link_click on click, with given `linkId` and `context`.
  *
@@ -26,7 +26,7 @@ import { recordEvent } from '@woocommerce/tracks';
  * @param {string} props.linkId Forwarded to {@link wcadmin_pfw_documentation_link_click}
  * @param {string} props.context Forwarded to {@link wcadmin_pfw_documentation_link_click}
  * @param {string} [props.target='_blank']
- * @param {string} [props.rel='noreferrer']
+ * @param {string} [props.rel='noopener']
  * @param {Function} [props.onClick]
  * @param {...import('react').AnchorHTMLAttributes} props.props
  */
@@ -35,7 +35,7 @@ export default function documentationLinkProps( {
 	linkId,
 	context,
 	target = '_blank',
-	rel = 'noreferrer',
+	rel = 'noopener',
 	onClick,
 	...props
 } ) {
--- a/TRACKING.md
+++ b/TRACKING.md
@@ -19,15 +19,15 @@ Clicking on "Connect" Pinterest account button.
 #### Emitters
 - [`AccountConnection`](assets/source/setup-guide/app/components/Account/Connection.js#L60)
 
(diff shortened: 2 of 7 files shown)


assets/source/setup-guide/app/helpers/documentation-link-props.js
```diff
@@ -0,0 +1,60 @@
...
+ * @property {string} href Href to which the user was navigated to.
+ */
+
+/**
+ * Creates properties for an external documentation link.
+ * May take any other props to be extended and forwarded to a link element (`<a>`, `<Button isLink>`).
+ *
+ * Sets `target="_blank" rel="noreferrer"` and onClick handler that fires track event.
+ *
+ * @fires pfw_documentation_link_click on click, with given `linkId` and `context`.
```
eason9487: 💅  Seems to have missed the prefix.

```suggestion
 * @fires wcadmin_pfw_documentation_link_click on click, with given `linkId` and `context`.
```

assets/source/setup-guide/app/helpers/documentation-link-props.js
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * @param {Object} props React props.
+ * @param {string} props.href Href to used by link and in track event.
+ * @param {string} props.linkId Forwarded to {@link wcadmin_pfw_documentation_link_click}
+ * @param {string} props.context Forwarded to {@link wcadmin_pfw_documentation_link_click}
+ * @param {string} [props.target='_blank']
+ * @param {string} [props.rel='noreferrer']
+ * @param {Function} props.onClick
```
eason9487: 💅  Looks like it'd be proper to mark `onClick` as an optional prop?

```suggestion
 * @param {Function} [props.onClick]
```

assets/source/setup-guide/app/helpers/documentation-link-props.js
```diff
@@ -0,0 +1,60 @@
...
+	context,
+	target = '_blank',
+	rel = 'noreferrer',
+	onClick,
+	...props
+} ) {
+	return {
+		href,
+		target,
+		rel,
```
eason9487, line 49: 📜 
I was thinking would it be better if use `'noopener'` as the base `rel` value internally when the `target` is `'_blank'`, and then the passing `rel` is combined and deduplicated as the final `rel`? Thus, `documentationLinkProps( { ref: 'noreferrer' } )` would return `{ ref: 'noopener noreferrer' }`.

assets/source/setup-guide/app/helpers/documentation-link-props.js
```diff
@@ -0,0 +1,60 @@
...
+		...props,
+		onClick: ( event ) => {
+			if ( onClick ) {
+				onClick( event );
+			}
```
eason9487, line 54: 📜 
A little bit of concern here when using this helper like this:

```js
// Case A
const buttonA = (
	<Button
		isLink
		onClick={ () => { /* do something */ } }
		{ ...documentationLinkProps( {
			// ...omitted
		} ) }
	>
		Hi there
	</Button>
);

// Case B
const buttonB = (
	<Button
		isLink
		{ ...documentationLinkProps( {
			// ...omitted
		} ) }
		onClick={ () => { /* do something */ } }
	>
		Hi there
	</Button>
);
```

The `onClick` handler of **Case A** will be overwritten and not be called. And the handler of **Case B** will overwrite the tracking event handler. And these potential issues cannot be checked by codes or lint tools.
- For Case A, it's easy to find during development, but it could happen accidentally when moving code.
- For Case B, it's easy to miss that tracking event is overwritten and not triggered, after all, tracking occurs in the background and probably is not set `localStorage.setItem('debug', 'wc-admin*')` or DevTool Console does not monitor the verbose log.

However, considering that links using this helper usually don't have an additional `onClick` handler, perhaps it would be better to enhance the use of JSDoc regarding `onClick`.

eason9487 (approved):
Testing well. Left two non-urgent discussions and two presumably typo comments. None of them should be a blocker. LGTM.

eason9487:
> This PR exercises Component hooks described in p7H4VZ-2IH-p2. However, I created a helper, not a hook, as I don't really get why it would need to be a hook. I'd appreciate a comment from somebody more experienced in React.

I would say it's probably for the flexibility to expand in the future when it needs to be a real React hook. For example, we have a `useButton` even thougt it doesn't have any real React hook things:

```js
function useButton( props = {} ) {
	return {
		role: 'button',
		tabIndex: 0,
		...props,
		onClick: mergeEvent( props.onClick, () => recordTrackEvent('button_clicked') ),
	};
}
```

Someday we want the `onClick` reference to be changed only when needed instead of every rendering, then adjust it to:

```js
function useButton( props = {} ) {
	const onClick = useCallback( () => {
		mergeEvent( props.onClick, () => recordTrackEvent( 'button_clicked' ) );
	}, [ props.onClick ] );

	return {
		role: 'button',
		tabIndex: 0,
		...props,
		onClick,
	};
}
```

Because it was treated as a React hook from the beginning and follows its limitations, e.g. we don't call it conditionally or in a for loop, then there should be no need to make any changes on consumer sides.

assets/source/setup-guide/app/helpers/documentation-link-props.js
tomalec: Like a polyfill for IE? I like the idea!

assets/source/setup-guide/app/helpers/documentation-link-props.js
tomalec: Thanks for catching that! IMHO, silent future bugs are a quite serious thing.
Do you have any idea how to solve that?

I'd try to emphasize that somewhat in the docs, and change the usage in this PR to 
```jsx
<Button
{ ...documentationLinkProps( {
	href:
		wcSettings.pinterest_for_woocommerce
			.pinterestLinks.adGuidelines,
	linkId: 'ad-guidelines',
	context: view,
	rel: 'noopener',
	// Other Button props.
	isLink,
} ) }>
```
To somewhat suggest that it's safer to pass all the props through the helper.

tomalec:
> Because it was treated as a React hook from the beginning and follows its limitations, e.g. we don't call it conditionally or in a for loop, then there should be no need to make any changes on consumer sides.

Got it. So play safe today, and adhere to sticker constraints than needed, to make potential future changes easier.

Currently, we do not do much "hook things" with props, and we have one conditional usage.
Do you think it's better to 
- change it to be a hook, and adapt the usage
- YAGNI, leave it as a non-hook, and hook-ify it once needed.

-----

> Someday we want the `onClick` reference to be changed only when needed instead of every rendering, then adjust it to

Actually, I was thinking, should we do that today?

assets/source/setup-guide/app/helpers/documentation-link-props.js
eason9487: > Do you have any idea how to solve that?

I haven't used the "Component hooks" pattern in practice, my opinion is just based on my imagination and empirical inference, hope it could still bring some idea though.

The pattern looks more likely to be used in a fundamental components library. These core components might share a lot of reusable or composable props. Thus, with this pattern, it creates default props for components that are designed to be overridden if there are other props specified after its sequence.

For example, a library provides a `<RichTextarea>` that has a default `onPaste` handler for accepting Markdown syntax, and the default handler is composited by a bunch of Component hooks. And the docs would indicate that if library users set `onPaste` on it, that means they want to handle the paste event themself instead of the default handler. And if library users need to append their handler to pre/post-process something, I would design other dedicated props for that and pass them to internal `useMyAwesomeMarkdownParser()` to merge events.

But in our application-level plugins development, this pattern does not seem to work so smoothly. Since when using React, if there is no particular reason, the default props object spreading is usually placed at the first because the convention is that they are fine to be overridden. And props that need to be synthesized would be processed by separate logic.

Back to this case. It's a bit tricky because we don't expect the tracking calling in `onClick` to be covered, but in the meantime, these tracking usages would probably not be added with additional `onClick` callbacks either. Probably all of them are working well with the current implementation in the future, eventually. But the problem is as you said "silent future bugs are a quite serious thing", which I totally agree with it. I've been thinking about this issue today till now, but I don't think of a solution based on this pattern. So I'd suggest considering another way if we still have no idea then.

eason9487:
> Currently, we do not do much "hook things" with props, and we have one conditional usage. Do you think it's better to
> 
> * change it to be a hook, and adapt the usage
> * YAGNI, leave it as a non-hook, and hook-ify it once needed.

We could have more discussions later if the conclusion of https://github.com/woocommerce/pinterest-for-woocommerce/pull/266#discussion_r753120675 is we keep using the Component hooks.

> Actually, I was thinking, should we do that today?

To keep all references without unnecessary changes usually is tough and brings complex logic when maintaining the `deps` array of all hooks. I usually use these principles to decide whether to handle reference changes of props/hooks more delicately. Whether inaccurate changes of the reference:
- affect unexpected changes in other dependent references on consumer side
- is a basis/shared hook/prop
- have impacts on render/calculation performance
- is easy and smooth to handle the unnecessary changes

If none of the above are met, then it usually does not matter if we don't do that. Just let them change every rendering 🙈 
```js
const something = a ? b : c;
const handleClick = () => {
  sendTracking( something );
};
```

assets/source/setup-guide/app/helpers/documentation-link-props.js
tomalec: After a second thought, I think that maybe it's too much magic. Also, it would introduce another concern similar to https://github.com/woocommerce/pinterest-for-woocommerce/pull/266#discussion_r751915176 and more:

What about 
1. `<MyElement target="_blank" {...documentationLinkProps}/>`
2. `<MyElement {...documentationLinkProps( { target="_blank", rel: false} )} />`
3. `<MyElement {...documentationLinkProps( { target="_blank", rel: ''} )} />`

WDYT of making it more straightforward, and a bit opposite to what we have now?:

https://github.com/woocommerce/pinterest-for-woocommerce/pull/266/commits/967a0aaf6e595bd805890ca5e874f4a86ac1827a
`documentationLinkProps` will set `rel: noopener` by default. Then the number of current use-cases, would set it to `noreferrer` ([which implies `noopener`](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noreferrer))


BTW, I'm not sure if those use cases should use `noreferrer`, as I thought Pinterest could be interested to track referrers to their docs.

tomalec:
Thanks for all the explanations @eason9487 :)

I addressed `noreferred`/`noopener` case a bit differently https://github.com/woocommerce/pinterest-for-woocommerce/pull/266#discussion_r754275669

Speaking of potentially silent future errors related to overwriting `onClick` https://github.com/woocommerce/pinterest-for-woocommerce/pull/266#discussion_r751915176
I added a small note in JSDoc https://github.com/woocommerce/pinterest-for-woocommerce/pull/266/commits/26b7d3402f5207e22366a77e570f1384a7246d8f
I'm not sure if that's enough, or should we add something more.

In general, I start to think. there is not much we can do to prevent silent bugs in the future. This is almost the same use case as with any other:
```jsx
<MyElement
	prop1
	{ ...foo }
	prop2
	/>
```
If a developer is not aware of what's inside `foo` they may overwrite it with `prop2`, or get the `prop1` overwritten.


If we aim to guard against introducing a regression bug of removing the track event, that may be easily missed by manual human testing, then we should add automated tests for that.

----

WDYT of the above? Do you think we can proceed with merging this PR?

assets/source/setup-guide/app/helpers/documentation-link-props.js
eason9487: > WDYT of making it more straightforward, and a bit opposite to what we have now?

I think that's good to make it straightforward and take `noopener` as default value. 👍 

> BTW, I'm not sure if those use cases should use `noreferrer`, as I thought Pinterest could be interested to track referrers to their docs.

I don't know either. Staying the same as the current codebase should be ok. And if Pinterest side needs this we can do it by separate PR.

eason9487:
Sorry for the late reply. 😅 

> Speaking of potentially silent future errors related to overwriting `onClick` [#266 (comment)](https://github.com/woocommerce/pinterest-for-woocommerce/pull/266#discussion_r751915176)
> I added a small note in JSDoc [26b7d34](https://github.com/woocommerce/pinterest-for-woocommerce/commit/26b7d3402f5207e22366a77e570f1384a7246d8f)
> I'm not sure if that's enough, or should we add something more.

Thanks for adding the note in JSDoc. I think it's clearly stated what to notice about `onClick`.

> In general, I start to think. there is not much we can do to prevent silent bugs in the future. This is almost the same use case as with any other:
> 
> // ...omited
> 
> If a developer is not aware of what's inside `foo` they may overwrite it with `prop2`, or get the `prop1` overwritten.

Indeed it is. Developers usually take care of the use of props object spreading. I'm just a bit worried the silent bugs could happen accidentally. It popped into my head last night that a new idea might be considered, but only for this kind of event tracking, not like the generalized pattern of Component hooks. I will post it in this PR later by a separate comment.

> WDYT of the above? Do you think we can proceed with merging this PR?

I think it's good to proceed with merging. Thanks for the all updates!

eason9487:
📜  +💡 
Mainly for proposing an idea. If want to adopt this method, it could be done by another PR.

> It popped into my head last night that a new idea might be considered, but only for this kind of event tracking, not like the generalized pattern of Component hooks. I will post it in this PR later by a separate comment.

The idea is to use a single global click event with `capture` type, then it doesn't need to bind `onClick` prop on the target React element. So it'd not overwrite others or be overwritten by others. But the limitation is it has to be a react hook.

```js
import { uniqueId } from 'lodash';
import { useRef, useEffect } from '@wordpress/element';
import { recordEvent } from '@woocommerce/tracks';

const DATA_ATTR = 'data-pwf-doc-link-id';
const payloadMap = new Map();

function handleClick( e ) {
	const id = e.target.getAttribute( DATA_ATTR );
	const payload = payloadMap.get( id );
	if ( payload ) {
		recordEvent( 'pfw_documentation_link_click', payload );
	}
}

export default function useDocLinkTracking( {
	href,
	linkId,
	context,
	target = '_blank',
	rel = 'noopener',
} ) {
	const idRef = useRef( null );
	if ( ! idRef.current ) {
		idRef.current = uniqueId();
	}

	useEffect( () => {
		if ( payloadMap.size === 0 ) {
			document.defaultView.addEventListener( 'click', handleClick, true );
		}

		const id = idRef.current;

		payloadMap.set( id, {
			link_id: linkId,
			context,
			href,
		} );

		return () => {
			payloadMap.delete( id );

			if ( payloadMap.size === 0 ) {
				document.defaultView.removeEventListener(
					'click',
					handleClick,
					true
				);
			}
		};
	}, [ href, linkId, context ] );

	return {
		[ DATA_ATTR ]: idRef.current,
		href,
		target,
		rel,
	};
}
```

Considering that registering a global click event for each tracking usages should not be a problem. So the click event could be added and removed inside the `useDocLinkTracking` hook, therefore, the `if ( payloadMap.size === 0 )` checks and the `payloadMap` map are unneeded.

Issue #206, opened 2021-09-17
Do not sync hidden/out of stock products


wjia88:
### Describe the bug:
During catalog ingestion, all products, including products that were marked as hidden/out of stock by the user were synced.

One user noticed that some products that were Out of Stock (and hidden from customers) in her store had been synced to Pinterest after ingestion was complete. Because of this, the user had to manually delete the undesired products from WooCommerce and wait for the integration to sync again, at which point the products were removed from Pinterest.

#### Expected behavior
By default if any products are marked as Out of Stock or Unavailable, do not include these in feed ingestion.

### Additional details:

Recommend updating the logic to check to make sure that
* Products are published (and not draft or private/password-protected) - For variations, we recommend checking the status of the parent variable product and not the child.
* Is a supported product type if relevant (e.g. simple products or variations etc..)
* Is visible in the store’s catalog (it’s not manually hidden from the store)

Out-of-stock products are by default visible and they become hidden only if the `woocommerce_hide_out_of_stock_items
` WooCommerce option is set to ‘yes’; therefore, we recommend only sync those out-of-stock products that are not hidden from the store catalog.

WooCommerce’s `\WC_Product::is_visible` method can be used to check if a product is visible or not. This internal method checks multiple options and applies filters to check whether a product should be visible in store’s catalog.

In Google listings & Ads if the product is a variation, we apply an internal WooCommerce filter called `woocommerce_hide_invisible_variations` to make sure that the variation is actually visible before syncing it. This is to mimic the conditions set by the `\WC_Product_Variable::get_available_variations` method to retrieve available variations. It confirms that disabled variations and variations with empty prices are not synced to Merchant Center.

findingsimple:
> product that were marked as hidden / out of stock by the user were synced

Looking at the code, the extension does support the availability field and maps it to Woo stock status https://github.com/woocommerce/pinterest-for-woocommerce/blob/develop/src/ProductsXmlFeed.php#L219

My understanding is it should be fine to sync "out of stock" items based on this.

However...

I suspect we should better support the `woocommerce_hide_out_of_stock_items` options

<img width="610" alt="Screen Shot 2021-10-12 at 9 41 42 pm" src="https://user-images.githubusercontent.com/355014/136945622-fe30d0a3-9dab-4edb-99f6-96c5f702e68f.png">

I suspect we should better support the Catalog Visibility option and not sync a product if it is set to "Hidden"
[...]

Issue #223, opened 2021-10-12
Simplify register_feed()


findingsimple:
This is a follow up from https://github.com/woocommerce/pinterest-for-woocommerce/issues/211

See https://github.com/woocommerce/pinterest-for-woocommerce/issues/211#issuecomment-932209153

> in register_feed we are trying to find feed profiles created by Woo plugin. register_feed code can be simplified by using `/v3/catalogs/<merchant>/feed_profiles/` api which returns all feed profiles for a merchant and includes which feed profiles were created by Woo.

We can include this as part of multi feed effort (will cross-link when we have some issue/s).

budzanowski:
Fixed with https://github.com/woocommerce/pinterest-for-woocommerce/pull/305

PR #271, opened 2021-11-22
Refactor feed generation process.

### LINKED ISSUE #185: Product sync architecture review (supporting sync problems, scaling to many products)
Product sync is a fundamentally important feature. It's also a complex thing to implement so it's robust to errors/issues and scales to many products (1000s => 100000s and beyond).

While reviewing various product-sync related PRs, I've noticed a few things that might need follow up ([example](https://github.com/woocommerce/pinterest-for-woocommerce/pull/182#issuecomment-903347696)). Also I've found the sync code generally hard to review/follow – though that [might be mitigated with more documentation on the process & state](https://github.com/woocommerce/pinterest-for-woocommerce/issues/132).

If we review the sync code and process, and document it clearly, this will give us a much clearer understanding of what could be improved, and help support sync issues that may come up for users.

Specific items to review:

- How feed / sync state is managed and stored.
- What stages (states) there are in the feed sync process, e.g.: registering with Pinterest, generating a file, waiting for file to be requested, getting errors/status from Pinterest, disabling cancelling/unregistering sync.
- How async / scheduled actions are scheduled and how they interact with each other and state.
- How feed generation scales with many products.
- How feed re-generation or dirty state is handled.

Best case scenario, the current code is robust and handles all the above items well. If this is the case, can document the system clearly (e.g. code comments, markdown, maybe a flow chart) and close this issue. Alternatively, if some issues are found, we can log issues to follow up.

### LINKED ISSUE #242: feed xml: Remove parent item node when child items (variants are present)
Feed xml includes both parent item and child items (variants). Pinterest feed ingestion creates pins for parent item as well as child items. 

For e.g. in attached sample xml [sample.txt](https://github.com/woocommerce/pinterest-for-woocommerce/files/7481057/sample.txt)

Parent item 12 V-Neck T-Shirt has 3 child items and on pinterest 4 pins are created.

Screen Shot 2021-10-22 at 9 25 37 AM

Parent item needs to be removed from feed xml

### LINKED ISSUE #257: Implement multiple feed files generation ( server-side )
### General information
Pinterest service is able to operate on separate feeds for different countries, locales, etc. Each feed requires a separate file with country/locale-specific content.

Update https://github.com/woocommerce/pinterest-for-woocommerce/blob/develop/src/ProductSync.php code to support registration, creation and upload of multiple feed files.

No alternatives available - this is a new feature.

### Additional context 

This has a precondition: 
[Product sync architecture review (supporting sync problems, scaling to many products)](https://github.com/woocommerce/pinterest-for-woocommerce/issues/185)

budzanowski:
This PR updates the feed generation process.
The main goals for the PR:

- simplified and streamlined generation process
- better error handling
- multi-feed support on the server-side
- predictable scalability
- extensibility of the process

Bonus:
- removes variable parents from XML ( bug #242 )
- if https://github.com/woocommerce/pinterest-for-woocommerce/pull/284 will be merged then additionally #206 will be fixed.

There are very small changes in the UI compared to the original implementation. The feed status no longer shows the total number of products for processing. The rationale for this change comes from two sources:
1. We no longer precalculate the products list and store it in a DB option. Instead, the implementation fetches pages of product ids during each generation cycle. The presence of a large DB option containing hundreds of thousands of items that would be read and updated during batches generation was flagged as a problem by merchants that have used Facebook For WooCommerce plugin. This plugin was using the same method. After changing to the pagination method the merchants have observed better performance. Learning from that experience I have used the same approach here.
2. When we will have multiple feeds the total number of products may not be a very useful number bc each feed may have a different total number.

Usage of Action Scheduler Jobs Framework ( https://github.com/woocommerce/action-scheduler-job-framework ) will be highly beneficial when we will introduce multiple feed generation. During each generation cycle, we will be able to determine if a particular product should be a part of a particular feed and the `ProductsXmlFeed` class will understand the location parameter passed ( currently it does nothing ) to determine the shape of the feed element for that feed.

Most of the parts of the code are ready for multiple feeds. During the refactor it has become obvious that if this capability would not be implemented alongside the refactoring it would be a wasted effort. There are places that still assume that we have a singular feed. Those places need to wait for UI changes part of the Multiple-feeds feature.

This is based on improvements from https://github.com/woocommerce/pinterest-for-woocommerce/pull/240 ( now in develop )

TODO

- [ ]  https://github.com/woocommerce/pinterest-for-woocommerce/pull/271#issuecomment-992674799

 =

Closes #257
Closes #242
Closes #185 

Please regenerate autoloader files after changing to this branch ( easiest way would be to just delete vendors folder and then composer install )

There are two distinct scenarios:

- First: fresh install
1. Install the plugin.
2. Make the connection.

- Second: updating over an existing version
1. Update the plugin.
2. Old actions should be removed
3. Feed file should have the location postfix added
4. Feed configuration should be updated on the server to match the change from 3.

### Changelog entry
Fix - Remove parent variable products from the feed file. ( #242 )
Update - Feed generation process refactor. ( #185 )
Add - Multiple feed backend.
Tweak - Additional logging of the feed generation process.
>


class-pinterest-for-woocommerce.php
```diff
@@ -230,8 +243,9 @@ public function init_plugin() {
 			add_action( 'wp_head', array( $this, 'maybe_inject_verification_code' ) );
 			add_action( 'wp_head', array( Pinterest\RichPins::class, 'maybe_inject_rich_pins_opengraph_tags' ) );
 			add_action( 'wp', array( Pinterest\SaveToPinterest::class, 'maybe_init' ) );
+			add_action( 'init', array( $this, 'maybe_update_plugin' ) );
 			add_action( 'init', array( Pinterest\Tracking::class, 'maybe_init' ) );
-			add_action( 'init', array( Pinterest\ProductSync::class, 'maybe_init' ) );
+			add_action( 'init', array( Pinterest\ProductSync::class, 'maybe_init' ), 20 ); // Wait for possible update before starting sync.
```
JPry: Minor: It would be better to put the `maybe_update_plugin` hook at an earlier priority instead of changing the priority of these.

pinterest-for-woocommerce.php
```diff
@@ -44,7 +44,7 @@
 }
 
 define( 'PINTEREST_FOR_WOOCOMMERCE_PLUGIN_FILE', __FILE__ );
-define( 'PINTEREST_FOR_WOOCOMMERCE_VERSION', '1.0.1' ); // WRCS: DEFINED_VERSION.
```
JPry, line 47: Don't we want to only change this value during the release process?

src/API/FeedIssues.php
```diff
@@ -56,7 +56,7 @@ public function __construct() {
 	public function get_feed_issues( WP_REST_Request $request ) {
 
 		try {
-			$feed_id = Pinterest\ProductSync::get_registered_feed_id();
+			$feed_id = Pinterest\FeedRegistration::get_registered_feed_id();
```
JPry: While we're updating this line, it would be better to do a full import at the top of the file so that we don't have a partial namespace here.

```suggestion
			$feed_id = FeedRegistration::get_registered_feed_id();
```

src/API/FeedState.php
```diff
@@ -336,7 +325,7 @@ private function add_feed_registration_state( $result ) {
 	 */
 	private function add_feed_sync_status( $result ) {
 
-		$feed_id = Pinterest\ProductSync::get_registered_feed_id();
+		$feed_id = Pinterest\FeedRegistration::get_registered_feed_id();
```
JPry:
```suggestion
		$feed_id = FeedRegistration::get_registered_feed_id();
```

src/FeedGenerator.php
```diff
@@ -0,0 +1,475 @@
...
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+use Automattic\WooCommerce\ActionSchedulerJobFramework\Utilities\BatchQueryOffset;
+use Automattic\WooCommerce\ActionSchedulerJobFramework\AbstractChainedJob;
+use Automattic\WooCommerce\ActionSchedulerJobFramework\Proxies\ActionSchedulerInterface;
+use Automattic\WooCommerce\Pinterest\Utilities\FeedLogger;
+/**
```
JPry, line 21:
```suggestion

/**
```

src/FeedGenerator.php
```diff
@@ -0,0 +1,475 @@
...
+	/**
+	 * Get a set of items for the batch.
+	 *
+	 * NOTE: when using an OFFSET based query to retrieve items it's recommended to order by the item ID while
+	 * ASCENDING. This is so that any newly added items will not disrupt the query offset.
+	 *
+	 * @param int   $batch_number The batch number increments for each new batch in the job cycle.
+	 * @param array $args         The args for the job.
+	 *
+	 * @throws Exception On error. The failure will be logged by Action Scheduler and the job chain will stop.
```
JPry, line 178: Minor: This `Exception` needs an import to help the IDE (even though it's within a comment).

JPry: This method should also have an `@return` comment in the docblock.

src/FeedGenerator.php
```diff
@@ -0,0 +1,475 @@
...
+	public static function cancel_jobs() {
+		as_unschedule_all_actions( self::ACTION_START_FEED_GENERATOR, array(), PINTEREST_FOR_WOOCOMMERCE_PREFIX );
+	}
+
+	/**
+	 * Start the queue processing.
+	 *
+	 * @since x.x.x
+	 */
+	public function start_generation() {
```
JPry: Does this method need to be `public`? If it's only for adding it to the the appropriate WordPress hook system, we can instead wrap the method call in an anonymous function to keep this either `protected` or `private`.

src/FeedGenerator.php
```diff
@@ -0,0 +1,475 @@
...
+	 *
+	 * @since x.x.x
+	 * @param ActionSchedulerInterface $action_scheduler           Action Scheduler proxy.
+	 * @param LocalFeedConfigs         $local_feeds_configurations Locations configuration class.
+	 */
+	public function __construct( ActionSchedulerInterface $action_scheduler, $local_feeds_configurations ) {
+		parent::__construct( $action_scheduler );
+		$this->configurations = $local_feeds_configurations;
+
+		add_action( self::ACTION_START_FEED_GENERATOR, array( $this, 'start_generation' ) );
```
JPry: We should avoid putting hooks into the constructor and instead trigger them from a separate method.

src/FeedGenerator.php
```diff
@@ -0,0 +1,475 @@
...
+		return 'pinterest';
+	}
+
+	/**
+	 * Prepare a fresh temporary file for each local configuration.
+	 * Files is populated with the XML headers.
+	 *
+	 * @since x.x.x
+	 */
+	public function prepare_temporary_files(): void {
```
JPry: Does this method need to be `public`?

src/FeedGenerator.php
```diff
@@ -0,0 +1,475 @@
...
+			$this->check_write_for_io_errors( $bytes_written, $config['tmp_file'] );
+		}
+	}
+
+	/**
+	 * Add XML footer to all of the temporary feed files.
+	 *
+	 * @since x.x.x
+	 */
+	public function add_footer_to_temporary_feed_files(): void {
```
JPry: Does this method need to be `public`?

src/FeedGenerator.php
```diff
@@ -0,0 +1,475 @@
...
+	 * Checks the status of the file write operation and throws if issues are found.
+	 * Utility function for functions using file_put_contents.
+	 *
+	 * @since x.x.x
+	 * @param integer $bytes_written How much data was written to the file.
+	 * @param string  $file          File location.
+	 *
+	 * @throws \Exception Can't open or write to the file.
+	 */
+	public function check_write_for_io_errors( $bytes_written, $file ): void {
```
JPry: Does this method need to be `public`?

src/FeedGenerator.php
```diff
@@ -0,0 +1,475 @@
...
+		}
+	}
+
+	/**
+	 * React to errors during feed files generation process.
+	 *
+	 * @since x.x.x
+	 * @param \Throwable $th Exception handled.
+	 */
+	public function handle_error( $th ) {
```
JPry: Does this method need to be `public`?

src/FeedGenerator.php
```diff
@@ -0,0 +1,475 @@
...
+	}
+
+	/**
+	 * Rename temporary feed files to final name.
+	 * This is the last step of the feed file generation process.
+	 *
+	 * @since x.x.x
+	 * @throws \Exception Renaming not possible.
+	 */
+	public function rename_temporary_feed_files_to_final(): void {
```
JPry: Does this method need to be `public`?

src/FeedGenerator.php
```diff
@@ -0,0 +1,475 @@
...
+		);
+
+		return array_map( 'intval', $product_ids );
+	}
+
+	/**
+	 * Processes a batch of items. The middle part of the generation process.
+	 * Can run multiple times depending on the catalog size.
+	 *
+	 * @since 1.1.0
```
JPry:
```suggestion
	 * @since x.x.x
```

src/FeedRegistration.php
```diff
@@ -0,0 +1,239 @@
...
+	 *
+	 * @since x.x.x
+	 * @param LocalFeedConfigs $local_feeds_configurations Locations configuration class.
+	 * @param FeedGenerator    $feed_generator Feed generator class.
+	 */
+	public function __construct( $local_feeds_configurations, $feed_generator ) {
+		$this->configurations = $local_feeds_configurations;
+		$this->feed_generator = $feed_generator;
+
+		add_action( self::ACTION_HANDLE_FEED_REGISTRATION, array( $this, 'handle_feed_registration' ) );
```
JPry, line 58: We should avoid putting hooks into the constructor and instead trigger them from a separate method.

src/FeedRegistration.php
```diff
@@ -0,0 +1,239 @@
...
+	 * and on a scheduled basis.
+	 *
+	 * @return mixed
+	 *
+	 * @throws \Exception PHP Exception.
+	 */
+	public function handle_feed_registration() {
+
+		if ( ! $this->feed_file_exists() ) {
+			self::log( 'Feed didn\'t fully generate yet. Retrying later.', 'debug' );
```
JPry: We should switch to double quotes here to avoid using a `\` in the string.

```suggestion
			self::log( "Feed didn't fully generate yet. Retrying later.", 'debug' );
```

src/FeedRegistration.php
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	/**
+	 * Handles feed registration using the given arguments.
+	 * Will try to create a merchant if none exists.
+	 * Also if a different feed is registered, it will update using the URL in the
+	 * $feed_args.
+	 *
+	 * @param array $feed_args The arguments used to create the feed.
+	 * @param array $location List of served locations.
```
JPry: Isn't `$location` a `string`?

src/FeedRegistration.php
```diff
@@ -0,0 +1,239 @@
...
+			self::log( 'Pinterest returned a Declined status for product_pin_approval_status' );
+		} elseif ( ! empty( $merchant['data']->id ) && ! isset( $merchant['data']->product_pin_feed_profile ) ) {
+			// No feed registered, but we got a merchant.
+			$registered = self::do_add_merchant_feed( $merchant['data']->id, $feed_args );
+		} elseif ( $feed_args['feed_location'] === $merchant['data']->product_pin_feed_profile->location_config->full_feed_fetch_location ) {
+			// Feed registered.
+			$registered = $merchant['data']->product_pin_feed_profile->id;
+			self::log( 'Feed registered for merchant: ' . $feed_args['feed_location'] );
+		} else {
+			$product_pin_feed_profile    = $merchant['data']->product_pin_feed_profile;
```
JPry, line 144: This section of this method seems to be doing quite a lot. It would make it easier to read to break this up into its own method.

src/LocalFeedConfigs.php
```diff
@@ -0,0 +1,91 @@
...
+		$file_name_base = trailingslashit( wp_get_upload_dir()['basedir'] ) . PINTEREST_FOR_WOOCOMMERCE_LOG_PREFIX . '-';
+		$url_base       = trailingslashit( wp_get_upload_dir()['baseurl'] ) . PINTEREST_FOR_WOOCOMMERCE_LOG_PREFIX . '-';
+		array_walk(
+			$feed_ids,
+			function ( &$id, $location ) use ( $file_name_base, $url_base ) {
+				$id = array(
+					'feed_id'   => $id,
+					'feed_file' => $file_name_base . $id . '-' . $location . '.xml',
+					'tmp_file'  => $file_name_base . $id . '-' . $location . '-tmp.xml',
+					'feed_url'  => $url_base . $id . '-' . $location . '.xml',
```
JPry: Interpolation is a bit easier to read than concatenation, especially for so many variables put together.

```suggestion
					'feed_file' => "{$file_name_base}{$id}-{$location}.xml",
					'tmp_file'  => "{$file_name_base}{$id}-{$location}-tmp.xml",
					'feed_url'  => "{$url_base}{$id}-{$location}.xml",
```

src/PluginUpdate.php
```diff
@@ -0,0 +1,110 @@
...
+		as_unschedule_all_actions( PINTEREST_FOR_WOOCOMMERCE_PREFIX . '-feed-generation', array(), PINTEREST_FOR_WOOCOMMERCE_PREFIX );
+
+		/*
+		 * 2. Move feed file to a new location.
+		 */
+		$feed_id          = Pinterest_For_Woocommerce()::get_data( 'local_feed_id' );
+		$default_location = Pinterest_For_Woocommerce()::get_base_country() ?? 'US';
+		if ( $feed_id ) {
+			// Generate new configurations.
+			$new_configs = ( new LocalFeedConfigs( $default_location ) )->get_configurations();
```
JPry: The constructor for `LocalFeedConfigs` is expecting an `array`, not a `string`.

src/ProductFeedStatus.php
```diff
@@ -54,8 +45,7 @@ class ProductFeedStatus {
 	 */
 	public static function get() {
 
-		$local_feed  = self::get_local_feed();
-		$data_prefix = PINTEREST_FOR_WOOCOMMERCE_PREFIX . '_feed_' . $local_feed['feed_id'] . '_';
+		$data_prefix = PINTEREST_FOR_WOOCOMMERCE_PREFIX . '_feeds_';
```
JPry: We repeat this often, so it might make more sense to have its own constant in this class.

JPry (changes requested):

src/FeedRegistration.php
findingsimple: I think this is probably prepping for a move to support multi countries/feeds

JPry: @jconroy Right, `$location` should be a key in the array of feeds. See line 179.

pinterest-for-woocommerce.php
budzanowski: Yes, but I don't really know if this will be 1.1.0 or 2.0.0 or 1.1.x something. This is covered during the release by WooRelease `--product_version` flag the `// WRCS: DEFINED_VERSION. ` indicates this to be auto bumped by the script when the flag is present.

src/API/FeedIssues.php
budzanowski: I have blindly followed the structure of the code in this file already - I can ofc fix it.

class-pinterest-for-woocommerce.php
budzanowski: It looks like moving with the update to 'plugins_loaded' should do the trick.

src/ProductFeedStatus.php
budzanowski: Yes.

src/PluginUpdate.php
budzanowski: Good catch.

src/FeedRegistration.php
budzanowski: This was already in the plugin. I have just moved it here, so no new code was added. I agree that this needs to be done in a slightly more organized and factorized fashion. This part of the code is scheduled for a rewrite using a different approach https://github.com/woocommerce/pinterest-for-woocommerce/issues/223 so I have made a decision not to change it now.

src/FeedRegistration.php
budzanowski: This is preparation for a multi-feed.

src/FeedRegistration.php
budzanowski: Yep.

src/FeedGenerator.php
budzanowski: yep.

budzanowski:
It looks like on each REST call that plugin is making we are setting up the whole plugin stack - maybe we can avoid that to make the calls much leaner. This also applies to the original implementation.

src/FeedGenerator.php
budzanowski: Yes.

src/FeedRegistration.php
```diff
@@ -47,7 +47,15 @@ class FeedRegistration {
 	public function __construct( $local_feeds_configurations, $feed_generator ) {
 		$this->configurations = $local_feeds_configurations;
 		$this->feed_generator = $feed_generator;
+		$this->init();
```
JPry: Moving the hooks to a separate method is 👍🏻 

But the hooks are still called every time you call `new FeedRegistration()`. The `init()` method should be called separately from the constructor. Just like `$feed_generator` here:

https://github.com/woocommerce/pinterest-for-woocommerce/blob/28248c5f5a3e99b2a906df84886c80be6b238cc2/src/ProductSync.php#L107-L113

src/PluginUpdate.php
```diff
@@ -46,7 +46,7 @@ public static function update_to_2_0_0() {
 		 * 2. Move feed file to a new location.
 		 */
 		$feed_id          = Pinterest_For_Woocommerce()::get_data( 'local_feed_id' );
-		$default_location = Pinterest_For_Woocommerce()::get_base_country() ?? 'US';
+		$default_location = array( Pinterest_For_Woocommerce()::get_base_country() ?? 'US' );
```
JPry: We are using `$default location` as an array key, which means we cannot have `$default_location` be an array. Line 54 below will be a problem:

https://github.com/woocommerce/pinterest-for-woocommerce/blob/17d3c336545a41ef4c1760285bd98bf1336b88ed/src/PluginUpdate.php#L54

src/FeedRegistration.php
JPry: > This is preparation for a multi-feed.

I understand, but the `$location` variable is used as the key in the `$feeds` array. With the latest changes, this is down on line 187:

https://github.com/woocommerce/pinterest-for-woocommerce/blob/310fefddb85f66bf7c4a8ddc866fd52e675554c0/src/FeedRegistration.php#L187

This is where having the `$location` variable as an array is a problem, because an array key cannot be an array. Something needs to change so that this can properly work.

src/FeedRegistration.php
JPry: > so I have made a decision not to change it now.

Sounds good 👍🏻

JPry (changes requested):

src/FeedRegistration.php
budzanowski: totally missed that :/

src/PluginUpdate.php
budzanowski: Oh man, sorry for that. I have made the changes and left the self review and testing for today.

src/FeedRegistration.php
budzanowski: Yes, so this is more a problem with docblock for now. So far this function was called only with $location as a string identifiing a single country. I am adjusting the dockblock now, and it will be change later when more multifeed code will be added.

budzanowski:
@JPry Thank you very much for the review. I have made all the requested changes. I will do some additional testing now and will ping you again for more scrutiny. You can still review and test now, ofc, no problem in that but I probably need to do one or two adjustments to the code so things will move.

budzanowski:
Closed in favor of #368

PR #283, opened 2021-11-26
PHP Unit Tests support

puntope:

* Added PHP Unit test support in the project.
* Added GH PHP Test Unit Workflow
* Added Prepare PHP Action (Will allow refactoring of PHPCS workflow)
* Refactoring of PHPCS workflow
* Updated documentation about PHP Unit test in README

### Install Test Dependencies

Install composer dependencies:

`composer install`

To run the unit tests you need WordPress, WooCommerce, and the WordPress Unit Test lib (included in the core development repository).

Install them using the install-wp-tests.sh script:

`$ ./bin/install-wp-tests.sh <db-name> <db-user> <db-pass> <db-host>
Example:

`$ ./bin/install-wp-tests.sh wordpress_tests root root localhost`
This script installs the test dependencies into your system's temporary directory and also creates a test database.

You can also specify the path to their directories by setting the following environment variables:

`WP_TESTS_DIR`: WordPress Unit Test lib directory
`WP_CORE_DIR`: WordPress core directory
`WC_DIR`: WooCommerce directory

### Running Tests
Change to the plugin root directory and type:

`$ vendor/bin/phpunit`
The tests will execute and you'll be presented with a summary.

### Unit Tests and PHP 8
We currently do not support running unit tests on PHP 8.


budzanowski:
This looks great!
I have two comments/requests: 
1. Could you please add relevant comments to package.json so we could use `npm run ...` commands locally to install the tests and run the test suite.
2. Could you please check if it would be possible to incorporate https://packagist.org/packages/spatie/phpunit-watcher for local work? Or something else that serves the same purpose. I love to have run on change behaviour while I am codding. But that is just an additional thing. Absolutely not required for this.

puntope:
> 1. Could you please add relevant comments to package.json so we could use `npm run ...` commands locally to install the tests and run the test suite.

Added some commands 👌 

1. Default

`npm run test:php` Runs with default config (wordpress_test root root localhost)

2. Custom config
`npm run test:php-prepare <db_name> <db_user> <db_pass> <db_host>` 
`npm run test:php-run` 

> Could you please check if it would be possible to incorporate https://packagist.org/packages/spatie/phpunit-watcher for local work? Or something else that serves the same purpose. I love to have run on change behaviour while I am codding. But that is just an additional thing. Absolutely not required for this.

I can create a separate issue for it since I'm in porter and will not have time for it until Monday. So that way we can have rn the tests.

budzanowski:
This looks good. We can skip the second part, as I have said it is just something cool to have but not essential. :shipit:

budzanowski (approved):

PR #285, opened 2021-12-01
Rename `test-unit` task to `test:js`

tomalec:

- Rename test-unit task to test:js, 
	for consistency and to mismatch with `test:php` unit tests.
- Reorder npm scripts alphabetically + pre/post.

1. `npm run test:js`
2. review the [package.json](https://github.com/woocommerce/pinterest-for-woocommerce/blob/update/npm-scripts/package.json#L101-L117) source

### DIFF
--- a/.github/workflows/js-unit-tests.yml
+++ b/.github/workflows/js-unit-tests.yml
@@ -19,4 +19,4 @@ jobs:
         uses: ./.github/actions/prepare-node
 
       - name: Run JavaScript unit tests
-        run: npm run test-unit
+        run: npm run test:js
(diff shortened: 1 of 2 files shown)


puntope:
> 1. npm run test:js

* Makes sense, I like this more than "test-unit".  👌 
* Tests work

💅 

However, I would love to have also consistency between projects in mind:

https://github.com/woocommerce/google-listings-and-ads/blob/develop/package.json

That allows to get the advantage for example of Ctrl+R "test" and then run always the same command. Or just to don't have a mess every time I'm running tests on different projects trying to remember if it was npm test-js, npm test or npm test-unit 

I'm ok to change it as well in GLA.

puntope (approved):

PR #286, opened 2021-12-02
Tweak the UI of claim website

### LINKED ISSUE #110: Woo site domain formatted like copy text in onboarding
### Describe the bug:
Screen Shot 2021-07-29 at 3 55 21 PM

In the `Claim your website` onboarding step, the site url/domain has no special formatting, which looks a little strange.

@shivanarrthine Can you advise how this should look? I'm thinking the type/style should be slightly different from the regular copy in some way.

### Steps to reproduce:
1. Fresh woo site & pinterest account – ensure the domain has not been claimed.
2. Install Pinterest for WooCommerce and start onboarding. Note this was observed on the branch for PR #108 (may be on `develop` too).

eason9487:

Closes #110

- Adjust the UI of website URL and error message block for the `<ClaimWebsite>`.
- Combine the processing status of claiming website to the verification button for the `<ClaimWebsite>`.
- Add JSDoc, simplify variable and replace hardcoded strings with constants for the `<ClaimWebsite>`.

#### 🎥 . Claim website success

Kapture 2021-12-02 at 11 03 24

#### 🎥 . Claim website failure

Kapture 2021-12-02 at 12 00 09

#### Prepare

If you've verified website before and want to redo the claim website process for testing, the preparing steps are:

1.  Go to https://www.pinterest.com/settings/claim and click the "Unclaim" button.
2. This plugin has cached data in the `pinterest_for_woocommerce` option of `wp_options` DB table, so we need to reset the status as well. You could update it by executing the below script on the DevTool Console:

```js
wp.apiFetch({path: '/wc-admin/options?options=pinterest_for_woocommerce'}).then((data) => {
  data.pinterest_for_woocommerce.account_data.verified_domains.length = 0;
  wp.apiFetch({
    path: '/wc-admin/options',
    method: 'POST',
    data,
  });
});
```

#### Test with a publicly accessible website

1. Go the onboarding setup: `/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Fonboarding&step=claim-website`
1. Before verification, the "Continue" button should be disabled.
1. Click the "Start verification" button.
1. Check if the button is displayed the correct processing status.
1. After successful verification, the "Continue" button should be enabled.
1. Refresh current page, it should show the same verified status as the previous step.
1. Click the "Continue" button and it should bring you to step 3.

#### Test with a local website

1. To test the success/failure result, it could be mocked by adding codes at line 95 in [ClaimWebsite.js](https://github.com/woocommerce/pinterest-for-woocommerce/blob/2b407c4de259d822f9b82d57b7bd20ecf9570826/assets/source/setup-guide/app/steps/ClaimWebsite.js#L95
):
    ```js
    // Simulate success result
    await new Promise( ( r ) => setTimeout( r, 1500 ) );
    setStatus( STATUS.SUCCESS );
    return;

    // Simulate failure result
    await new Promise( ( r ) => setTimeout( r, 1500 ) );
    throw {
    	data: { pinterest_code: 75 },
    	message: 'You may not have permission to this domain, or an error might have occurred. Please try again.',
    };
    ```
1. Go the onboarding setup: `/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Fonboarding&step=claim-website`
1. Before verification, the "Continue" button should be disabled.
1. Click the "Start verification" button.
1. Check if the button is displayed the correct processing status.
1. Check if the error message is displayed as expected.
1. After successful verification, the "Continue" button should be enabled. Click on it should bring you to step 3.

#### Check the Connection tab

The `<ClaimWebsite>` is also used in the Connection tab, and the adjusted logic associated with this page is the "Continue" button, which should not appear here.

1. Go to `/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Fconnection`
2. Check if "Continue" button is not shown.

### Changelog entry

> Tweak - Adjust the UI of claim website.


assets/source/setup-guide/app/constants.js
```diff
@@ -0,0 +1,22 @@
...
+ * Enum of general process status.
+ *
+ * @readonly
+ * @enum {string}
+ */
+export const PROCESS_STATUS = Object.freeze( {
+	...LABEL_STATUS,
+	IDLE: 'idle',
+	ERROR: 'error',
+} );
```
ecgan, line 22: 💅  I feel like we shouldn't have `LABEL_STATUS`, it is used in `if` condition check in `VerifyButton` which feels not very straightforward and readable.

I would prefer we have `PROCESS_STATUS` only like this: 

```suggestion
/**
 * Enum of general process status.
 *
 * @readonly
 * @enum {string}
 */
export const PROCESS_STATUS = Object.freeze( {
	IDLE: 'idle',
	PENDING: 'pending',
	SUCCESS: 'success',
	ERROR: 'error',
} );
```

Note that the line order also indicate the sequence: it starts at idle, and then pending, and then success or error, which IMO is easy to understand and less mental overhead.

assets/source/setup-guide/app/components/StatusLabel/index.js
```diff
@@ -0,0 +1,45 @@
...
+}
+
+/**
+ * Renders a text with a prefix graphic marking.
+ *
+ * @param {Object} props React props.
+ * @param {import('../../constants').LABEL_STATUS} props.status Label status.
+ * @param {string} props.text Label text.
+ */
+export default function StatusLabel( { status, text } ) {
```
ecgan, line 37: 💅  `StatusLabel` is used inside `VerifyButton`. There is a bit of overlap between both of them, as in they both look at the `status` and then decide what to render.

I'm thinking it would be better to make this `StatusLabel` more of a layout component and doesn't care about status, e.g. a `IconLabel` that accepts `icon` and `text`. Then, the handling the status icon and text can be put inside `VerifyButton` component. With this, only `VerifyButton` will look at `status`.

assets/source/setup-guide/app/components/UrlInputControl/index.js
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Renders a <InputControl> with a link icon prefix.
+ *
+ * @param {Object} props React props.
+ * @param {string} [props.className] Additional CSS class name to be appended.
+ * @param {...*} [props.restProps] Props to be forwarded to the InputControl.
+ *
+ * @see module:@wordpress/components#__experimentalInputControl
+ */
+export default function UrlInputControl( { className, ...restProps } ) {
```
ecgan, line 22: 💅  This is surprisingly similar to https://github.com/woocommerce/google-listings-and-ads/blob/9e7066e78ceac80971f228000e8a187c3add2b0d/js/src/components/app-input-link-control/index.js 😆 , which makes me wonder if we can somehow share the code and reuse it, or just make an exact copy of the files and make the file directory structure similar (e.g. move this component up to the `assets/source/components` folder and put a code comment referencing the one in Google Listings and Ads).

assets/source/setup-guide/app/steps/ClaimWebsite.js
```diff
@@ -35,47 +43,69 @@ const StaticError = ( { reqError } ) => {
...
-			setStatus( 'success' );
+			setStatus( STATUS.SUCCESS );
 		} catch ( error ) {
-			setStatus( 'error' );
+			setStatus( STATUS.ERROR );
 			setReqError( error );
 
 			createNotice(
-				'error',
+				STATUS.ERROR,
```
ecgan: 💅  For this `createNotice`, I'd say we should use `'error'` instead of `STATUS.ERROR`. By using `STATUS.ERROR`, it gives a false impression that `createNotice` can work with any other `STATUS`.

```suggestion
				'error',
```

assets/source/setup-guide/app/steps/ClaimWebsite.js
```diff
@@ -85,24 +115,25 @@ const ClaimWebsite = ( { goToNextStep, view } ) => {
...
+			[ STATUS.PENDING ]: __( 'Verifying…', 'pinterest-for-woocommerce' ),
+			[ STATUS.ERROR ]: __( 'Try again', 'pinterest-for-woocommerce' ),
+			[ STATUS.SUCCESS ]: __( 'Verified', 'pinterest-for-woocommerce' ),
 		};
 
+		const text = buttonLabels[ status ];
+
+		if ( Object.values( LABEL_STATUS ).includes( status ) ) {
+			return <StatusLabel status={ status } text={ text } />;
+		}
```
ecgan, line 133: 💅  Related to my other comment about changing StatusLabel to IconLabel, this here would then become something like this:

```js
if ( status === STATUS.PENDING ) {
	return (
		<IconLabel icon={<Spinner />} label={__( 'Verifying…', 'pinterest-for-woocommerce' )}
	)
}

if ( status === STATUS.SUCCESS ) {
	return (
		<IconLabel icon={<Icon icon="yes-alt" />} label={__( 'Verified', 'pinterest-for-woocommerce' )}
	)
}
```

There is no passing down of `status`. `IconLabel` is responsible for the icon and label layout. It is also clearer to see what would be rendered by just looking at `VerifyButton`, without needing to go into `IconLabel` or `StatusLabel`.

I would also refactor this `VerifyButton` into a separate folder / file, so that it can have its own CSS file and its own imports at the top of the file.

assets/source/setup-guide/app/style.scss
```diff
@@ -345,40 +300,18 @@ $root: ".woocommerce-setup-guide";
...
-				}
-
-				#{$root} {
-
-					&__step {
-
-						&-status {
-							padding: 0;
-						}
+					> *:not(:first-child) {
```
ecgan: 💅  `*` is not necessary. 

```suggestion
					> :not(:last-child) {
```

assets/source/setup-guide/app/style.scss
```diff
@@ -345,40 +300,18 @@ $root: ".woocommerce-setup-guide";
...
-				}
-
-				.has-idle {
-
-					#{$root} {
-
-						&__step-status__icon {
-							display: none;
-						}
+					> *:first-child {
```
ecgan: 💅  `*` is not necessary.

```suggestion
					> :first-child {
```

assets/source/setup-guide/app/style.scss
```diff
@@ -345,40 +300,18 @@ $root: ".woocommerce-setup-guide";
...
+						margin-bottom: $grid-unit-10;
 					}
 				}
 
-				.claim-error {
-					color: var(--wc-red);
+				.components-notice.is-error {
+					margin: 0;
+					background-color: #f8ebea;
```
ecgan: 💅  I usually prefer to stick with the standard background color provided by the component so that things are more consistent, and avoid doing too much customization like this, but I guess that's how it is designed for this page. 😅

ecgan (approved):
Tested, working as expected. 👍 

I left quite a number of 💅  comments below. Most of them relate to the code structure and responsibilities around `StatusLabel` and `VerifyButton`. If you want to change it, you can change it in this PR, or open a separate code refactor PR. 

There are also two PHP file changes, which we might want to get a PHP expert to review. 😄

assets/source/setup-guide/app/style.scss
eason9487: Thanks! And I found that it should still select `:not(:last-child)`, I updated your code suggestion to fix it by the way.

assets/source/setup-guide/app/steps/ClaimWebsite.js
eason9487: Thanks for catching this! Yes, it shouldn't be changed. I changed it accidentally when I searched and replaced these hardcoded strings with editor. 🤦‍♂️

assets/source/setup-guide/app/style.scss
eason9487: image

The first one is the default background color of `<Notice>` component. Since the difference is very obvious, I thought the customization was needed here.

assets/source/setup-guide/app/components/UrlInputControl/index.js
eason9487: > This is surprisingly similar to [...], which makes me wonder if we can somehow share the code and reuse it

Yeah, I copied from there. I think this should be a 📜  or an even wider topic if want to share and reuse them across our repos.

> or just make an exact copy of the files and make the file directory structure similar (e.g. move this component up to the `assets/source/components` folder and put a code comment referencing the one in Google Listings and Ads).

I have checked other design change tasks during development, and this URL link component is the only one used here, so I think it's reasonable to put it here according to the current directory structure. In the future, if there's a wider range of reuse or directory structure adjustment, we could move it at then.

I'm not sure if I need to add references between two or more repos. For example, sometimes I refer to other code implementations owned by WooCommerce organization, but unless there's something in it that needs special attention or there're some issues, I won't add a reference to the source.

Do you think there is some purpose missing here that I didn't notice and need to add a code comment to reference the source? Or is there a licensing issue?

assets/source/setup-guide/app/constants.js
eason9487: > I feel like we shouldn't have `LABEL_STATUS`, it is used in `if` condition check in `VerifyButton` which feels not very straightforward and readable.

`LABEL_STATUS` is also used to indicate the `status` prop that it accepts in the `<StatusLabel>` component and its JSDoc. And `PROCESS_STATUS` is not the status set accepted by this component.

https://github.com/woocommerce/pinterest-for-woocommerce/blob/70205b028a0c460753f32ea58057ef5e3dbe1b75/assets/source/setup-guide/app/components/StatusLabel/index.js#L30-L37

I tried back and forth to get `@enum` to be more flexible with exporting/importing references, but it doesn't seem to work. For example, the following way would make the type of `status` appear as `any | any`

```js
/**
 * @typedef {import('../../constants').PROCESS_STATUS} PROCESS_STATUS
 */
/**
 * @param {Object} props React props.
 * @param {PROCESS_STATUS.PENDING|PROCESS_STATUS.SUCCESS} props.status Label status.
 */
```

image

---

> Note that the line order also indicate the sequence: it starts at idle, and then pending, and then success or error, which IMO is easy to understand and less mental overhead.

They have finite-state machine relation, in which the order of states does not always switch according to the above arrangement. I don't think that the finite-state machine relation should be implicitly stated by the order of code appearance to indicate the order of its states. Because it may be only one (happy path) of the state transition paths, but not all of them.

assets/source/setup-guide/app/components/StatusLabel/index.js
eason9487: > I'm thinking it would be better to make this `StatusLabel` more of a layout component and doesn't care about status, e.g. a `IconLabel` that accepts `icon` and `text`.

It's not only a layout component. There's [high dependent styling](https://github.com/woocommerce/pinterest-for-woocommerce/blob/70205b028a0c460753f32ea58057ef5e3dbe1b75/assets/source/setup-guide/app/components/StatusLabel/style.scss) that depends on the status. With the proposed `<IconLabel>` which only takes care of layout, then those status-related styles will be placed elsewhere.  Like the Overview of Catalog tab, these future reusable styles are [scattered around outside](https://github.com/woocommerce/pinterest-for-woocommerce/blob/70205b028a0c460753f32ea58057ef5e3dbe1b75/assets/source/setup-guide/app/style.scss#L618-L643).

image

Rather than combining status icons and layout components, and setting the status icon and additional layout styles for the combined React elements in some external places, I think it is better to put highly dependent and reusable styles into a dedicated component.

The `<StatusLabel>` component and its `status` prop are similar in concept to `<Notice>` of `wp-components`, and even the *status icon + text* usages in **Overview** above could be migrated to use `<StatusLabel>` in the future.

assets/source/setup-guide/app/steps/ClaimWebsite.js
eason9487: Same as https://github.com/woocommerce/pinterest-for-woocommerce/pull/286#discussion_r763646812. And `<StatusLabel>` should already describe what will be rendered via its JSDoc. I'm happy to adjust it if not clear enough.

assets/source/setup-guide/app/style.scss
findingsimple: @shivanarrthine what do you think? Should we just stick with the default background color of the component?

assets/source/setup-guide/app/components/UrlInputControl/index.js
findingsimple: > I think this should be a 📜 or an even wider topic if want to share and reuse them across our repos.

I think that would be a neat discussion/p2 to have

> Do you think there is some purpose missing here that I didn't notice and need to add a code comment to reference the source? Or is there a licensing issue?

I don't think we should get hung up on this, I'm aligned with...

> I think it's reasonable to put it here according to the current directory structure. In the future, if there's a wider range of reuse or directory structure adjustment, we could move it at then.

assets/source/setup-guide/app/components/StatusLabel/index.js
ecgan: > There's [high dependent styling](https://github.com/woocommerce/pinterest-for-woocommerce/blob/70205b028a0c460753f32ea58057ef5e3dbe1b75/assets/source/setup-guide/app/components/StatusLabel/style.scss) that depends on the status. 

Yeah I noticed the styling. 🙂 

> With the proposed `<IconLabel>` which only takes care of layout, then those status-related styles will be placed elsewhere.

Yes, it is related to https://github.com/woocommerce/pinterest-for-woocommerce/pull/286#discussion_r763011970, my thinking is that those status-related style would be defined in the parent level `VerifyButton`, the parent would dictate what to render, including the styling.

> Like the Overview of Catalog tab, these future reusable styles are [scattered around outside](https://github.com/woocommerce/pinterest-for-woocommerce/blob/70205b028a0c460753f32ea58057ef5e3dbe1b75/assets/source/setup-guide/app/style.scss#L618-L643).

I didn't see them as "scattered around outside" though. I think it is fine that the parent decide what to render and how it is styled. 

Here's an example of my thinking: Let's say that we have a defined yellow warning color, like the screenshot you provided. There _might_ be places where we want to have a different warning color in different pages, and in those cases we probably gonna need a way (React props, or CSS) for the page or parent component to override the defined default yellow warning color anyway. The defined yellow warning color serves as a good convenient consistent default; the parent should have a way to dictate and override the default. 

About "future reusable styles", yeah that is a good point. We can certainly refactor those styles and make them more reusable. 😄 👍

assets/source/setup-guide/app/constants.js
ecgan: > > I feel like we shouldn't have `LABEL_STATUS`, it is used in `if` condition check in `VerifyButton` which feels not very straightforward and readable.
> 
> `LABEL_STATUS` is also used to indicate the `status` prop that it accepts in the `<StatusLabel>` component and its JSDoc.
> 

Sorry, I should have been clearer 😅 My comment here was related to https://github.com/woocommerce/pinterest-for-woocommerce/pull/286#discussion_r762906837 and https://github.com/woocommerce/pinterest-for-woocommerce/pull/286#discussion_r763011970 (changing `StatusLabel` into `IconLabel`). With those changes, we wouldn't need `LABEL_STATUS`.

assets/source/setup-guide/app/style.scss
shivanarrthine: Thanks for raising this. Let's stick to using the default background color for the component.

eason9487: Updated by https://github.com/woocommerce/pinterest-for-woocommerce/pull/286/commits/d8a65906258a6bb0a4849ac7b829b94cb7ad9751.

image

ianlin (approved):
Reviewed two PHP files, LGTM. 👍 

✅  Do not use `wp_parse_url( home_url(), PHP_URL_HOST )` in order to get the url with its protocol.
✅  Remove the fixed text `Your domain could not be automatically verified.` from the backend to match the new design of showing error messages.

PR #290, opened 2021-12-03
Fire track event for modal

tomalec:

- Support multiline Track Event properties' description.
- Fire track event for modal

implements part of #232.

### Screenshots:

modal open
modal event

0. Enable tracks logging, by `localStorage.setItem('debug', 'wc-admin*')`
1. Go to [`/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Fonboarding`](https://wp581wc580.test/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Fonboarding) 
2. (Connect account)
3. Click on "disconnect"
	- you should get `wcadmin_pfw_modal_open: { context: 'account-disconnection' }`
4. Click on overlay | "x" | "cancel" | press <kbd>Esc</kbd>
	- you should get `wcadmin_pfw_modal_closed: { context: 'account-disconnection', action: 'dismiss' }`
5. Reopen, click "Yes, I'm sure"
	- you should get `wcadmin_pfw_modal_closed: { context: 'account-disconnection', action: 'confirm' }`

(no changelog entry, we will add a single one for all tracks)

### DIFF
--- a/assets/source/setup-guide/app/components/Account/Connection.js
+++ b/assets/source/setup-guide/app/components/Account/Connection.js
@@ -56,7 +56,7 @@ const PinterestLogo = () => {
  * @event wcadmin_pfw_modal_closed
  * @property {string} context Indicate which modal this is for.
  * @property {string} action
- * 				`confirm` - When the final "got it" button is clicked.
+ * 				`confirm` - When the final "Yes, I'm sure" button is clicked.
  * 				`dismiss` -  When the modal is dismissed by clicking on "x", "cancel", overlay, or by pressing a keystroke.
  */
 
--- a/assets/source/setup-guide/app/components/Account/Connection.js
+++ b/assets/source/setup-guide/app/components/Account/Connection.js
@@ -48,13 +48,15 @@ const PinterestLogo = () => {
  * Opening a modal.
  *
  * @event wcadmin_pfw_modal_open
- * @property {string} context Indicate which modal this is for.
+ * @property {string} name Which modal is it?
+ * @property {string} context `'settings' | 'wizard'` In which context it was used?
  */
 /**
  * Closing a modal.
  *
  * @event wcadmin_pfw_modal_closed
- * @property {string} context Indicate which modal this is for.
+ * @property {string} name Which modal is it?
+ * @property {string} context `'settings' | 'wizard'` In which context it was used?
  * @property {string} action
  * 				`confirm` - When the final "Yes, I'm sure" button is clicked.
  * 				`dismiss` -  When the modal is dismissed by clicking on "x", "cancel", overlay, or by pressing a keystroke.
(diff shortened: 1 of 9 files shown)


assets/source/setup-guide/app/components/Account/Connection.js
```diff
@@ -44,12 +44,29 @@ const PinterestLogo = () => {
...
+ * @event wcadmin_pfw_modal_open
+ * @property {string} context Indicate which modal this is for.
+ */
+/**
+ * Closing a modal.
+ *
+ * @event wcadmin_pfw_modal_closed
+ * @property {string} context Indicate which modal this is for.
+ * @property {string} action
+ * 				`confirm` - When the final "got it" button is clicked.
```
eason9487: 💅  Looks like it should be the "Yes, I'm sure" button.

woo-tracking-jsdoc/publish.js
```diff
@@ -80,7 +80,12 @@ exports.publish = function ( data ) {
 						/\|/g,
 						'\\|'
 					);
-					mdResult += `\`${ property.name }\` | \`${ type }\` | ${ description }\n`;
+					mdResult += `\`${
+						property.name
+					}\` | \`${ type }\` | ${ description.replace(
+						/\s*\n\s*/,
```
eason9487: 💅 
How about adding the `g` flag in case of the `description` has more than one newline char?

eason9487 (commented):
Tested locally. All events were triggered as expected.

But there's a ❓ or ❔  concern. The upper-layer component `<SetupAccount>`, which uses `<AccountConnection>`, is also used in the Connection tab: `/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Fconnection`.

image

Do we need to make it easier to distinguish which page the model event was triggered in the event tracking data? And just noticed the tracking events `pfw_account_disconnect_button_click` of disconnecting button have the same concern.

tomalec:
Regarding 
> But there's a :question: or :grey_question: concern. The upper-layer component `<SetupAccount>`, which uses `<AccountConnection>`, is also used in the Connection tab: `/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Fconnection`.

I'm not 100% how to approach this
[Track Events spreadsheet](https://docs.google.com/spreadsheets/d/1u25mOcsODgcdPMy_rIagUAWHCfzsvJI1e8UPrJvYtVE/edit#gid=903716843) states: 
> I17: Context is used to indicate which modal this is for

If the question is "Which modal", one may say, that on both pages this is the same "Account Disconnection" modal, exactly same content, same action, same purpose. Just opened from a different location.

So, IMHO, we are back into `path` vs `context` discussion.

tomalec:
On the other hand. The event name is `modal_open`, not `disconnect_modal_open`. If it's "Q: _what modal?_, A: _Account Disconnection_", then the property should be more like a `name` or `type`.
If it's `context` it rather matches the "Q: In what context this exact modal was opened?".

Now, given in the entire Plugin, we have a single **type/name/kind** of modal - the Account Disconnection modal. Which is used on two different pages/locations/context. I'll change the event description (and implementation), to something like:
```
 * @event wcadmin_pfw_modal_open 
 * Fired when any modal is open. 
 * @property {string} name What modal is it? // To be future prepared for other kinds of modals
 * @property {string} context Indicate in what context it was opened.  // IMHO, We could use `path` instead, but let's leave it as it was for now.
```

tomalec:
@eason9487  Thanks for the review, I addressed all your comments, I believe it's ready for another round.

tomalec:
I'm not 100% sure about the tests from https://github.com/woocommerce/pinterest-for-woocommerce/pull/290/commits/be2191e77ea126295303a3544a8f1c6483bfb1ba, as that's actually testing the `<Modal>` features, maybe I should rather stub `<Modal>` and check that we fire an event whenever it's dismissed? WDYT?

eason9487 (approved):
LGTM. Thanks for the elaborate thoughts!

I also think it's more reasonable to use `name` to indicate which modal it is, and use `context` first to indicate the modal is opened for/from what process/place. About whether there're more consistent and suitable properties, would be happy to discuss later.

> I'm not 100% sure about the tests from [be2191e](https://github.com/woocommerce/pinterest-for-woocommerce/commit/be2191e77ea126295303a3544a8f1c6483bfb1ba), as that's actually testing the `<Modal>` features, maybe I should rather stub `<Modal>` and check that we fire an event whenever it's dismissed? WDYT?

I think it's ok, after all, the tracking events are tied to the modal behaviors. 😂

PR #293, opened 2021-12-08
Fix/228 wrong feed registration status multiple feed profiles

### LINKED ISSUE #228: Feed registration status is incorrect when user has more than one feed profile
Steps to reproduce:
1. Create a feed profile from https://www.pinterest.com/product-catalogs/ directly and then disable it
2. Register feed profile from Woocommerce plugin to same user

Issue happened on build tagged as beta.4

image

3. You will see error under Woo plugin _product feed not active_ even if Woocommerce feed profile is active
image

Reason:

Feed profile to check registration status is not picked correctly. In https://github.com/woocommerce/pinterest-for-woocommerce/pull/182/files feed registration logic was changed to fix if user has multiple feed profile by saving feed profile id in _feed_registered_ data variable but this variable is not checked when getting feed profile registration state in _add_feed_registration_state_

Fix:

We need to fetch feed profile based on _feed_registered_ data variable and use that instead of $merchant['data']->product_pin_feed_profile in add_feed_registration_state

Additional Details:
I checked current method in `get_merchant_feed` is actually fetching feed report but feed profile. To fetch feed profile you need to use following api `https://api.pinterest.com/v3/catalogs/<merchant_id>/feed_profiles/` to fetch all feed profiles and then filter feed profile based on feed profile id from _feed_registered_ data variable. Currently there is no api for fetching feed profile by feed profile id. We are planning to add that and will notify when it is available.

ksere:

Closes #228 .

When a merchant has multiple feeds, we should get the feed report, in order to populate the status of the feed, using the stored feed_id. 

### Screenshots:
image

1. Create a feed profile using https://www.pinterest.com/product-catalogs/ directly and then disable it.
2. Install & finish onboarding of the plugin so that a new feed profile is registered by the plugin.
3. Verify that the status shown in the product catalog page is of the expected feed (the one created by the plugin).

### Changelog entry

> Fix - Get feed report for the stored Feed ID instead for the first feed returned by the API.
> Add - get_merchant_feed_report() method that fetches the feed_report
> Tweak - get_merchant_feed() now fetches the feed instead of the feed_report.

### DIFF
--- a/src/API/Base.php
+++ b/src/API/Base.php
@@ -465,7 +465,7 @@ class Base {
 	 * @param string $merchant_id The merchant ID the feed belongs to.
 	 * @param string $feed_id     The ID of the feed.
 	 *
-	 * @return mixed
+	 * @return object
 	 *
 	 * @throws \Exception PHP Exception.
 	 */
--- a/src/API/FeedState.php
+++ b/src/API/FeedState.php
@@ -264,10 +264,6 @@ class FeedState extends VendorAPI {
 
 			$feed = Base::get_merchant_feed( $merchant_id, $feed_id );
 
-			if ( ! $feed ) {
-				throw new \Exception( esc_html__( 'Could not get feed info.', 'pinterest-for-woocommerce' ) );
-			}
-
 			if ( 'ACTIVE' !== $feed->feed_status ) {
 				throw new \Exception( esc_html__( 'Product feed not active.', 'pinterest-for-woocommerce' ) );
 			}


findingsimple:
Thanks @ksere and @alexiglesias31 

I've requested a review from Ventures

src/API/Base.php
```diff
@@ -460,14 +460,74 @@ public static function update_merchant_feed( $merchant_id, $feed_id, $args ) {
 
 
 	/**
-	 * Get a specific merchant's feed using the given arguments.
+	 * Get a specific merchant feed using the given arguments.
 	 *
 	 * @param string $merchant_id The merchant ID the feed belongs to.
 	 * @param string $feed_id     The ID of the feed.
 	 *
 	 * @return mixed
```
JPry: This `@return` should be updated to match the structure of `$feed_profile`. That's the only possible return value, as everything else results in a thrown exception.

src/API/FeedState.php
```diff
@@ -261,7 +262,13 @@ private function add_feed_registration_state( $result ) {
 				throw new \Exception( esc_html__( 'Could not get merchant info.', 'pinterest-for-woocommerce' ) );
 			}
 
-			if ( 'ACTIVE' !== $merchant['data']->product_pin_feed_profile->feed_status ) {
+			$feed = Base::get_merchant_feed( $merchant_id, $feed_id );
+
+			if ( ! $feed ) {
+				throw new \Exception( esc_html__( 'Could not get feed info.', 'pinterest-for-woocommerce' ) );
+			}
```
JPry: This check is unnecessary, as the `get_merchant_feed()` method can only return a valid profile or throw an exception.

```suggestion
```

JPry (changes requested):

JPry (approved):
Sorry for the delay, this dropped off my radar to look at again. The changes look good to me 👍🏻

findingsimple:
Just need to get the 🟢 on JS linting

ecgan:
> Just need to get the 🟢 on JS linting

I looked into the [failed JavaScript and CSS Linting check](https://github.com/woocommerce/pinterest-for-woocommerce/runs/4877694888?check_suite_focus=true). It fails because this PR is coming from a public fork, and therefore the `GITHUB_TOKEN` has only read permissions (see screenshot below). Other PRs within this repo (not a fork) does not have this problem, as the `GITHUB_TOKEN` in those PR has write permissions.

image

References:

- [Automatic token authentication](https://docs.github.com/en/actions/security-guides/automatic-token-authentication)
- [GitHub Actions: Control permissions for GITHUB_TOKEN](https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/)

---

@ksere , heads up: there is a conflict in your PR here.

budzanowski:
This just needs merge conflicts removed.

Issue #292, opened 2021-12-07
Create action to generate a weekly build


JPry:
### Describe the solution you'd like 
Create a GitHub action to automatically generate a build on a regular recurring basis.

budzanowski:
We will likely not be using this.

PR #295, opened 2021-12-09
Track Save Button Click 

puntope:

Closes part of #232 

save_changes_button_click | Triggered when save changes button is clicked https://d.pr/i/4CiZgY | context: e.g. pinterest_settings or pinterest_connectionindication of what is being saved/changed 
-- | -- | -- 

0. Open the console and run `localStorage.setItem('debug', 'wc-admin*')` (be sure you have verbose mode activated)
1. With a completed onboarding go to Marketing -> Pinterest -> Settings Tab
2. Change some options and click save
3. You should see a debugging output showing the event `wcadmin_pfw_save_changes_button_click` with `pinterest_settings` in context prop and the options updated.
4. Change some options in Settings Tab, move to Connection Tab and save. 
5. You should see a debugging output showing the event `wcadmin_pfw_save_changes_button_click` with the properties updated. 
6. Check that reloading the page persists the saved options 
7. Check and run the unit tests with `npm run test:js` 

### Technical details

This follows the same strategy as in Analytics -> Settings -> Save. We send the context and the properties to be saved. Doesn't matter if they are actually saved in DB or not.

We only allow to track the properties registered in `TRACKABLE_SETTINGS` inside `prepare-for-tracking.js` this is for preventing `account_info` or any other current and future sensitive data to be tracked by accident. 

Regarding to context; I'm triggering `pinterest_settings` for both Settings and Connection. Since there is an issue with naming there. I'm not sure if we should add both  `pinterest_settings` and  `pinterest_connection` or only  `pinterest_settings`

### Changelog entry

> Tweak - Add tracking on Save Changes button


assets/source/setup-guide/app/components/SaveSettingsButton.js
```diff
@@ -12,13 +13,33 @@ import {
...
+	const updatedData = useSettingsSelect( 'getUpdatedData' );
 	const setAppSettings = useSettingsDispatch( true );
 	const createNotice = useCreateNotice();
 
 	const saveSettings = async () => {
+		recordEvent( 'pfw_save_changes_button_click', {
+			updatedData,
+			context: getContextByPath(),
+		} );
+
```
ianlin: ❓ I'm thinking whether we need to send this event after successfully saving data to the DB? I know the event name is `pfw_save_changes_button_click` meaning that it's a click event, just wondering which makes more sense from the perspective of people who will grab all the events for analysing.

assets/source/setup-guide/app/helpers/get-context-by-path.js
```diff
@@ -0,0 +1,19 @@
...
+ * @throws Exception if getPath is not working, returns an empty string as a result
+ */
+export default function getContextByPath() {
+	try {
+		const path = getPath();
+		return path.substring( 1 ).replace( /\//g, '_' );
+	} catch ( e ) {
+		return '';
+	}
+}
```
ianlin: 💅  I'd suggest passing explicit context from the components that uses `SaveSettingsButton` component. Because if the path has changed someday (tho very unlikely), the context that is sent for event tracking would be changed as well, then the data analysis people need to change their filtering conditions.

assets/source/setup-guide/app/data/settings/reducer.js
```diff
@@ -42,6 +42,14 @@ const settingsReducer = (
...
 			};
 			break;
+		case TYPES.SET_UPDATED_DATA:
+			state = {
+				...state,
+				updatedData: action.reset
+					? {}
+					: { ...state.updatedData, ...action.data },
+			};
+			break;
```
ianlin: 💅  I noticed that there is no default value for `updatedData` in the settings reducer, and it brought the following behaviour (very minor):

- Go to `/wp-admin/admin.php?page=wc-admin&path=/pinterest/settings`
- Click `Save changes` straightaway
- From the dev console the `updatedData` is `undefined`
- Click `Save changes` again
- From the dev console the `updatedData` is an empty object `{}`

When the data/analysis people try to filter out how many clicks that has no data change, the filter condition becomes inconsistent.

ianlin (commented):
Tested locally and it works as expected 👍 

✅  Tried to use the `develop` branch, no events were sent. Checkout to this branch, the events are sent correctly.
✅  The event name `wcadmin_pfw_save_changes_button_click ` was sent correctly.
✅  The context names were sent correctly for both `Settings` and `Connection` tab.
✅  The behaviour of changing options in `Settings` tab and click `Save changes` in `Connection` tab works perfectly.
✅  Can see the updated changes after reloading the page .

I left some comments below that might need some discussions, please take it a look and let me know what are your thoughts. 😃

assets/source/setup-guide/app/components/SaveSettingsButton.js
puntope: Good question, seems that we are tracking this in [Woo Tracker Snapshot implementation](https://github.com/woocommerce/google-listings-and-ads/blob/develop/src/Tracking/TrackerSnapshot.php): 

Basically we save there the settings the user has in the plugin. Which is a separated issue I'm working on. 

This track is just for track the attempt of saving doesn't matter if it's actually saved or not.

assets/source/setup-guide/app/data/settings/reducer.js
puntope: Amazing catch! 🕵️ 

Fixed in [429af73](https://github.com/woocommerce/pinterest-for-woocommerce/pull/295/commits/429af73ca403453f802fed7d34e43d44b784dac4)

assets/source/setup-guide/app/helpers/get-context-by-path.js
puntope: Interesting observation. I'm unsure about this one. 

For example, let's suppose it's manually hardcoded in the record event.

- How can we differentiate between Connection/Settings tab?  
- Let's imagine we named the event as it's right now: "pinterest_settings". If we change the path to "pinterest_config" and we still send "pinterest_settings" as context creates also a naming issue/confusion and Analytics needs to remember that as well. To consider also that having a different  name would be somehow helpful to identify changes in the user behaviour after the path changes.   
- Another advantage of this is to avoid issues in naming from the devs. For example, typos, or different naming ideas, `pinterest_wizard`, `pinterest_landing`, `pinterest-getting-started` ...etc  (I saw this inconsistencies already in the current code) 

But again, I'm completely unsure about this.

💅  Suggestion: 

We can keep this from now since, as you said, it's going to be unlikely that the path changes and it's only having impact in that save button (for now) plus the fact that there is not any candidate to differentiate between the calls from Connection Tab and Settings Tab, besides relaying on the path, or sending the view name as a prop from the parent component inside SaveSettingsButton.

We can open a further discussion in the parent #292  to confirm this

Thoughts?

assets/source/setup-guide/app/components/SaveSettingsButton.js
ianlin: Cool, thanks for sharing this:  [Woo Tracker Snapshot implementation](https://github.com/woocommerce/google-listings-and-ads/blob/develop/src/Tracking/TrackerSnapshot.php). Agree with this track is just for clicking and good to know we will be having the track of saving soon. 👍

assets/source/setup-guide/app/helpers/get-context-by-path.js
ianlin: Thanks for the detailed explanation! Agree with the three points you listed especially point two: keep the context identical with the path/page names would be better in this case.

> We can open a further discussion in the parent #292 to confirm this

I think you meant #232  😄 Sure, I will leave a comment there, for now it's perfect to merge.

ianlin (approved):
Good to go! 👍

assets/source/setup-guide/app/components/SaveSettingsButton.js
```diff
@@ -12,13 +13,33 @@ import {
...
+ */
 const SaveSettingsButton = () => {
 	const isSaving = useSettingsSelect( 'isSettingsUpdating' );
+	const updatedData = useSettingsSelect( 'getUpdatedData' );
 	const setAppSettings = useSettingsDispatch( true );
 	const createNotice = useCreateNotice();
 
 	const saveSettings = async () => {
+		recordEvent( 'pfw_save_changes_button_click', {
+			updatedData,
```
eason9487: 🚧  
I think the naming and value meaning of `updatedData` are confusing.

From the naming perspective, the event itself tracks the attempt of clicking the save button, but it doesn't matter if the settings are successfully saved or not. The **updated** in naming is easily interpreted to mean that the data update is complete/successful.

From the value meaning perspective, no matter each property appears in the `updatedData` or its value is `true`/`false`, they don't have a stable meaning. For example, if `updatedData` has

- `enable_debug_logging` property, it still couldn't tell where the `enable_debug_logging` is actually what value, because the event is called regardless of the API request is successful or not.
- `enable_debug_logging` property with `true` or `false` value, even if we ignore whether the data is successfully saved, it doesn't mean that the setting is "updated" to `true`/`false` at that moment.
    If a user checks and then unchecks the same setting, that setting will appear in `updatedData` with the same value against the corresponding value in DB. This means it not always could be interpreted to mean "a setting is updated to true/false", it could be no change.

    https://user-images.githubusercontent.com/17420811/146323294-793b8140-6f04-4664-9642-0f847933fafc.mp4

    (The `false` in the API response `pinterest_for_woocommerce: false` means that no update occurred in DB.)

I would suggest reconsidering, suppose as a data analyst, whether there is a known scenario that is analyzed by the clicking event of the "Save" button and group the settings data to be updated (but not necessarily actually be updated successfully) at that moment?

- If no, would it be proper to remove the `updatedData` at all?
- If yes, it'd be better to make the naming and the value meaning stable and clear.

assets/source/setup-guide/app/helpers/get-context-by-path.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * External dependencies
+ */
+import { getPath } from '@woocommerce/navigation';
+
+/**
+ * Returns the formatted context based on the path.
+ *
+ * @return {string} The path in snake case format
+ * @throws Exception if getPath is not working, returns an empty string as a result
```
eason9487: ❓ 
This function won't throw an exception.

ianlin (commented):
🚧 

Sorry for leaving this comment after I approved the PR 😅 

Thanks @jconroy for providing the info in `p1639630485065000-slack-C02B84SRR5M`, we do have a tracking naming conventions guide in an internal document.

From the document it says
> Don’t programmatically create event names. They should be hardcoded so they are easily searchable in the code base.

While there is no rule for context name (since it's a property in the event payload), it would be good to treat it similar to an event name for consistency.

eason9487 (changes requested):
Sorry, I think there are two concerns that need to be addressed.

1. 🚧  The tracking event property doesn't support nested data structure. The `updatedData` will be serialized to `'[object Object]'` string.
    
2. 🚧  The *ClaimWebsite.js* also uses `setAppSettings()`, which calls `updateSettings()` internally, to update options.
    https://github.com/woocommerce/pinterest-for-woocommerce/blob/429af73ca403453f802fed7d34e43d44b784dac4/assets/source/setup-guide/app/steps/ClaimWebsite.js#L75-L82
    This makes it possible for users' `account_data` to be sent for tracking as well. I think it seems to be exposed to an unintended place. Although the data in `account_data` may not be sensitive for now, there could be potential problems with accidentally sending sensitive user data or even security-related data to the tracking endpoint in the future. Please take a look at the video below.

    https://user-images.githubusercontent.com/17420811/146332425-bc89d135-1f32-463d-af90-4d8ed66112d4.mp4

    Steps to reproduce:
    1. Finish onboarding setup.
    1. Refer to the part of test steps in #286 to unclaim website.
    1. Go to the Connection tab, claim website, save changes.

assets/source/setup-guide/app/components/SaveSettingsButton.js
puntope: > If yes, it'd be better to make the naming and the value meaning stable and clear.

What about, `dataToBeSaved`? Any suggestion for naming?

> I would suggest reconsidering, suppose as a data analyst, whether there is a known scenario that is analyzed by the clicking event of the "Save" button and group the settings data to be updated (but not necessarily actually be updated successfully) at that moment?

Thats a valid point, some thoughts: 

Could be considered as more accurate to register the property even if the user checks and then unchecks again the property. Which is more close to the real user action. The user attempted to save changes and manipulated that props and as analyst I want to know that.

On the other side, also could be possible to compare the `updatedData` values against the previous values and only update them if there is a change. 

> f no, would it be proper to remove the updatedData at all?

By removing  `updatedData` it would be needed to trigger the event based on save instead of click. Which as far as I now is not what we want to achieve.  

Opinions, thoughts?

assets/source/setup-guide/app/helpers/get-context-by-path.js
puntope: It's outdated. See comments in https://github.com/woocommerce/pinterest-for-woocommerce/pull/295#pullrequestreview-833736094 

However you're right, I guess at the beginning Iw as adding a throw and then I change my opinion and forgot to update the docs.

assets/source/setup-guide/app/components/SaveSettingsButton.js
puntope: > If yes, it'd be better to make the naming and the value meaning stable and clear.

IN regards this... Checking the code again, the current action events were called `updateSettings` and also there is other one named `isUpdating`  so, I took as granted that `update` is the appropriate verb.

assets/source/setup-guide/app/helpers/get-context-by-path.js
```diff
@@ -0,0 +1,19 @@
...
+ */
+export default function getContextByPath() {
```
tomalec: :question: / :construction: 
I don't really get the reason **why** we need this function?

So far for the `context` property, we were hard coding identifiers, so we can get more granular than `path`, and have multiple triggers on the same page and on the same page.
For the triggers that need path, we pass it as is `/with/stashes`.

IMHO, context constructed as path, introduces quite an inconsistency in our Tracks.

BTW, for the `path` property, I'd rather use routing path than the query location path. But that's a separate topic.

Also, the logic with returning `''` is unclear to me, when it may happen, what does it mean?

puntope:
> While there is no rule for context name (since it's a property in the event payload), it would be good to treat it similar to an event name for consistency.

Just discovered that `ConnectionApp` is named as `SettingsApp` as well, and the view names sent as 'settings' on both SettingsApp and ConnectionApp. 

Opened an issue #302 about this.

assets/source/setup-guide/app/helpers/get-context-by-path.js
puntope: Fixed in https://github.com/woocommerce/pinterest-for-woocommerce/pull/295/commits/e51fbfd87d6b2a2e016196ee9ae73a2248a399b3

> I don't really get the reason why we need this function?

The idea as the beginning was to use the path as a context builder to preserve consistency. [After discussion ](p1639630485065000-slack-C02B84SRR5M) we decided to manually hardcoded the names and the function was considered a bad idea at the end. 

> Also, the logic with returning '' is unclear to me, when it may happen, what does it mean?

Since `getPath` is an external function and `context` is a string,  I decided that, in case it fails,  we could return an empty string instead of `null`, `undefined` or `false`.

puntope:
> While there is no rule for context name (since it's a property in the event payload), it would be good to treat it similar to an event name for consistency.

Fixed in [e51fbfd](https://github.com/woocommerce/pinterest-for-woocommerce/pull/295/commits/e51fbfd87d6b2a2e016196ee9ae73a2248a399b3)

assets/source/setup-guide/app/helpers/get-context-by-path.js
puntope: See also https://github.com/woocommerce/pinterest-for-woocommerce/pull/295/commits/e51fbfd87d6b2a2e016196ee9ae73a2248a399b3 where I fixed all related to this function

assets/source/setup-guide/app/components/SaveSettingsButton.js
eason9487: > Could be considered as more accurate to register the property even if the user checks and then unchecks again the property. Which is more close to the real user action. The user attempted to save changes and manipulated that props and as analyst I want to know that.

This could be a possible scenario, I'm a bit confused about what insights could be figured out from it though. because it's only possible to know that the user has manipulated this option on UI. It might not possible to tell what the user intended, for example: 
- they forgot they had changed the option and then changed it back
- or they know they changed the option, but then changed their mind and restore the option back
- or they did want to and did change it (and have to find a way to compare it from recorded analysis data to the actual saved option value) 

> By removing `updatedData` it would be needed to trigger the event based on save instead of click. Which as far as I now is not what we want to achieve.

Whether you decide *yes* or *no*, the reason I suggested reconsidering was, even though the spreadsheet requirements list "indication of what is being saved/changed", you could still suspect/question whether it might have wrong assumptions and then correct it if so.

I was mainly concerned if the data recorded does not have a concrete and stable meaning, then it may not work for the intended scenario, or may further affect the analysis results and mislead to the wrong insights. So if you decide to send this tracking event with **settings** as event properties, it would be good to also use JSDoc, such as [wcadmin_pfw_get_started_faq](https://github.com/woocommerce/pinterest-for-woocommerce/blob/429af73ca403453f802fed7d34e43d44b784dac4/assets/source/setup-guide/app/views/LandingPageApp.js#L211-L214), to provide the concrete meaning for those event properties. With the well-documented definition, we can easily [refer to them here](https://github.com/woocommerce/pinterest-for-woocommerce/blob/429af73ca403453f802fed7d34e43d44b784dac4/TRACKING.md) without having to deduce their meaning from code or semantically inexpressible naming in the future.

puntope: State issues fixed in https://github.com/woocommerce/pinterest-for-woocommerce/pull/295/commits/3babdc6464b1d91839a42c8aa4e46c5f160bf393

I removed completely the implemented `updateState`

Reasons: 

- Reducing complexity
- @eason9487 Feedback
- Checking a similar implementation in `Analytics -> Settings -> Save` I saw that it just tracks all the settings, no matter if the user changed or not. 
- Allows to format the data as `Analytics -> Settings -> Save` with separated properties + the context
- Only allow pretended data to be tracked, excluding other sensitive data like `account_info` or trying `track_tag`

Now the Event is tracked like this:

Screenshot 2021-12-16 at 13 06 46

💅 

Inside `assets/source/setup-guide/app/helpers/prepare-for-tracking.js` I set `TRACKABLE_SETTINGS` constant for allowing properties in settings to be tracked. 

This is because I consider more critical to prevent sensitive data to be tracked than the tracking itself. In other words,I think is better to don't track a new property we forgot to add inside TRACKABLE_SETTINGS rather than forgot to block sensitive data on each setting ting addition. 

💅 

Notice a better and cleaner solution for this would be to add the settings properties (trackable) inside `state.settings` and  Connection properties (not trackable) outside `state.settings` for example `state.connection`.  But that would require more refactoring and it's out scope of this PR.

I refer again to #302 for going further on this

puntope:
> 1. 🚧  The tracking event property doesn't support nested data structure. The `updatedData` will be serialized to `'[object Object]'` string.

Fixed here: https://github.com/woocommerce/pinterest-for-woocommerce/pull/295/commits/3babdc6464b1d91839a42c8aa4e46c5f160bf393

puntope:
> 🚧 The ClaimWebsite.js also uses setAppSettings(), which calls updateSettings() internally, to update options.

Fixed also here https://github.com/woocommerce/pinterest-for-woocommerce/pull/295/commits/3babdc6464b1d91839a42c8aa4e46c5f160bf393

assets/source/setup-guide/app/components/SaveSettingsButton.js
puntope: > , it would be good to also use JSDoc, such as [wcadmin_pfw_get_started_faq](https://github.com/woocommerce/pinterest-for-woocommerce/blob/429af73ca403453f802fed7d34e43d44b784dac4/assets/source/setup-guide/app/views/LandingPageApp.js#L211-L214), to provide the concrete meaning for those event properties.

Adding missing props here https://github.com/woocommerce/pinterest-for-woocommerce/pull/295/commits/5d4eaa9bd9cb593e3496939ca06a551d28714f2f

assets/source/setup-guide/app/helpers/get-context-by-path.js
puntope: Updated here, https://github.com/woocommerce/pinterest-for-woocommerce/pull/295/commits/a2d7a4c93491393ebde99e0fba62c0f840831037

We don't really need to pass any view. Since `pinterest_connection` is not being used as a proper naming. In the parent component that refers again and again to 'settings' in the views.

I refer again to #302 for going further on this

Let e know your thoughts about including the view `pinterest_connection`/`pinterest_settings` or only `pinterest_settings`

assets/source/setup-guide/app/data/settings/actions.js
```diff
@@ -40,6 +40,13 @@ export function setIsUpdating( isUpdating ) {
 	};
 }
 
+/**
+ * Update the settings in the store or in DB
+ *
+ * @param {Object} data The data to be updated
+ * @param {boolean} saveToDb If true it persists the changes in DB
```
eason9487: 💅  It'd be more precise to indicate `saveToDb` is an optional parameter.

```suggestion
 * @param {boolean} [saveToDb=false] If true it persists the changes in DB
```

eason9487 (approved):

eason9487:
Thanks for the updates! 👍 
Apart from [this review comment](https://github.com/woocommerce/pinterest-for-woocommerce/pull/295#discussion_r770396046) that might continue to be discussed, the concerns I commented on has been resolved. Approved in advance.

puntope:
> Apart from [this review comment](https://github.com/woocommerce/pinterest-for-woocommerce/pull/295#discussion_r770396046) that

Since I started with this issue by accident, I rely on @tomalec to make the final decision about the merge.

assets/source/setup-guide/app/components/SaveSettingsButton.js
```diff
@@ -12,13 +13,55 @@ import {
...
+ *	erase_plugin_data: true|false,
+ *	product_sync_enabled: true|false,
+ *	rich_pins_on_posts: true|false,
+ *	rich_pins_on_products: true|false,
+ *	save_to_pinterest: true|false,
+ *	track_conversions: true|false,
+ * 	context: 'pinterest_settings'
+ * }
+ * ```
+ * @return {JSX.Element} Rendered element
```
tomalec: :nail_care: 

I'd shorten it to 
```suggestion
 * @fires wcadmin_pfw_save_changes_button_click with `{ context: 'pinterest_settings', … }`
 *
 * @return {JSX.Element} Rendered element
```

As putting every other property with actually unknown value, IMHO blurs the picture. I'd start to think- are those all props, or is it a subset of event props? is any of them set to a specific value?

So instead of 
image

we'd get
image

assets/source/setup-guide/app/views/ConnectionApp.js
```diff
@@ -54,14 +55,18 @@ const SettingsApp = () => {
...
 					/>
 
-					{ isGroup1Visible && <ClaimWebsite view="settings" /> }
-					{ isGroup2Visible && <SetupTracking view="settings" /> }
+					{ isGroup1Visible && (
+						<ClaimWebsite view={ SETTINGS_VIEW } />
+					) }
+					{ isGroup2Visible && (
+						<SetupTracking view={ SETTINGS_VIEW } />
+					) }
```
tomalec, line 69: :nail_care: / :scroll: 
I think what we can still do here in regards to `context` and https://github.com/woocommerce/pinterest-for-woocommerce/issues/302 is to forward the `view` to  `SaveSettingsButton` and use it as the in `recordEvent`. So from the perspective of `SaveSettingsButton` it will set the context to whatever is given to it. Then #302 would need to simply fix the view prop being set to a distinct value.

tomalec (commented):
I left one cosmetic comment and one that could be addressed here or added to the scope of #302 :) 
Also, I'd run `npm run doc:tracking` and commit updated `TRACKING.md` as well..
Other than that LGTM

assets/source/setup-guide/app/components/SaveSettingsButton.test.js
```diff
@@ -0,0 +1,42 @@
...
+		useCreateNotice: () => () => {},
+		useSettingsSelect: () => {},
+		useSettingsDispatch: () => () => {},
+	};
+} );
+jest.mock( '@woocommerce/tracks', () => {
+	return {
+		recordEvent: jest.fn(),
+	};
+} );
```
tomalec: :nail_care: 
If you are mocking default export explicitly, instead of auto mocking everything, WDYT of giving this mock a name.
To get 
>     expect( recordEvent ).toBeCalledWith(...expected)

instead of 
>     expect(jest.fn()).toBeCalledWith(...expected)


```suggestion
jest.mock( '@woocommerce/tracks', () => {
	return {
		recordEvent: jest.fn().mockName( 'recordEvent' ),
	};
} );
```


Otherwise, I'd just do jest.mock( '@woocommerce/tracks' );

assets/source/setup-guide/app/components/SaveSettingsButton.test.js
```diff
@@ -0,0 +1,42 @@
...
+
+	it( `${ eventName } is called on Save settings`, () => {
+		const { getByRole } = render( <SaveSettingsButton /> );
+		const saveSettingsBtn = getByRole( 'button' );
+
+		fireEvent.click( saveSettingsBtn );
+		expect( recordEvent.mock.calls[ 0 ][ 0 ] ).toBe( eventName );
+		expect( recordEvent.mock.calls[ 0 ][ 1 ].context ).toBe(
+			'pinterest_settings'
+		);
```
tomalec: :nail_care: 
Personally, I prefer custom assertions like, `.toBeCalledOnceWithExactly` as they are easier to read in code and in failure message, plus cater for things like ``Cannot read property 0 of undefined``

puntope: Fixed in https://github.com/woocommerce/pinterest-for-woocommerce/pull/295/commits/0afbde0966e4b9124d2ae8a573d788ea88213746

assets/source/setup-guide/app/components/SaveSettingsButton.test.js
puntope: Fixed in https://github.com/woocommerce/pinterest-for-woocommerce/pull/295/commits/0afbde0966e4b9124d2ae8a573d788ea88213746

assets/source/setup-guide/app/views/ConnectionApp.js
puntope: Fixed in https://github.com/woocommerce/pinterest-for-woocommerce/pull/295/commits/0afbde0966e4b9124d2ae8a573d788ea88213746

assets/source/setup-guide/app/components/SaveSettingsButton.js
```diff
@@ -12,13 +13,43 @@ import {
...
+ * @property {boolean} rich_pins_on_products Indicates if Add Rich Pins for Products option is checked
+ * @property {boolean} save_to_pinterest Indicates if Save to Pinterest option is checked
+ * @property {boolean} track_conversions Indicates if Track Conversion option is checked
+ * @property {string} context The context in which the event is recorded
+ */
+
+/**
+ * Save Settings button component
+ *
+ * @fires wcadmin_pfw_save_changes_button_click with `{ context: 'pinterest_settings', … }`
```
tomalec: It's no longer hard-coded here.
```suggestion
 * @fires wcadmin_pfw_save_changes_button_click with `{ context: view, … }`
```

tomalec (approved):
Left one minor JSDoc leftover, other than that LGTM

Issue #195, opened 2021-08-26
Missing optional fields for Feed ingestion that can improve Merchant performance and experience


arcmode:
Tested with v1.0.0-beta.1 due to problems with beta.2

### Is your feature request related to a problem? 
The problem is visibility at the root, with implications in merchant performance and experience

#### How to reproduce the problem
Upload sample products and wait for ingestion to complete. Navigate to Marketing/Pinterest in woocommerce and observe the alerts (see screenshot)

### Describe the solution you'd like 

Field: google_product_category 
Context Used to classify items. Can be used in Product Group Filters (this is a Pinterest feature that Merchants use to promote their catalogs with Ads)

Requirements:

Must match Google Product Category taxonomy. Any region is acceptable (ex. US). Google category id is acceptable as well. https://support.google.com/merchants/answer/6324436?hl=en

Example:
2271
-or-
Apparel & Accessories > Clothing > Dresses

Country | Google Taxonomy URL
Argentina | https://www.google.com/basepages/producttype/taxonomy.es-ES.txt
Australia | https://www.google.com/basepages/producttype/taxonomy.en-AU.txt
Austria | https://www.google.com/basepages/producttype/taxonomy.de-DE.txt
Belgium - French | https://www.google.com/basepages/producttype/taxonomy.fr-FR.txt
Belgium - Dutch | https://www.google.com/basepages/producttype/taxonomy.nl-NL.txt
Brazil | https://www.google.com/basepages/producttype/taxonomy.pt-BR.txt
Canada - English | https://www.google.com/basepages/producttype/taxonomy.en-US.txt
Canada - French | https://www.google.com/basepages/producttype/taxonomy.fr-FR.txt
Chile | https://www.google.com/basepages/producttype/taxonomy.es-ES.txt
China | https://www.google.com/basepages/producttype/taxonomy.en-US.txt
Colombia | https://www.google.com/basepages/producttype/taxonomy.es-ES.txt
Czechia | https://www.google.com/basepages/producttype/taxonomy.cs-CZ.txt
Denmark | https://www.google.com/basepages/producttype/taxonomy.da-DK.txt
Finland | https://www.google.com/basepages/producttype/taxonomy.en-US.txt
France | https://www.google.com/basepages/producttype/taxonomy.fr-FR.txt
Germany | https://www.google.com/basepages/producttype/taxonomy.de-DE.txt
Greece | https://www.google.com/basepages/producttype/taxonomy.en-US.txt
[...]

PR #303, opened 2021-12-16
Add product attributes

mikkamp:

Resolves part of #195 

This PR adds the product attribute functionality based on the Google Listings and Ads extension. This initial PR copies the basic framework for adding product attributes and includes the first attribute (condition).

Overview of changes:
- Add PHP view framework for displaying input elements in the backend
- Add Admin class for loading the backend resources
- Add product attributes and a manager to handle multiple attributes
- Add initial condition attribute
- Compile CSS and load it for product editing pages
- Add the attributes to the XML feed

1. Update to this PR 
2. Set the condition in the Pinterest tab when editing several products
3. Set the conditions in the Pinterest tab when editing several variations
4. Regenerate the feed and monitor the list of the issues (Note: I haven't found a good way to speed up the regeneration of a feed besides waiting 24h)
5. Check the issues list and confirm that the condition warning is no longer shown for the products you edited.
image
6. Check the XML feed and confirm we see the condition added as follows:
```
<g:condition>new</g:condition>
```

### Changelog entry

* Add - Product attribute for condition.

### DIFF
--- a/src/Admin/Input/Form.php
+++ b/src/Admin/Input/Form.php
@@ -1,0 +1,274 @@
+<?php
+/**
+ * Class Form
+ *
+ * @package Automattic\WooCommerce\Pinterest\Admin\Input
+ */
+
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\Pinterest\Admin\Input;
+
+use Automattic\WooCommerce\Pinterest\Exception\ValidateInterface;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class Form
+ */
+class Form implements FormInterface {
+
+	use ValidateInterface;
+
+	/**
+	 * Form name.
+	 *
+	 * @var string
+	 */
+	protected $name = '';
+
+	/**
+	 * Form data.
+	 *
+	 * @var mixed
+	 */
+	protected $data;
+
+	/**
+	 * Children.
+	 *
+	 * @var FormInterface[]
+	 */
+	protected $children = array();
+
+	/**
+	 * Parent form.
+	 *
+	 * @var FormInterface
+	 */
+	protected $parent;
+
+	/**
+	 * Is form submitted.
+	 *
+	 * @var bool
+	 */
+	protected $is_submitted = false;
+
+	/**
+	 * Form constructor.
+	 *
+	 * @param mixed $data Form data.
+	 */
+	public function __construct( $data = null ) {
+		$this->set_data( $data );
+	}
+
+	/**
+	 * Get form name.
+	 *
+	 * @return string
+	 */
+	public function get_name(): string {
+		return $this->name;
+	}
+
+	/**
+	 * Set form name.
+	 *
+	 * @param string $name Form name.
+	 *
+	 * @return FormInterface
+	 */
+	public function set_name( string $name ): FormInterface {
+		$this->name = $name;
+
+		return $this;
+	}
+
+	/**
+	 * Get form children.
+	 *
+	 * @return FormInterface[]
+	 */
+	public function get_children(): array {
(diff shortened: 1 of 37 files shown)


budzanowski:
Oh my :)

src/Admin/Input/Form.php
```diff
@@ -0,0 +1,274 @@
...
+	 * @return FormInterface
+	 *
+	 * @throws FormException If form is already submitted.
+	 */
+	public function remove( string $name ): FormInterface {
+		if ( $this->is_submitted ) {
+			throw FormException::cannot_modify_submitted();
+		}
+
+		if ( isset( $this->children[ $name ] ) ) {
```
budzanowski:
```suggestion
		if ( $this->has( $name ) ) {
```
We already have it so why not - just a nitpick, safe to ignore :)

src/Admin/Product/Attributes/VariationsAttributes.php
```diff
@@ -0,0 +1,170 @@
...
+	/**
+	 * Register a service.
+	 */
+	public function register(): void {
+		add_action(
+			'woocommerce_product_after_variable_attributes',
+			function ( int $variation_index, array $variation_data, WP_Post $variation ) {
+				$this->render_attributes_form( $variation_index, $variation );
+			},
+			90,
```
budzanowski, line 58: Could we maybe comment on why this priority was chosen?

src/Admin/Product/Attributes/VariationsAttributes.php
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	/**
+	 * Render the attributes form for variations.
+	 *
+	 * @param int     $variation_index Position in the loop.
+	 * @param WP_Post $variation       Post data.
+	 */
+	private function render_attributes_form( int $variation_index, WP_Post $variation ) {
+		/**
```
budzanowski, line 78: Do we really need to comment on this?

src/ProductsXmlFeed.php
```diff
@@ -79,11 +81,32 @@ public static function get_xml_item( $product ) {
 			}
 		}
 
+		$xml .= self::get_attributes_xml( $product, "\t\t\t" );
```
budzanowski, line 84: Could we call it like any other field through the loop over `self::$feed_item_structure` ? Not breaking this pattern would be the best option imo.

budzanowski:
Looks great. The only real thing I have is the https://github.com/woocommerce/pinterest-for-woocommerce/pull/303#pullrequestreview-845404578 comment. I think we should stick with the pattern. All other comments are optional.

src/ProductsXmlFeed.php
mikkamp: That makes sense. My original thought was to just let the attribute manager class to handle all attributes (in this PR we only add one but there are more to be added).

But if we prefer to stick with the structure then we can add a function here for each attribute we add.

src/Admin/Product/Attributes/VariationsAttributes.php
mikkamp: It's to clarify the difference between $variation which is an instance of WP_Post and $product which is an instance of WC_Product_Variation. It's not strictly necessary but I copied this code over directly from GLA so I left it as is.

src/ProductsXmlFeed.php
mikkamp: Based off a discussion in Slack we decided to stick with one function for all attributes. The reason being that attributes are optionally set for a product, and not all attributes are available for each product type. Added a comment to clarify this behaviour.

budzanowski (approved):

PR #305, opened 2021-12-17
Feature/250 api v4

### LINKED ISSUE #249: Multiple catalog created on pinterest with no possibility to delete them
Placeholder for forum issue to come back to investigate https://wordpress.org/support/topic/multiple-catalog-created-on-pinterest-with-no-possibility-to-delete-them/

Will update with more information when I have a moment

### LINKED ISSUE #250: v4 API Endpoint Migration
v3 API endpoints will be deprecated in January 2022 and a number of new APIs are becoming available. 

The v3 API endpoints will still be available until the end of Q1 (end of March) so we won’t need to notify merchants to update until a later release (which we should be able to do with a remote inbox notification or two).

The new endpoints generalize internal business logic, helping to reduce the number of integration steps and lowering the risk that Pinterest API updates break the integration.

**Goal/s:**
* Upgrade ads/v3 to ads/v4 endpoints.
* Use new catalog APIs for merchant registration, domain verification, and catalog ingestion.

## Merchant

* [ ] Create/Add/Update Merchant (New single endpoint to handle the various use cases in setting up a merchant account. Simply provide the required fields and the Merchant account will either be created or updated. The feed profile will also either be created or updated based on the locale, country, and currency you specify. See param descriptions for more information.
* [ ] Disconnect merchant feed 🆕  (This handles the disconnect of the merchant from your platform - can be added to disconnect processes)
* [ ] Get Merchant Feed profiles 🆕  (mostly relevant for work on multi-feed work support and https://github.com/woocommerce/pinterest-for-woocommerce/issues/223)

## Advertiser

* [ ] Create Advertiser
* [ ] Get list of advertisers associated with user
* [ ] Connect Advertiser 🆕 (An additional required step to associate the advertiser and tag with your platform on Pinterest.
)
* [ ] Disconnect Advertiser 🆕 (A required step when changing advertisers or removing an account to disconnect advertiser and tag from your platform on Pinterest.)

## Tags

* [ ] Create Tag ([new endpoint](https://developers.pinterest.com/docs/redoc/web_platform_ads_v4/#tag/conversion_tags-(adsv4)) with option to enable enhanced match flag/s)
* [ ] Get all tags

# Technical

## Base API Version Parameter

https://github.com/woocommerce/pinterest-for-woocommerce/blob/9d5cd583a5decc4a8342d1e48f14a0a581332387/src/API/Base.php#L25-L26

## Merchant

**Create/Add/Update Merchant**

https://github.com/woocommerce/pinterest-for-woocommerce/blob/296316e9be7aabe1afdfd5110d19c072c1003495/src/API/Base.php#L401-L459

**Disconnect**

https://github.com/woocommerce/pinterest-for-woocommerce/blob/296316e9be7aabe1afdfd5110d19c072c1003495/src/API/AuthDisconnect.php#L44

---

## Advertiser

**Create Advertiser**

https://github.com/woocommerce/pinterest-for-woocommerce/blob/296316e9be7aabe1afdfd5110d19c072c1003495/src/API/Base.php#L300-L320
 
**Get Advertisers**

https://github.com/woocommerce/pinterest-for-woocommerce/blob/296316e9be7aabe1afdfd5110d19c072c1003495/src/API/Base.php#L330-L333

https://github.com/woocommerce/pinterest-for-woocommerce/blob/9df1bf7c1eefc08a451d71ff5b18e64609d887b9/src/API/Advertisers.php#L36-L76

---

## Tags

**Create Tag**

https://github.com/woocommerce/pinterest-for-woocommerce/blob/296316e9be7aabe1afdfd5110d19c072c1003495/src/API/Base.php#L348-L368

**Get all tags**

https://github.com/woocommerce/pinterest-for-woocommerce/blob/296316e9be7aabe1afdfd5110d19c072c1003495/src/API/Base.php#L336-L345

alexiglesias31:

Closes #249
Closes #250

#

- Upgrade ads/v3 to ads/v4 endpoints.
- Use new catalog APIs for merchant registration, domain verification, and catalog ingestion.

Edit: As a side effect this PR by implementing the `disconnect` method, flushes all feeds created by the plugin, something that up to this point wasn't possible, causing merchants to report multiple feeds created with no way to delete. 

1. Install the plugin.
2. Make the connection.

### Changelog entry

>


findingsimple:
Requested a review from Ventures

class-pinterest-for-woocommerce.php
```diff
@@ -566,25 +570,87 @@ public static function save_token( $token ) {
...
+				if ( 'success' !== $result['status'] ) {
+					throw new \Exception( esc_html__( 'Response error on disconnect merchant.', 'pinterest-for-woocommerce' ), 400 );
+				}
+
+				// Disconnect the advertiser from Pinterest.
+				$connected_advertiser = self::get_setting( 'tracking_advertiser', null );
+				$connected_tag        = self::get_setting( 'tracking_tag', null );
+
+				if ( $connected_advertiser && $connected_tag ) {
+					$result = Pinterest\API\Base::disconnect_advertiser( $connected_advertiser, $connected_tag );
```
JPry: Is there a reason we're not calling `Pinterest\API\AdvertiserConnect::disconnect_advertiser()` like we're doing below in the `maybe_disconnect_advertiser()` method?

src/API/AdvertiserConnect.php
```diff
@@ -0,0 +1,132 @@
...
+ * Handle Pinterest Advertiser Connect
+ *
+ * @package     Pinterest_For_Woocommerce/API
+ * @version     1.0.0
+ */
+
+namespace Automattic\WooCommerce\Pinterest\API;
+
+use \WP_REST_Server;
+use \WP_REST_Request;
```
JPry, line 12: Two things: 

* We should import the `WP_Error` class, because it is referenced in a code comment and as a possible return value (line 74)
* When importing, the leading `\` character is unnecessary and should be skipped.

```suggestion
use WP_Error;
use WP_REST_Server;
use WP_REST_Request;
```

src/API/AdvertiserConnect.php
```diff
@@ -0,0 +1,132 @@
...
+	 *
+	 * @return array|WP_Error
+	 *
+	 * @throws \Exception PHP Exception.
+	 */
+	public function connect_advertiser( WP_REST_Request $request ) {
+
+		try {
+
+			$advertiser_id = $request->has_param( 'advrtsr_id' ) ? $request->get_param( 'advrtsr_id' ) : false;
```
JPry, line 50: Is `advrtsr_id` a typo, or is this the correct parameter name?

src/API/Base.php
```diff
@@ -88,13 +90,22 @@ public static function make_request( $endpoint, $method = 'POST', $payload = arr
 		}
 
 		try {
-			$api     = empty( $api ) ? '' : trailingslashit( $api );
+			$api         = empty( $api ) ? '' : trailingslashit( $api );
+			$api_version = 'ads/' === $api ? self::API_ADS_VERSION : self::API_VERSION;
+
 			$request = array(
-				'url'    => self::API_DOMAIN . '/' . $api . 'v' . self::API_VERSION . '/' . $endpoint,
+				'url'    => self::API_DOMAIN . '/' . $api . 'v' . $api_version . '/' . $endpoint,
```
JPry: Minor: Since we're adjusting this line anyway, this would be more readable with concatenation:

```suggestion
				'url'    => self::API_DOMAIN . "/{$api}v{$api_version}/{$endpoint}",
```

src/API/Base.php
```diff
@@ -88,13 +90,22 @@ public static function make_request( $endpoint, $method = 'POST', $payload = arr
...
 				'method' => $method,
 				'args'   => $payload,
 			);
 
+			if ( 'ads/' === $api && 'POST' === $method ) {
```
JPry, line 103: Just a note that these slight differences in the Ads API make me think that there should be a separate child class for handling Ads API calls. But we can open a new issue to address that in the future.

src/API/Base.php
```diff
@@ -320,6 +340,46 @@ public static function create_advertiser( $tos_id ) {
...
+	 * Connect the advertiser with the platform.
+	 *
+	 * @param string $advertiser_id The advertiser ID.
+	 * @param string $tag_id        The tag ID.
+	 *
+	 * @return mixed
+	 */
+	public static function connect_advertiser( $advertiser_id, $tag_id ) {
+		return self::make_request(
+			'advertisers/' . $advertiser_id . '/connect/',
```
JPry:
```suggestion
			"advertisers/{$advertiser_id}/connect/",
```

src/API/Base.php
```diff
@@ -320,6 +340,46 @@ public static function create_advertiser( $tos_id ) {
...
+	 * Disconnect advertiser from the platform.
+	 *
+	 * @param string $advertiser_id The advertiser ID.
+	 * @param string $tag_id        The tag ID.
+	 *
+	 * @return mixed
+	 */
+	public static function disconnect_advertiser( $advertiser_id, $tag_id ) {
+		return self::make_request(
+			'advertisers/' . $advertiser_id . '/disconnect/',
```
JPry:
```suggestion
			"advertisers/{$advertiser_id}/disconnect/",
```

src/API/Base.php
```diff
@@ -341,7 +401,7 @@ public static function get_advertisers( $pinterest_user = null ) {
 	 * @return mixed
 	 */
 	public static function get_advertiser_tags( $advertiser_id ) {
-		return self::make_request( 'advertisers/' . $advertiser_id . '/tags/', 'GET', array(), 'ads' );
+		return self::make_request( 'advertisers/' . $advertiser_id . '/conversion_tags/', 'GET', array(), 'ads' );
```
JPry:
```suggestion
		return self::make_request( "advertisers/{$advertiser_id}/conversion_tags/", 'GET', array(), 'ads' );
```

src/API/Base.php
```diff
@@ -357,11 +417,10 @@ public static function create_tag( $advertiser_id ) {
 		$tag_name = apply_filters( 'pinterest_for_woocommerce_default_tag_name', esc_html__( 'Auto-created by Pinterest for WooCommerce', 'pinterest-for-woocommerce' ) );
 
 		return self::make_request(
-			'tags/',
+			'advertisers/' . $advertiser_id . '/conversion_tags',
```
JPry:
```suggestion
			"advertisers/{$advertiser_id}/conversion_tags",
```

src/API/Base.php
```diff
@@ -399,55 +458,35 @@ public static function get_merchant( $merchant_id ) {
...
-	 * Adds the merchant's feed using the given arguments.
+	 * Creates a merchant for the authenticated user or updates the existing one. On success the Merchant ID is returned instead of the full merchant object.
 	 *
-	 * @param string $merchant_id The merchant ID the feed belongs to.
-	 * @param array  $args        The arguments to be passed to the API request.
+	 * @param array $args Payload to be sent to the request.
 	 *
 	 * @return mixed
 	 */
-	public static function add_merchant_feed( $merchant_id, $args ) {
```
JPry, line 428: Just a note that we should be cautious about removing functions declared as `public`. While it's currently unlikely that anyone else is writing code dependent on ours, this is how compatibility breaks happen.

If we need to remove some functions/methods, we should consider having some way to mark these as deprecated. Even better is to ensure the only methods marked `public` absolutely must be `public`.

I think I'll open an issue so what we can more comprehensively determine how to handle this.

src/API/Base.php
```diff
@@ -460,9 +499,9 @@ public static function update_merchant_feed( $merchant_id, $feed_id, $args ) {
 	 *
 	 * @return mixed
 	 */
-	public static function get_merchant_feed( $merchant_id, $feed_id ) {
+	public static function get_merchant_feed_report( $merchant_id, $feed_id ) {
 		return self::make_request(
-			"catalogs/datasource/feed_report/{$merchant_id}/",
+			'catalogs/datasource/feed_report/' . $merchant_id . '/',
```
JPry: This should revert back to interpolation.

```suggestion
			"catalogs/datasource/feed_report/{$merchant_id}/",
```

src/Feeds.php
```diff
@@ -0,0 +1,140 @@
...
+		}
+	}
+
+
+	/**
+	 * Verify if the local feed is already registered to the merchant
+	 *
+	 * @param string $merchant_id The merchant ID.
+	 *
+	 * @return string|false Returns the ID of the feed if properly registered or false otherwise.
```
JPry: We should only have one return type. The best option would be to throw an exception that the calling code can handle, or another workable option would be to return an empty string.

src/Feeds.php
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * Pinterest for WooCommerce Feeds related helper methods
+ *
+ * @package     Pinterest_For_WooCommerce/Classes/
+ * @version     1.0.0
+ */
+
+namespace Automattic\WooCommerce\Pinterest;
+
+if ( ! defined( 'ABSPATH' ) ) {
```
JPry, line 11: Let's import our classes here, then we can use them without any prefixes in the code within this file.

```suggestion

use Automattic\WooCommerce\Pinterest\API\Base;
use Exception;

if ( ! defined( 'ABSPATH' ) ) {
```

src/Feeds.php
```diff
@@ -0,0 +1,140 @@
...
+ */
+class Feeds {
+
+	/**
+	 * Get a specific merchant feed using the given arguments.
+	 *
+	 * @param string $merchant_id The merchant ID the feed belongs to.
+	 * @param string $feed_id     The ID of the feed.
+	 *
+	 * @return mixed
```
JPry: It looks like we only have one possible return value that isn't an `Exception`. Should this be `object` instead of `mixed`?

src/Feeds.php
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+
+	/**
+	 * Get merchant's feed based on feed location
+	 *
+	 * @param string $merchant_id   The merchant ID.
+	 * @param string $feed_location The feed full location.
+	 *
+	 * @return mixed
```
JPry: Same as above, I think this should be `object` instead of `mixed`.

src/Merchants.php
```diff
@@ -51,26 +51,48 @@ public static function get_merchant() {
...
+		return $merchant;
+	}
+
+
+	/**
+	 * Get merchant object using the merchant ID.
+	 *
+	 * @param string $merchant_id The merchant ID.
+	 *
+	 * @return mixed|boolean
```
JPry: We should only have one return value from this method, and we can allow the thrown exception to bubble up to the calling code. That would allow the method body to simply be `return Base::get_merchant( $merchant_id );`.

Or even better, eliminate this method and just call `Base::get_merchant( $merchant_id )` directly with exception handling 🙂

src/Tracking.php
```diff
@@ -104,6 +104,16 @@ public static function maybe_init() {
 
 		// Print to head.
 		add_action( 'wp_head', array( __CLASS__, 'print_script' ) );
+
+		// Verify if advertiser and tag need to be connected due to a plugin upgrade.
+		try {
+
```
JPry:
```suggestion
```

src/Tracking.php
```diff
@@ -483,4 +493,22 @@ public static function print_script() {
...
+	 */
+	private static function maybe_connect_advertiser_tag() {
+
+		$is_connected         = Pinterest_For_Woocommerce()::get_data( 'is_advertiser_connected' );
+		$connected_advertiser = Pinterest_For_Woocommerce()::get_setting( 'tracking_advertiser' );
+		$connected_tag        = Pinterest_For_Woocommerce()::get_setting( 'tracking_tag' );
+
+		// Check if advertiser is already connected.
+		if ( ! $is_connected && $connected_advertiser && $connected_tag ) {
+			API\AdvertiserConnect::connect_advertiser_and_tag( $connected_advertiser, $connected_tag );
```
JPry: We should use an import so we can reference `AdvertiserConnect` directly.

JPry (changes requested):

assets/source/setup-guide/app/steps/SetupTracking.js
```diff
@@ -218,9 +225,66 @@ const SetupTracking = ( { view } ) => {
...
+		[
+			appSettings,
+			setIsSaving,
+			setAppSettings,
+			connectAdvertiser,
+			createNotice,
+		]
 	);
 
+	const connectAdvertiser = useCallback( async () => {
```
ecgan: I suggest making this `connectAdvertiser` accept the advertiser ID and tracking tag as parameters, so it doesn't need to depend on `appSettings`, and to make it clearer that the action of connecting advertiser requires the two parameters.

assets/source/setup-guide/app/steps/SetupTracking.js
```diff
@@ -218,9 +225,66 @@ const SetupTracking = ( { view } ) => {
...
+
+		try {
+			const results = await apiFetch( {
+				path:
+					wcSettings.pinterest_for_woocommerce.apiRoute +
+					'/tagowner/?advrtsr_id=' +
+					appSettings.tracking_advertiser +
+					'&tag_id=' +
+					appSettings.tracking_tag,
+				method: 'GET',
```
ecgan: To "connect advertiser" is to _create a connection_ to the advertiser, so it should be a `POST` method.

assets/source/setup-guide/app/steps/SetupTracking.js
```diff
@@ -206,6 +207,12 @@ const SetupTracking = ( { view } ) => {
 				await setAppSettings( {
 					[ name ]: value ?? ! appSettings[ name ],
 				} );
+				if (
+					appSettings?.tracking_advertiser &&
+					appSettings?.tracking_tag
+				) {
+					await connectAdvertiser();
```
ecgan: I'm not sure if I'm understanding things correctly here. We are calling `connectAdvertiser` in `saveOptions`, and `saveOptions` is called in `handleOptionChange`, so we are connecting advertiser in every option change? Is this expected?

ecgan (commented):
Left a few comments in the front-end JS work.

ecgan:
This is related to my comment in https://github.com/woocommerce/pinterest-for-woocommerce/pull/305#discussion_r774117988. When I change the advertiser account in the select control, a notice "Advertiser connected successfully." would occasionally show up. Is this expected?

📹 See my demo video with my voice annotation:

https://user-images.githubusercontent.com/417342/147146254-98796bcb-6f63-4c59-aa3d-b1911cd78b4a.mp4

assets/source/setup-guide/app/steps/SetupTracking.js
```diff
@@ -218,9 +225,66 @@ const SetupTracking = ( { view } ) => {
...
+		setIsFetching( true );
+
+		try {
+			const results = await apiFetch( {
+				path:
+					wcSettings.pinterest_for_woocommerce.apiRoute +
+					'/tagowner/?advrtsr_id=' +
+					appSettings.tracking_advertiser +
+					'&tag_id=' +
+					appSettings.tracking_tag,
```
ecgan: I'd prefer to do this using template literals instead of string concatenation spanning multiple lines.

Also, with regards to https://github.com/woocommerce/pinterest-for-woocommerce/pull/305#discussion_r774116143, if the method is changed to `POST`, it would be good to put the advertiser and tag in POST body instead of query string.

ecgan (commented):
Added one more comment in JS.

src/API/AdvertiserConnect.php
alexiglesias31: It's following the same name convention than in previous version, imho no need to change this.

assets/source/setup-guide/app/steps/SetupTracking.js
alexiglesias31: `handleOptionChange` is called each time that selected advertiser or tag change, since ads/v4 API requires to connect the advertiser with the tag, we need to verify that current combination of advertiser and tag is properly connected to the platform and reconnect it each time that selected values changes. The reconnect parameter in the response indicates that the combination changed or not. The remote request to the platform only is called if that combination is different than the previously connected one.

alexiglesias31: Imho we need to verify if the current combination is connected or not every time that the selected options change

ecgan: @alexiglesias31 , thank you for the info. 🙂 

Just want to be sure, have you checked out the demo video I posted in https://github.com/woocommerce/pinterest-for-woocommerce/pull/305#issuecomment-999819703? I have only two advertiser accounts, I keep switching back and forth between them, and the notice "Advertiser connected successfully." would occasionally randomly show up for either one of them. Is this expected?

> we need to verify that current combination of advertiser and tag is properly connected to the platform and reconnect it each time that selected values changes.

I was expecting that verification and connection would only happen when users click on the "Complete Setup" button. Is that something that is not feasible, or perhaps invalid or out of scope? 🤔

alexiglesias31: @ecgan we added a couple of modifications in order than connect advertiser only runs when users click the "Complete Setup" or "Save Settings" buttons.

ecgan: @alexiglesias31 , thank you. I'll look into the modifications probably tomorrow. Meanwhile, a heads-up for you: there is now a conflict in the PR. 😅

alexiglesias31: Perfect @ecgan. Merge conflict already solved 😉.

assets/source/setup-guide/app/helpers/effects.js
```diff
@@ -4,6 +4,9 @@
 import { useSelect, useDispatch } from '@wordpress/data';
 import { useEffect, useCallback } from '@wordpress/element';
 
+import { __ } from '@wordpress/i18n';
```
ecgan: Remove empty line in between the imports.

```suggestion
import { useEffect, useCallback } from '@wordpress/element';
import { __ } from '@wordpress/i18n';
```

assets/source/setup-guide/app/helpers/effects.js
```diff
@@ -54,3 +57,52 @@ export const useBodyClasses = ( style ) => {
 		};
 	}, [ style ] );
 };
+
+export const useConnectAdvertiser = () => {
```
ecgan: ❤️ Good job in creating a new hook to share the connect advertiser logic! 🙂 👍

assets/source/setup-guide/app/steps/SetupTracking.js
```diff
@@ -283,10 +286,16 @@ const SetupTracking = ( { view = 'settings' } ) => {
...
+			appSettings.tracking_tag
+		);
+
 		recordEvent( 'pfw_setup', {
 			target: 'complete',
 			trigger: 'setup-tracking-complete',
 		} );
+
 		// Force reload WC admin page to initiate the relevant dependencies of the Dashboard page.
 		const path = getNewPath( {}, '/pinterest/settings', {} );
```
ecgan: 🚧 In `connectAdvertiser`, we have a try-catch block around the API call. If there is an error in the API call, we will catch the error, display an error notice, _and then_ return to this `handleCompleteSetup` function and call `recordEvent` and proceed to the new page using `window.location`, which is not desired.

Another thing to note is that `connectAdvertiser` is called from two places: `SetupTracking.js` without try-catch block, and `SaveSettingsButton.js` with try-catch block. I think it would be good to make both places similar.

I can think of two ideas to solve this: 

Idea 1: In `useConnectAdvertiser`, we catch the error and re-throw the error again, so that the parent component is able to know there is an error in calling `await connectAdvertiser`.

Idea 2: We don't need `useConnectAdvertiser` hook that wraps around `useCreateNotice`. Instead, we just need a regular `connectAdvertiser` function that wraps around the `await apiFetch` API call, and let the parent component call the function in the try-catch block and display the error notice. `SetupTracking.js` and `SaveSettingsButton.js` already have `createNotice`.

Personally, I'm leaning towards idea 2. Feel free to implement other ideas you see fit. 🙂

ecgan (changes requested):
Reviewed the JS code. They have been simplified compared to my previous round of review, which is great. 👍 

Found an important blocker 🚧 related to missing error handling in `SetupTracking.js`.

assets/source/setup-guide/app/helpers/effects.js
```diff
@@ -54,3 +55,26 @@ export const useBodyClasses = ( style ) => {
...
+			if ( trackingAdvertiser === results.connected ) {
+				return true;
+			} else {
+				return false;
+			}
+		} catch ( error ) {
+			throw error;
+		}
+	}, [] );
+};
```
ecgan: `useCallback` is not needed. We can just do it like the following. Note that it is not a hook nor an effect, so consider putting this in other place outside of this `effects.js` file.

```js
const connectAdvertiser = async ( trackingAdvertiser, trackingTag ) => {
	try {
		const results = await apiFetch( {
			path: `${ wcSettings.pinterest_for_woocommerce.apiRoute }/tagowner/`,
			data: {
				advrtsr_id: trackingAdvertiser,
				tag_id: trackingTag,
			},
			method: 'POST',
		} );

		if ( trackingAdvertiser === results.connected ) {
			return true;
		}

		return false;
	} catch ( error ) {
		throw error;
	}
};
```

assets/source/setup-guide/app/components/SaveSettingsButton.js
```diff
@@ -53,6 +56,26 @@ const SaveSettingsButton = ( { view } ) => {
...
+			if (
+				appSettings?.tracking_advertiser &&
+				appSettings?.tracking_tag
+			) {
+				const result = await connectAdvertiser(
+					appSettings.tracking_advertiser,
+					appSettings.tracking_tag
+				);
+
+				if ( result.reconnected ) {
```
ecgan: 🚧 `connectAdvertiser` now returns boolean true/false, there is no `reconnected` property anymore, so this `if` statement will always be evaluated to `false`.

I assume this should now be the following instead?

```suggestion
				if ( result ) {
```

assets/source/setup-guide/app/components/SaveSettingsButton.js
```diff
@@ -53,6 +56,26 @@ const SaveSettingsButton = ( { view } ) => {
...
+					createNotice(
+						'success',
+						__(
+							'The advertiser was connected successfully.',
+							'pinterest-for-woocommerce'
+						)
+					);
```
ecgan, line 74: When `connectAdvertiser` is successful, _two notices_ will be displayed: `'The advertiser was connected successfully.'` and `'Your settings have been saved successfully.'` Just want to be sure, is this expected? Or should there be only one notice?

assets/source/setup-guide/app/helpers/effects.js
```diff
@@ -54,3 +55,26 @@ export const useBodyClasses = ( style ) => {
...
+					tag_id: trackingTag,
+				},
+				method: 'POST',
+			} );
+
+			if ( trackingAdvertiser === results.connected ) {
+				return true;
+			} else {
+				return false;
+			}
```
ecgan: There is an ESLint error here. You can use ESLint plugins for your editor to catch this.

image

assets/source/setup-guide/app/steps/SetupTracking.js
```diff
@@ -283,14 +286,40 @@ const SetupTracking = ( { view = 'settings' } ) => {
...
-		} );
-		// Force reload WC admin page to initiate the relevant dependencies of the Dashboard page.
-		const path = getNewPath( {}, '/pinterest/settings', {} );
-
-		window.location = new URL( wcSettings.adminUrl + path );
+		try {
+			const result = await connectAdvertiser(
+				appSettings.tracking_advertiser,
+				appSettings.tracking_tag
+			);
```
ecgan, line 292: `connectAdvertiser` returns boolean true/false. If it returns false (which seems to mean account is not connected), it should not proceed with `recordEvent` and redirect with `window.location`, right? 🤔

ecgan (changes requested):
Reviewed the JS code, left some questions below, and requesting for some JS code changes.

assets/source/setup-guide/app/components/SaveSettingsButton.js
alexiglesias31: Yeap, it is expected, `'The advertiser was connected successfully.'` will be displayed only when the advertiser is reconnected (when the advertiser or the tag are changed).

assets/source/setup-guide/app/steps/SetupTracking.js
alexiglesias31: The boolean true/false indicates if the account was reconnected (reconnected means that a new advertiser or tag is connected to the platform). If the advertiser and tag were already connected (no changes in the previous values), `connectAdvertiser` will return false.
If there is a problem connecting the advertiser the method will throw an error that will be handled by the catch block.
So, if `connectAdvertiser` doesn't throw an error it should proceed with `recordEvent` and redirect process.

alexiglesias31:
> Reviewed the JS code, left some questions below, and requesting for some JS code changes.

Fixed the suggestions 👍

assets/source/setup-guide/app/helpers/connect-advertiser.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * External dependencies
+ */
+import apiFetch from '@wordpress/api-fetch';
+
+const connectAdvertiser = async ( trackingAdvertiser, trackingTag ) => {
+	try {
+		const results = await apiFetch( {
+			path: `${ wcSettings.pinterest_for_woocommerce.apiRoute }/tagowner/`,
```
ecgan: Remove the trailing slash, so that we can see the API request better in browser dev tools' network panel.

```suggestion
			path: `${ wcSettings.pinterest_for_woocommerce.apiRoute }/tagowner`,
```

For reference, with the trailing slash, this is what we see in network panel:

image

assets/source/setup-guide/app/helpers/connect-advertiser.js
```diff
@@ -0,0 +1,27 @@
...
+				tag_id: trackingTag,
+			},
+			method: 'POST',
+		} );
+
+		if ( trackingAdvertiser === results.connected ) {
+			return true;
+		}
+
+		return false;
```
ecgan: Based on your comment in https://github.com/woocommerce/pinterest-for-woocommerce/pull/305#discussion_r780374933:

> The boolean true/false indicates if the account was reconnected (reconnected means that a new advertiser or tag is connected to the platform). If the advertiser and tag were already connected (no changes in the previous values), `connectAdvertiser` will return false.

I think this line should be like the following instead?

```suggestion
		return results.reconnected;
```

`results.connected` is the account ID, and `results.reconnected` is boolean true/false after changing advertiser and tracking tag.

image

📹 Here's a voice-annotated demo video (based on the current code [`3c82593` (#305)](https://github.com/woocommerce/pinterest-for-woocommerce/pull/305/commits/3c825936a6dbc7c66e0adacd0488c9344c766c8d), not the suggested change) showing that the notice `'The advertiser was connected successfully.'` will always show even if we did not change the advertiser and tracking tag.

https://user-images.githubusercontent.com/417342/148743590-99a0d7a6-441a-4dc4-8bd7-294349d68ab9.mov

ecgan (changes requested):
Tested the code, found that the account connection notices do not behave as per mentioned in the comments, left two comments with suggestions and demo video below. I think the JS code is pretty much there, so once the issues are fixed, the JS code should be good to go.

Tagging @JPry for his review on the PHP code.

assets/source/setup-guide/app/helpers/connect-advertiser.js
alexiglesias31: Good caught @ecgan. Correct, we need to return the reconnected parameter in the request 👍

ecgan (approved):
Tested the code. The account connection notice issue that I found in my previous review (see https://github.com/woocommerce/pinterest-for-woocommerce/pull/305#pullrequestreview-847539783) has been fixed. To be more specific: when we change the plugin settings without changing the account advertiser and tracking tag settings, we will see only one notice that says "Your settings have been saved successfully."; when we do change the account advertiser and/or tracking tag settings and they are reconnected, we will see two notices that say "Your settings have been saved successfully." and "The advertiser was connected successfully.".

Reviewed the JS code. LGTM 👍

JPry (approved):
LGTM 👍🏻

Issue #307, opened 2021-12-21
Steps 2,3 are not clickable in onboarding wizard 


tomalec:
### Describe the bug:

Steps 2,3 in the stepper header are not clickable, only 1 is.
![non-clickable-steps](https://user-images.githubusercontent.com/17435/146969127-0059e195-7bd2-48ff-98a0-36c18b9575f0.gif)

### Steps to reproduce:

1. Go to the onboarding wizard `/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Fonboarding`
2. Complete step 1, click "Continue"
3. Go back to step 1,
4. Try clicking step 2 in the header -> this should work
5. Go to step 2 by clicking "Continue"
6. Complete step 2, click "Continue"
7. Try going back to step 2 by clicking in the stepper header -> this should work
8. Go back, by browser navigation, or by modifying the URL
9. Try going to step 3 -> this should work

#### Expected behavior

Any step following a complete one should be clickable.

#### Actual behavior

Only the first one is clickable.

### Additional details:

1. It seems the reason for it is an incomplete implementation at https://github.com/woocommerce/pinterest-for-woocommerce/blob/b50e790ee19f0e2795d18f6e45e14531db6d7448/assets/source/setup-guide/app/views/WizardApp.js#L99. The condition check if a given step is following the completed one, but `isComplete` flag [is not set anywhere](https://github.com/woocommerce/pinterest-for-woocommerce/search?q=isComplete).

PR #308, opened 2021-12-21
Add `wcadmin_pfw_setup` tracking events.

tomalec:

- Implement `wcadmin_pfw_setup` tracking event. Part of #232.
- Add some test coverage for `LandingPageApp`.
- Add a custom matcher `toContainURLSearchParam` for easier WP-admin location assertions.

### Screenshots:
Setup events

#### Happy path
0. Disconnect your account, to restart the onboarding.
	1. Go to `/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Fconnection`
	2. Click "Disconnect", confirm
1. Enable track logging with `localStorage.setItem('debug', 'wc-admin*')` in dev tools
2. Go to landing page `/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Flanding`
3. Click on "Get started"
	-  `wcadmin_pfw_setup { target: 'onboarding', trigger: 'get-started' }` should be fired
4. Connect the account (this will eventually take you to step 2)
5. Click "Set up your business account", to get back to step 1
	- `wcadmin_pfw_setup { target: 'setup-account', trigger: 'wizard-stepper' }` should be fired
	- other steps are not clickable due to https://github.com/woocommerce/pinterest-for-woocommerce/issues/307, but should fire the event once enabled.
6. Click "Continue"
	- `wcadmin_pfw_setup { target: 'claim-website' , trigger: 'setup-account-continue' }` should be fired
7. Click "Start verification"
8. Click "Continue"
	- `wcadmin_pfw_setup { target: 'setup-tracking', trigger: 'claim-website-continue' }` should be fired
8. Click "Complete setup"
	- `wcadmin_pfw_setup { target: 'complete', trigger: 'setup-tracking-complete' }` should be fired
9. Manually go back to `/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Flanding`
10. Click on "Get started"
	-  No `wcadmin_pfw_setup` should be fired
	
#### Try again
I'm not sure what's the recommended way to achieve those states, but you can get there by mocking

##### Tags
fetch tags

0. Change the path in https://github.com/woocommerce/pinterest-for-woocommerce/blob/2a6532352c8c3e3a44c8913e09bc753ede306274/assets/source/setup-guide/app/steps/SetupTracking.js#L152 to `'/fooooo/' +`, to mimic a problem in fetching tags.
1. Go to `/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Fonboarding&step=setup-tracking`
2. Click "Try Again"
	- `wcadmin_pfw_setup { target: 'fetch-tags', trigger: 'setup-tracking-try-again' }` should be fired

##### Advertisers
fetch advertisers

0. Do the 0. from the section above. https://github.com/woocommerce/pinterest-for-woocommerce/blob/2a6532352c8c3e3a44c8913e09bc753ede306274/assets/source/setup-guide/app/steps/SetupTracking.js#L152 to `'/fooooo/' +`
1. Add `appSettings.tracking_advertiser = false;` to https://github.com/woocommerce/pinterest-for-woocommerce/blob/add/232-setup-event/assets/source/setup-guide/app/steps/SetupTracking.js#L63:L63
1. Go to `/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Fonboarding&step=setup-tracking`
2. Click "Try Again"
	- `wcadmin_pfw_setup { target: 'fetch-advertisers', trigger: 'setup-tracking-try-again' }` should be fired

### Additional notes
0. I think the custom matcher added here, could be eventually offloaded to an external package. Like `@woocommerce/test-utils`?
1. I didn't add test for events in Setup Tracking and Wizard, as those components does not have any tests, and are quite complicated. Currently, rendering them throws a number of exceptions. I think we can offload this to a separate issue, as at least to me it would require a significant effort, and probably some cleanup in the those bigger components. For example, to avoid errors like:
	```
	      ReferenceError: Cannot access 'fetchTags' before initialization
	  
	        106 | 		setIsFetching( false );
	        107 | 	}, [
	      > 108 | 		fetchTags,
	            | 		^
	  ```

### Changelog entry

### DIFF
--- a/assets/source/setup-guide/app/views/LandingPageApp.js
+++ b/assets/source/setup-guide/app/views/LandingPageApp.js
@@ -26,14 +26,41 @@ import UnsupportedCountryNotice from '../components/UnsupportedCountryNotice';
 const tosHref = 'https://business.pinterest.com/business-terms-of-service/';
 
 /**
+ * Triggered on events during setup,
+ * like starting, ending, or navigating between steps.
+ *
+ * @event wcadmin_pfw_setup
+ *
+ * @property {string} target Setup phase that the user navigates to.
+ * @property {string} trigger UI element that triggered the action, e.g. `wizard-stepper` or `get-started` button.
+ */
+
+/**
  * Welcome Section Card.
  * To be used in getting started page.
  *
  * @fires wcadmin_pfw_documentation_link_click with `{ link_id: 'terms-of-service', context: 'welcome-section' }`
+ * @fires wcadmin_pfw_setup with `{ target: 'onboarding', trigger: 'get-started' }` when "Get started" button is clicked for incomplete setup.
  *
  * @return {JSX.Element} Rendered element.
  */
 const WelcomeSection = () => {
+	const handleGetStarted = () => {
+		if ( ! wcSettings.pinterest_for_woocommerce.isSetupComplete ) {
+			recordEvent( 'pfw_setup', {
+				target: 'onboarding',
+				trigger: 'get-started',
+			} );
+		}
+		getHistory().push(
+			getNewPath(
+				{},
+				wcSettings.pinterest_for_woocommerce.isSetupComplete
+					? '/pinterest/catalog'
+					: '/pinterest/onboarding'
+			)
+		);
+	};
 	return (
 		<Card className="woocommerce-table pinterest-for-woocommerce-landing-page__welcome-section">
 			<Flex>
@@ -53,20 +80,7 @@ const WelcomeSection = () => {
 					</Text>
 
 					<Text variant="body">
-						<Button
-							isPrimary
-							onClick={ () =>
-								getHistory().push(
-									getNewPath(
-										{},
-										wcSettings.pinterest_for_woocommerce
-											.isSetupComplete
-											? '/pinterest/catalog'
-											: '/pinterest/onboarding'
-									)
-								)
-							}
-						>
+						<Button isPrimary onClick={ handleGetStarted }>
 							{ __( 'Get started', 'pinterest-for-woocommerce' ) }
 						</Button>
 					</Text>
--- a/assets/source/setup-guide/app/views/LandingPageApp.test.js
+++ b/assets/source/setup-guide/app/views/LandingPageApp.test.js
@@ -1,0 +1,90 @@
+jest.mock( '@woocommerce/tracks' );
+
+/**
+ * External dependencies
+ */
+import { recordEvent } from '@woocommerce/tracks';
+import { fireEvent, render } from '@testing-library/react';
+
+/**
+ * Internal dependencies
+ */
+import LandingPageApp from './LandingPageApp';
+import '../../../tests/custom-matchers';
+
+recordEvent.mockName( 'recordEvent' );
+jest.mock( '@woocommerce/settings', () => ( {
+	getSetting: jest
+		.fn()
+		.mockName( 'getSetting' )
+		.mockReturnValue( { es: 'Spain' } ),
+} ) );
+
+afterEach( () => {
+	jest.clearAllMocks();
+} );
+
+describe( 'LandingPageApp component', () => {
+	describe( 'when rendered', () => {
+		let getByRole;
+		beforeEach( () => {
+			// Render connected component.
+			getByRole = render( <LandingPageApp /> ).getByRole;
+		} );
+		describe( 'while `wcSettings.pinterest_for_woocommerce.isSetupComplete = false`', () => {
+			beforeEach( () => {
+				wcSettings.pinterest_for_woocommerce.isSetupComplete = false;
+			} );
+			describe( 'once "Get started" button is clicked', () => {
+				beforeEach( () => {
+					// Find & click the button.
+					const getStartedButton = getByRole( 'button', {
+						name: 'Get started',
+					} );
+					fireEvent.click( getStartedButton );
+				} );
+
+				it( "should call `wcadmin_pfw_setup { target: 'onboarding', trigger: 'get-started' }` track event", () => {
+					expect( recordEvent ).toHaveBeenCalledWith( 'pfw_setup', {
+						target: 'onboarding',
+						trigger: 'get-started',
+					} );
+				} );
+
+				it( 'should navigate to `/pinterest/onboarding`', () => {
+					expect( window.location ).toContainURLSearchParam(
+						'path',
+						'/pinterest/onboarding'
+					);
+				} );
+			} );
+		} );
+		describe( 'while `wcSettings.pinterest_for_woocommerce.isSetupComplete = true`', () => {
+			beforeEach( () => {
+				wcSettings.pinterest_for_woocommerce.isSetupComplete = true;
+			} );
+			describe( 'once "Get started" button is clicked', () => {
+				beforeEach( () => {
+					// Find & click the button.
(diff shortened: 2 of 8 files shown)


assets/source/setup-guide/app/views/LandingPageApp.js
```diff
@@ -53,20 +80,7 @@ const WelcomeSection = () => {
...
-										{},
-										wcSettings.pinterest_for_woocommerce
-											.isSetupComplete
-											? '/pinterest/catalog'
-											: '/pinterest/onboarding'
-									)
-								)
-							}
-						>
+						<Button isPrimary onClick={ handleGetStarted }>
```
puntope, line 83: 👍 

Appreciated the bit of refactoring instead of adding it here directly,  thanks!

assets/source/setup-guide/app/views/LandingPageApp.test.js
```diff
@@ -0,0 +1,90 @@
...
+ * External dependencies
+ */
+import { recordEvent } from '@woocommerce/tracks';
+import { fireEvent, render } from '@testing-library/react';
+
+/**
+ * Internal dependencies
+ */
+import LandingPageApp from './LandingPageApp';
+import '../../../tests/custom-matchers';
```
puntope, line 13: 💅 

Totally out scope. But the more I see the code in pfw and gla, the more I'm wondering about  using[ jsconfig paths ](https://intellij-support.jetbrains.com/hc/en-us/community/posts/360002490579-Custom-import-paths-jsconfig-json)

We can have something like `~tests/custom-matchers`, `~components/my-component`, `~data/reducer`... etc Instead of `../../../my-stuff`

assets/source/setup-guide/app/views/LandingPageApp.test.js
```diff
@@ -0,0 +1,90 @@
...
+		} );
+		describe( 'while `wcSettings.pinterest_for_woocommerce.isSetupComplete = false`', () => {
+			beforeEach( () => {
+				wcSettings.pinterest_for_woocommerce.isSetupComplete = false;
+			} );
+			describe( 'once "Get started" button is clicked', () => {
+				beforeEach( () => {
+					// Find & click the button.
+					const getStartedButton = getByRole( 'button', {
+						name: 'Get started',
```
puntope, line 42: til

jest.config.js
```diff
@@ -30,6 +30,7 @@ module.exports = {
 	globals: {
 		wcSettings: {
 			pinterest_for_woocommerce: {
+				pluginVersion: '1.2.3',
```
puntope, line 34: ❓ 

Why is this for?

assets/source/tests/custom-matchers.js
```diff
@@ -0,0 +1,49 @@
...
+	/**
+	 * Custom matcher to check the presence and optionally the value of a search query param in the given location.
+	 *
+	 * @param {Location} receivedLocation Location to be checked, for example `window.location`.
+	 * @param {string} param Parameter to check the presence of.
+	 * @param {string} [value] Value of the parameter to be checked.
+	 * @return {{message: string, pass: boolean}} Jest compatible matcher output.
+	 */
+	toContainURLSearchParam( receivedLocation, param, value ) {
+		const hasValue = arguments.length === 3;
```
puntope, line 11: 💅 

I guess this for allowing falsy values like null or undefined as the expected param value?

puntope (approved):
✅  LGTM 

- Check the happy path and worked as expected ✅ 
- Check Try again and worked as expected ✅ 
- Check Advertisers and worked as expected ✅ 
- Checked the tests and the utils ✅  

💅 / ❓ 
Not too much to add here. Just a few comments/questions regarding clarifications and opinions

Good work

puntope:
⚠️ 
> other steps are not clickable due to Steps 2,3 are not clickable in onboarding wizard  #307, but should fire the event once enabled.

This was not tested on my side

puntope:
> I think the custom matcher added here, could be eventually offloaded to an external package. Like @woocommerce/test-utils?

💅 

I like the idea, it looks promising, but for promoting it as an external package I think we should be more strict with how we test actually the package. Specially since it's a testing package supposed to 100% reliable, so, in case of bug it could impact in other tests as well in cascade. (Not saying that it is buggy this one, just missing the tests for it)

puntope:
> and probably some cleanup in the those bigger components

💅 
Agree.  For the record... after checking ... it's a component of around 550 lines with a lot of "deep-nested" conditionals and components. So, definitely a refactor would help to maintain and test it.

jest.config.js
tomalec: `LandinPageApp` requires the presence of `wcSettings.pinterest_for_woocommerce.pluginVersion` to run:
https://github.com/woocommerce/pinterest-for-woocommerce/blob/add/232-setup-event/assets/source/setup-guide/app/views/LandingPageApp.js#L271:L271

assets/source/tests/custom-matchers.js
tomalec: Yes, to allow "expect the location to have URL param `path` of `/some/thing`" as well as just  "expect the location to have URL param `redirect`". This is to follow the convention of [`toHaveProperty`](https://jestjs.io/docs/expect#tohavepropertykeypath-value)

> Custom matcher to check **the presence and optionally the value** of a search query param in the given location.

Do you think it could be stated more clearly?

tomalec:
> but for promoting it as an external package I think we should be more strict with how we test actually the package. Specially since it's a testing package supposed to 100% reliable, so, in case of bug it could impact in other tests as well in cascade. (Not saying that it is buggy this one, just missing the tests for it)

I totally agree, that once we separate it as an external package, we should write tests for that package.
(FWIW, https://github.com/WordPress/gutenberg/tree/51454d4ef88680d3479518b057883cbf06d2a833/packages/e2e-test-utils does not have their own tests, but I believe we can do better here)

assets/source/setup-guide/app/views/LandingPageApp.test.js
tomalec: Oh, I thought we have it here, but it's only set for jest config https://github.com/woocommerce/pinterest-for-woocommerce/blob/add/232-setup-event/jest.config.js#L24:L24, but not actually aliased in webpack config.

I also vote to make the change we did in GLA (https://github.com/woocommerce/google-listings-and-ads/pull/192) here

tomalec: But, I'd rather make this change after we merge the tracking feature branch, to avoid merge conflicts.

tomalec: Created an issue at https://github.com/woocommerce/pinterest-for-woocommerce/issues/310

tomalec:
Thanks for the review @puntope. I believe I addressed all your comments. Is there anything else, more left?

assets/source/tests/custom-matchers.js
puntope: Was more about 

`const hasValue = arguments.length === 3;` instead of just checking `value` prop directly.

puntope:
> Thanks for the review @puntope. I believe I addressed all your comments. Is there anything else, more left?

This is ready for merge IMO :)

PR #312, opened 2021-12-28
Refactor `AccountConnection` component.

tomalec:
### Current look

Account Connection card has four distinct states

- connected
- connected but no `accountData`/`id` (yet or at all)
- connection state unknown
- not connected

(see screenshots below)

I believe there are a few things we can improve from a UX perspective (and one DevX).
I propose to introduce the following changes:

- Show the Pinterest logo, regardless of the state, to reduce the number of moving parts.
- Show the "Disconnect" button even if there is no account data, or before it's loaded. To help users disconnect even if some account data gets corrupted, or simply without waiting for everything to load.
- Show placeholder instead of a spinner when we wait only for the account details.
- Set the minimum height of the card's content to `40px` = the height of the spinner, to reduce the difference between states, and reduce the layout shift when the spinner disappears and the content is loaded.
- Move the spinner to the middle, to let the user anticipate both things to appear: left-aligned text and right-aligned button.
- Make the code DRY, by moving repeated markup outside of conditions.

@shivanarrthine  I'd appreciate your opinion here :)

#### Before 

before Screenshot from 2021-12-28 12-02-48

#### After

Screenshot from 2021-12-28 16-08-03

1. Go to the `/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Fconnection`
2. Watch the loading state
3. Disconnect account
4. Watch connect state for a second before get redirected to welcome page (BTW, is that redirect desired here?)
5. Click "Get started" to go to `/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Fonboarding`
6. Connect it there,
7. Get back to the first step of onboarding stepper
7. See the connected state

- I was not able to reproduce the preloading of connection status other than programmatically, by setting `isConnected` to false manually in https://github.com/woocommerce/pinterest-for-woocommerce/blob/update/account-connection/assets/source/setup-guide/app/steps/SetupAccount.js#L174:L174

### Changelog entry

> Tweak - Improved preloading state of Account Connection Card

### DIFF
--- a/assets/source/setup-guide/app/style.scss
+++ b/assets/source/setup-guide/app/style.scss
@@ -258,7 +258,7 @@ $root: ".woocommerce-setup-guide";
 
 				.connection-info {
 					// Set minimum height to the height of the spinner preloader.
-					min-height: calc( 40px );
+					min-height: calc(40px);
 
 					.logo {
 						display: flex;
@@ -266,6 +266,7 @@ $root: ".woocommerce-setup-guide";
 					}
 
 					.connection-info__placeholder {
+
 						@include placeholder();
 						width: 50%;
 						min-width: 10em;
--- a/assets/source/setup-guide/app/style.scss
+++ b/assets/source/setup-guide/app/style.scss
@@ -4,6 +4,23 @@ $root: ".woocommerce-setup-guide";
 @import "~@wordpress/base-styles/mixins";
 @import "~@wordpress/base-styles/breakpoints";
 
+// Adds animation to placeholder section
+// Copied from WC-admin
+// https://github.com/woocommerce/woocommerce-admin/blob/9f68059c7df466ee7a559c555bcb5a7b9e421e6d/packages/style-build/abstracts/_mixins.scss#L22-L35
+@mixin placeholder( $lighten-percentage: 30% ) {
+	animation: loading-fade 1.6s ease-in-out infinite;
+	background-color: $gray-100;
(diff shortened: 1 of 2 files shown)


assets/source/setup-guide/app/style.scss
```diff
@@ -240,9 +257,24 @@ $root: ".woocommerce-setup-guide";
 			&__setup-account {
 
 				.connection-info {
+					// Set minimum height to the height of the spinner preloader.
+					min-height: calc(40px);
```
eason9487: ❔ 
Not sure why it needs `calc`? It's a leftover after finding the height of spinner is not exposed? 🤔

eason9487 (approved):
LGTM. Left a comment for an unsure thing but it should be a non-blocking item no matter keep or remove it.

assets/source/setup-guide/app/style.scss
tomalec: Thanks for catching this, yes that was a leftover. Fixed in https://github.com/woocommerce/pinterest-for-woocommerce/pull/312/commits/839abef87e4803abdc89becd3dc0cda239523974

shivanarrthine:
Hey @tomalec, this looks great! Definitely in agreement with the proposed changes - this would help improve the UX of the component and create some consistency between the states.

tomalec:
Thanks for the reviews :)

PR #318, opened 2022-01-10
Fix/307 steps 2 3 are not clickable in onboarding wizard

### LINKED ISSUE #307: Steps 2,3 are not clickable in onboarding wizard

jorgemd24:

Closes #307.

In this PR I am proposing to fix the issue related to the stepper header that was not clickable on the onboarding wizard. Because of this issue, the tracking events added #308 were not possible to be tested with the onboarding steps buttons thus I am adding additional tests to check this functionality.

### Screenshots:

https://user-images.githubusercontent.com/2488994/148749165-d9dcb0fa-7103-426f-9050-bed44877c035.mp4

### Prerequisites: 

- Disconnect your account. You can do it going to: /wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Fconnection and click disconnect in the Linked account section

1. Go to [/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Fonboarding](/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Flanding)
2. Refresh the page, steps 1,2 and 3 should not be clickable.
3. Connect your account and click continue.
4. Steps 1 and 2 should be clickable. (If your domain was already verified from an older setup, step 3 should be clickable)
5. Go back to step 1.
6. Click step 2 in the header.
7. Complete step 2, click "Continue".
8. Try going back to step 2 or 1 by clicking in the stepper header.
9. Try going to step 3 by clicking in the stepper header.
10. Complete the setup

### Changelog entry

> Fix - Clicking buttons in the stepper header

### DIFF
--- a/assets/source/setup-guide/app/views/WizardApp.js
+++ b/assets/source/setup-guide/app/views/WizardApp.js
@@ -11,6 +11,7 @@ import {
 	getQuery,
 	updateQueryString,
 } from '@woocommerce/navigation';
+import { useRef } from 'react';
 
 /**
  * Internal dependencies
@@ -45,13 +46,22 @@ const WizardApp = () => {
 		wcSettings.pinterest_for_woocommerce.isBusinessConnected
 	);
 
+	const appSettings = useSettingsSelect();
+	const isDomainVerified = useSettingsSelect( 'isDomainVerified' );
+	const isTrackingConfigured = useSettingsSelect( 'isTrackingConfigured' );
+	const mounted = useRef( false );
+
 	useEffect( () => {
+		mounted.current = true;
+
 		if ( ! isConnected ) {
 			setIsBusinessConnected( false );
 		}
-	}, [ isConnected, setIsBusinessConnected ] );
 
-	const appSettings = useSettingsSelect();
+		return () => {
+			mounted.current = false;
+		};
+	}, [ isConnected, setIsBusinessConnected ] );
 
 	useBodyClasses( 'wizard' );
 	useCreateNotice()( wcSettings.pinterest_for_woocommerce.error );
@@ -70,11 +80,13 @@ const WizardApp = () => {
 				setIsBusinessConnected,
 				isBusinessConnected,
 			},
+			isCompleted: isBusinessConnected,
 		},
 		{
 			key: 'claim-website',
 			container: ClaimWebsite,
 			label: __( 'Claim your website', 'pinterest-for-woocommerce' ),
+			isCompleted: isDomainVerified,
 		},
 		{
 			key: 'setup-tracking',
@@ -83,11 +95,12 @@ const WizardApp = () => {
 				'Track conversions with the Pinterest tag',
 				'pinterest-for-woocommerce'
 			),
+			isCompleted: isTrackingConfigured,
 		},
 	];
 
 	const getSteps = () => {
-		return steps.map( ( step, index ) => {
+		return steps.map( ( step ) => {
 			const container = createElement( step.container, {
 				query: getQuery(),
 				step,
@@ -104,8 +117,7 @@ const WizardApp = () => {
 				</div>
 			);
 
-			const previousStep = steps[ index - 1 ];
-			if ( ! previousStep || previousStep.isComplete ) {
+			if ( step.isCompleted ) {
 				step.onClick = ( key ) => {
 					recordEvent( 'pfw_setup', {
 						target: key,
@@ -147,7 +159,10 @@ const WizardApp = () => {
 	};
 
 	getHistory().listen( () => {
-		setCurrentStep( getCurrentStep() );
+		//prevent to update the state on an unmounted component
+		if ( mounted.current ) {
+			setCurrentStep( getCurrentStep() );
+		}
 	} );
 
 	if ( ! currentStep ) {
--- a/assets/source/setup-guide/app/views/WizardApp.js
+++ b/assets/source/setup-guide/app/views/WizardApp.js
@@ -47,7 +47,6 @@ const WizardApp = () => {
 
 	const appSettings = useSettingsSelect();
 	const isDomainVerified = useSettingsSelect( 'isDomainVerified' );
-	const isTrackingConfigured = useSettingsSelect( 'isTrackingConfigured' );
 
 	//Subscribe history listening
 	useEffect( () => {
@@ -82,13 +81,13 @@ const WizardApp = () => {
 				setIsBusinessConnected,
 				isBusinessConnected,
 			},
-			isClickable: isBusinessConnected,
+			isClickable: true,
 		},
 		{
 			key: 'claim-website',
 			container: ClaimWebsite,
 			label: __( 'Claim your website', 'pinterest-for-woocommerce' ),
-			isClickable: isBusinessConnected || isDomainVerified,
+			isClickable: isBusinessConnected,
 		},
 		{
 			key: 'setup-tracking',
@@ -97,7 +96,7 @@ const WizardApp = () => {
 				'Track conversions with the Pinterest tag',
 				'pinterest-for-woocommerce'
 			),
-			isClickable: isDomainVerified || isTrackingConfigured,
+			isClickable: isBusinessConnected && isDomainVerified,
 		},
 	];
 
--- a/assets/source/setup-guide/app/views/WizardApp.js
+++ b/assets/source/setup-guide/app/views/WizardApp.js
@@ -11,7 +11,6 @@ import {
 	getQuery,
 	updateQueryString,
 } from '@woocommerce/navigation';
-import { useRef } from 'react';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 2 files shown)


assets/source/setup-guide/app/views/WizardApp.js
```diff
@@ -70,11 +82,13 @@ const WizardApp = () => {
 				setIsBusinessConnected,
 				isBusinessConnected,
 			},
+			isClickable: isBusinessConnected,
```
eason9487: ❓  I think step 1 should be always clickable no matter user's account has been connected or not?

assets/source/setup-guide/app/views/WizardApp.js
```diff
@@ -70,11 +82,13 @@ const WizardApp = () => {
...
 		},
 		{
 			key: 'claim-website',
 			container: ClaimWebsite,
 			label: __( 'Claim your website', 'pinterest-for-woocommerce' ),
+			isClickable: isBusinessConnected || isDomainVerified,
```
eason9487: ❓  Should it only check `isBusinessConnected`? Otherwise, it can switch to step 2 before connecting the account.

Kapture 2022-01-13 at 18 21 27

assets/source/setup-guide/app/views/WizardApp.js
```diff
@@ -83,11 +97,12 @@ const WizardApp = () => {
 				'Track conversions with the Pinterest tag',
 				'pinterest-for-woocommerce'
 			),
+			isClickable: isDomainVerified || isTrackingConfigured,
```
eason9487: ❓  I think this condition should be `isBusinessConnected && isDomainVerified`. Or it can switch to step 3 before completing steps 1 and 2.

Kapture 2022-01-13 at 18 23 39

assets/source/setup-guide/app/views/WizardApp.js
```diff
@@ -45,14 +45,26 @@ const WizardApp = () => {
...
+
+	//Subscribe history listening
+	useEffect( () => {
+		const unlisten = getHistory().listen( () => {
+			setCurrentStep( getQuery().step );
+		} );
+
+		//cleanup function to avoid memory leaks
+		return unlisten;
+	}, [] );
```
eason9487: 💅 
The internal state `currentStep` was a bit redundant before this PR because the same state `step` is kept in the URL query.

We could consider simplifying the state by using the `step` query as the single source by removing this block, and replacing the following code by `const currentStep = getCurrentStep();`:

https://github.com/woocommerce/pinterest-for-woocommerce/blob/cc2b31bbdbaf02f2e3a5e345f8e954a8b541131d/assets/source/setup-guide/app/views/WizardApp.js#L163-L167

assets/source/setup-guide/app/views/WizardApp.test.js
```diff
@@ -0,0 +1,188 @@
...
+jest.mock( '../steps/SetupAccount', () => () => null );
+jest.mock( '../steps/ClaimWebsite', () => () => null );
+jest.mock( '../steps/SetupTracking', () => () => null );
+
+const stepOne = /Set up your business account/;
+const stepTwo = /Claim your website/;
+const stepThree = /Track conversions/;
+
+afterEach( () => {
+	jest.clearAllMocks();
```
eason9487: ❔  I am curious about why this is needed? 🤔

assets/source/setup-guide/app/views/WizardApp.js
jorgemd24: I have set the value to true. 👍 

 See [9760a4c](https://github.com/woocommerce/pinterest-for-woocommerce/pull/318/commits/9760a4ce5078a5c874b86efa8216668aa72f710f).

assets/source/setup-guide/app/views/WizardApp.js
jorgemd24: I am not sure how you could get a verified domain without connecting your account? When I see the code in isDomainVerified hook

```		
export const isDomainVerified = ( state ) => {
	if ( undefined === state?.settings?.account_data ) {
		return;
	}
```
I see that is returning undefined if the account_data is undefined. If I try to reproduce it I get the following response from the API:

image

`account_data` is null therefore isDomainVerified does not return `true`. 

In case a user can have this situation, I updated the code see: [9760a4c.](https://github.com/woocommerce/pinterest-for-woocommerce/pull/318/commits/9760a4ce5078a5c874b86efa8216668aa72f710f)

assets/source/setup-guide/app/views/WizardApp.js
jorgemd24: See [9760a4c](https://github.com/woocommerce/pinterest-for-woocommerce/pull/318/commits/9760a4ce5078a5c874b86efa8216668aa72f710f).

assets/source/setup-guide/app/views/WizardApp.js
jorgemd24: If I understand correctly we could simplify the code doing the following steps:

1. We could remove  `const [ currentStep, setCurrentStep ] = useState();`  and  
 ```
useEffect( () => {
		const unlisten = getHistory().listen( () => {
			setCurrentStep( getQuery().step );
		} );
		//cleanup function to avoid memory leaks
		return unlisten;
	}, [] ); 
```


2. Replacing:
``` 
   if ( ! currentStep ) { 
 	setCurrentStep( getCurrentStep() ); 
 	return <Spinner />; 
 }
``` 
with `const currentStep = getCurrentStep();`

3. I guess that the function `updateQueryString( { step: key } ); `  will trigger a new rendering and `const currentStep ` will be assigned with the new step.

Let me know your thoughts. Thank you!

File: assets/source/setup-guide/app/views/WizardApp.test.js (hunk 5 above)
Comment (jorgemd24) [reply assets/source/setup-guide/app/views/WizardApp.test.js]: Thanks for pointing this out. I have double-checked and I see that is unnecessary as we are not really asserting any `fn.mock.calls` or similar.  

See [801bf25](https://github.com/woocommerce/pinterest-for-woocommerce/pull/318/commits/801bf25bb200d35f42c44583e1a4ff5a559bb318)

File: assets/source/setup-guide/app/views/WizardApp.js (hunk 2 above)
Comment (eason9487) [reply assets/source/setup-guide/app/views/WizardApp.js]: > I am not sure how you could get a verified domain without connecting your account?

General operations on UI won't get into this case.

Sorry 🙏  I should have added these notes. I made some changes locally to show that the order of dependence between steps should be a more appropriate condition in step 3, and the `isDomainVerified` should be logically unneeded in step 2.

File: assets/source/setup-guide/app/views/WizardApp.js (hunk 4 above)
Comment (eason9487) [reply assets/source/setup-guide/app/views/WizardApp.js]: > 3\. I guess that the function `updateQueryString( { step: key } ); `  will trigger a new rendering and `const currentStep ` will be assigned with the new step.

Yes, the client-side routing watches the URL changes and passes down parsed URL query to page components as one of props. So it will trigger re-render after updating to URL query.

`<WizardApp>` happens to be [one of the top-level page components](https://github.com/woocommerce/pinterest-for-woocommerce/blob/801bf25bb200d35f42c44583e1a4ff5a559bb318/assets/source/setup-guide/index.js#L65-L75), and the following code could inspect the `query` prop passing in from the routing handler.

```js
const WizardApp = ( { query } ) => {
	console.log( query );
	// ...omitted.
```

In other words, replacing `getQuery()` with `query` prop could further simplify this component. 😎

eason9487 (approved):
Thanks for the updates! LGTM.

We could have another round of code review quickly if we want to make changes for the non-blocking https://github.com/woocommerce/pinterest-for-woocommerce/pull/318#discussion_r783847828.

assets/source/setup-guide/app/views/WizardApp.js
jorgemd24: Thanks, @eason9487 for the clarification! See the changes here: [82bae3d](https://github.com/woocommerce/pinterest-for-woocommerce/pull/318/commits/82bae3d774a2d1c43fe578d6e9c9e989d66b441b)

assets/source/setup-guide/app/views/WizardApp.js
```diff
@@ -29,14 +25,16 @@ import {
 /**
  * Onboarding Wizard component.
  *
+ * @param {Object} props React props.
+ * @param {Object} [props.query] The current query string, parsed into an object, from history.
```
eason9487: it'd be nicer to specify `query` is not an optional prop and comes from page URL.

```suggestion
 * @param {Object} props.query The current query string, parsed into an object, from the page URL.
```

eason9487 (approved):
Thanks for the further adjustments! LGTM. The JSDoc could be polished a little bit and no need to block the merging.

assets/source/setup-guide/app/views/WizardApp.js
jorgemd24: Perfect here is the JSDoc updated: [8bb0e7](https://github.com/woocommerce/pinterest-for-woocommerce/pull/318/commits/8bb0e7fe820186b2bf2e48198ddd667ac9462cc9)

PR #320, opened 2022-01-10
Fetch parent id for variable product type xml.

budzanowski:
### Changes proposed in this Pull Request:

Closes #241 .

Check for parent product during `g:product_type` generation.

### Detailed test instructions:

1. Add variable products with variations.
2. Set product type.
3. Observe the XML for product type in the variable product section - it should be there.

### Changelog entry
Fix - Add product type for variation product in XML feed file.
>

PR #321, opened 2022-01-10
Add Feed file generation UT.

budzanowski:

Add unit tests for feed file generation. It will be useful when we will add more and more features to the XML.

The skipped test waits for the #320. We can reenable it when that PR will get merged.

### Changelog entry

> Add - Unit tests for feed file generation.


tests/Unit/FeedXMLGenerationTest.php
```diff
@@ -0,0 +1,405 @@
+<?php
+
+namespace Automattic\WooCommerce\Pinterest;
```
mikkamp: Would it be better to run this in a separate namespace and then to import the class names we need with use statements? Like [here](https://github.com/woocommerce/google-listings-and-ads/blob/develop/tests/Unit/MerchantCenter/ContactInformationTest.php#L4)

I think that will make it easier to load mocked versions of any classes we need without conflicting namespaces.

tests/Unit/FeedXMLGenerationTest.php
```diff
@@ -0,0 +1,405 @@
...
+/**
+ * Feed file generation testing class.
+ */
+class Pinterest_Test_Feed extends \WC_Unit_Test_Case {
```
mikkamp: Can we import this class so we don't need to prefix it with a `\`, ex:

`use WC_Unit_Test_Case;`

tests/Unit/FeedXMLGenerationTest.php
```diff
@@ -0,0 +1,405 @@
...
+
+	/**
+	 * @group feed
+	 */
+	public function testDescriptionForVariableProductXML() {
+		$description_method = $this->getProductsXmlFeedAttributeMethod( 'description' );
+
+		// By passing manually created Variable Product the create_variation_product will add children to it.
+		$product            = new \WC_Product_Variable();
+		$variation_product  = \WC_Helper_Product::create_variation_product( $product );
```
mikkamp: Can we also import these class names at the top (several other locations where we can then eliminate the trailing `\`)

tests/Unit/FeedXMLGenerationTest.php
```diff
@@ -0,0 +1,405 @@
...
+	}
+
+	/**
+	 * Helper function for extracting the static private members of the ProductsXmlFeed class.
+	 * Gets the property method then just pass the product and voila.
+	 *
+	 * @param string $attribute
+	 * @return function
+	 */
+	public function getProductsXmlFeedAttributeMethod( $attribute ) {
```
mikkamp: Can we make this protected / private or move it to a trait so we don't mix this us as one of the test methods?

tests/Unit/FeedXMLGenerationTest.php
```diff
@@ -0,0 +1,405 @@
...
+		$method_name = 'get_property_' . str_replace( ':', '_', $attribute );
+		$class       = new \ReflectionClass('Automattic\WooCommerce\Pinterest\ProductsXmlFeed');
```
mikkamp: Can we import both these classes at the top and use the `::class` so we don't need to worry about what namespace they are coming from here.

```suggestion
		$class       = new ReflectionClass( ProductsXmlFeed::class );
```

tests/Unit/FeedXMLGenerationTest.php
```diff
@@ -0,0 +1,405 @@
...
+		$method      = $class->getMethod( $method_name );
+		$method->setAccessible(true);
```
mikkamp: Minor: missing spacing around brackets.

tests/Unit/FeedXMLGenerationTest.php
```diff
@@ -0,0 +1,405 @@
...
+		,   'post_title'     => 'product image 2'
+		);
+		$attachment_id_2 = wp_insert_attachment( $attachment, 'product_image_2.png', $product->get_id() );
+
+		// Add attachment id as product image id.
+		$product->set_gallery_image_ids( [ $attachment_id_1, $attachment_id_2 ] );
+		$product->save();
+
+		$xml = $additional_image_link_method( $product );
+		$this->assertEquals( "<g:additional_image_link><![CDATA[http://example.org/wp-content/uploads/product_image_1.png,http://example.org/wp-content/uploads/product_image_2.png]]></g:additional_image_link>", $xml );
```
mikkamp: We can use single quotes here (same for any of the other tests where we aren't adding any control characters or variables).

tests/Unit/FeedXMLGenerationTest.php
```diff
@@ -0,0 +1,405 @@
...
+		$product                      = \WC_Helper_Product::create_simple_product();
+
+		$xml = $additional_image_link_method( $product );
+		// By default no galery images are set.
+		$this->assertEquals( "", $xml );
+
+		// Add dummy image entry.
+		$attachment = array(
+			'post_mime_type' => 'image/png'
+		,   'post_title'     => 'product image 1'
```
mikkamp: Minor: fix alignment and comma at the end of each line

tests/Unit/FeedXMLGenerationTest.php
```diff
@@ -0,0 +1,405 @@
...
+		);
+		$attachment_id_1 = wp_insert_attachment( $attachment, 'product_image_1.png', $product->get_id() );
+		// Product needs main image to use gallery so lets set this up here.
+		$product->set_image_id( $attachment_id_1 );
+
+		// Add second dummy image entry.
+		$attachment_id_2 = array(
+			'post_mime_type' => 'image/png'
+		,   'post_title'     => 'product image 2'
```
mikkamp: Minor: fix alignment and comma at the end of each line

tests/Unit/FeedXMLGenerationTest.php
```diff
@@ -0,0 +1,405 @@
...
+			)
+		);
+		$xml     = $sale_price_method( $product );
+		$this->assertEquals( "<sale_price>5USD</sale_price>", $xml );
+	}
+
+	/**
+	 * @group feed
+	 */
+	public function testPropertySalePriceVariableProductXML() {
```
mikkamp: We don't seem to be testing the sale price of the variable product here.

tests/Unit/FeedXMLGenerationTest.php
```diff
@@ -0,0 +1,405 @@
...
+		$variation_product = \WC_Helper_Product::create_variation_product( $product );
+		/* In UT flow we need to fetch the product again from the DB after creation.
+		 * This ensures correct initialization of visible variations.
+		 * Without that the variable price methods think that we don't have visible children.
+		 * Quirk of create_variation_product.
+		 */
+		$product = wc_get_product( $variation_product->get_id() );
+		$xml     = $price_method( $product );
+		// 10.00USD is the cheapest variation created by create_variation_product
+		$this->assertEquals( "<g:price>10.00USD</g:price>", $xml );
```
mikkamp: For a simple product we expect the price to be shown without decimals `15USD` whereas for a variable product we are testing it with decimals `10.00USD`. Shouldn't those both be in the same format in the feed?

tests/Unit/FeedXMLGenerationTest.php
```diff
@@ -0,0 +1,405 @@
...
+	/**
+	 * @group feed
+	 */
+	public function testPropertyAdditionalImageLinkXML() {
+		$additional_image_link_method = $this->getProductsXmlFeedAttributeMethod( 'g:additional_image_link' );
+		$product                      = \WC_Helper_Product::create_simple_product();
+
+		$xml = $additional_image_link_method( $product );
+		// By default no galery images are set.
+		$this->assertEquals( "", $xml );
```
mikkamp: Single quotes will be fine here.

tests/Unit/FeedXMLGenerationTest.php
```diff
@@ -0,0 +1,405 @@
...
+	 */
+	public function testPropertyProductTypeVariableProductXML() {
+		$product_type_method  = $this->getProductsXmlFeedAttributeMethod( 'g:product_type' );
+		$product              = new \WC_Product_Variable();
+		$variation_product    = \WC_Helper_Product::create_variation_product( $product );
+		$variation_product_id = $variation_product->get_children()[0];
+		$variation_product    = wc_get_product( $variation_product_id );
+		$xml                  = $product_type_method( $variation_product );
+		// create_simple_product gives the product 'Uncategorized' type.
+		$this->markTestSkipped( "Wating for https://github.com/woocommerce/pinterest-for-woocommerce/pull/320");
```
mikkamp: Can be removed now since #320 is merged.

tests/Unit/FeedXMLGenerationTest.php
```diff
@@ -0,0 +1,405 @@
...
+	 */
+	public function testDescriptionForSimpleProductXML() {
+		$description_method = $this->getProductsXmlFeedAttributeMethod( 'description' );
+
+		// No description set.
+		$product  = \WC_Helper_Product::create_simple_product();
+		$xml      = $description_method( $product );
+		$this->assertEquals( '', $xml );
+
+		$desc = 'Test description.';
```
mikkamp, line 128: Do we want to test with a little more complicated description containing html attributes or text that should be escaped?

mikkamp (commented):
Thanks for adding the feed tests. They work well, and the tests pass.

Just added some comments about improving some of the code.

tests/Unit/FeedXMLGenerationTest.php
budzanowski: Done.

tests/Unit/FeedXMLGenerationTest.php
budzanowski: We should but glancing over the `get_property_description` function it looks like this is not handled. I have added an issue to handle that and to modifiy the tests when the new feature will be added: https://github.com/woocommerce/pinterest-for-woocommerce/issues/323

tests/Unit/FeedXMLGenerationTest.php
budzanowski: Done.

tests/Unit/FeedXMLGenerationTest.php
budzanowski: Yep.

tests/Unit/FeedXMLGenerationTest.php
budzanowski: Good catch. Prices in case of `get_variation_regular_price` go through `wc_format_decimal`. I have created an issue to review and tackle this: https://github.com/woocommerce/pinterest-for-woocommerce/issues/325

budzanowski:
@mikkamp all suggestions addressed.

budzanowski:
If this is OK I can merge and start with tests for things added lately. I think that doing that in a separate PR will make things a bit cleaner and separate.

budzanowski:
https://github.com/woocommerce/pinterest-for-woocommerce/pull/326 has been created to test the new attributes capability of the Feed file.

tests/Unit/FeedXMLGenerationTest.php
```diff
@@ -0,0 +1,411 @@
+<?php
+
+namespace Automattic\WooCommerce\Pinterest\Tests\Unit\Feed;
+
+use \WC_Unit_Test_Case;
+use \WC_Product_Variable;
+use \WC_Helper_Product;
+use \ReflectionClass;
```
mikkamp: Minor: The `\` prefix is not needed here and personally I think it looks a little cleaner to leave it out.

tests/Unit/FeedXMLGenerationTest.php
```diff
@@ -0,0 +1,411 @@
...
+		$product           = WC_Helper_Product::create_simple_product();
+
+		$xml = $image_link_method( $product );
+		// By default no image link is set.
+		$this->assertEquals( '', $xml );
+
+		// Add dummy image entry.
+		$attachment = array(
+			'post_mime_type' => 'image/png'
+		,   'post_title'     => 'product image'
```
mikkamp: Just need to fix the alignment and the comma here.

mikkamp (approved):
Thanks for all the changes and the follow up issues. It's looking good and the tests run successfully. I've just left two small comments, but I'll go ahead and approve it since it's pretty much ready to go.

PR #324, opened 2022-01-11
Fix/315 Pinterest plugin is blocking some scripts

### LINKED ISSUE #315: Pinterest plugin is blocking some scripts
### Describe the bug:
After activating and configuring the Pinterest plugin, some scripts stopped working on our site's product page. We have an integration with the Hero chat service which fires off an event tracking script on the product template using "window.onload". This script is ignored when the Pinterest plugin is activated

What Customer said: 

> I figured it out! Both the Pinterest plugin's script and my scripts are using "window.onload" instead of "window.addEventListener('load', ...)". As such, your window.onload function call is replacing mine.

I've updated my code accordingly and I suggest your team do the same. The script with the window.onload is located at:
pinterest-for-woocommerce/assets/js/pinterest-for-woocommerce-tracking.min.js

alexiglesias31:

Closes #315.

The plugin was using window.onload method, which can override the same method in other scripts.

1. Create a script that uses window.onload method.
2. The script execution should not be overriden by the plugin.

### DIFF
--- a/assets/source/js/pinterest-for-woocommerce-tracking.js
+++ b/assets/source/js/pinterest-for-woocommerce-tracking.js
@@ -4,7 +4,8 @@
  * Bind to search form submit in order to track the search event.
  * Site search should use <form role='search'> to integrate with Pinterest for WooCommerce search event automatically.
  */
-window.onload = function () {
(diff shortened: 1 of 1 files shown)


ksere (approved):
LGTM

PR #326, opened 2022-01-12
Add UT to test product attributes.

budzanowski:

Adds a new test case to check the presence of the product attributes ( if set ). So far we only support `condition` and `google_category` and that is what is being tested. We have two assertions: the first one when no attribute is set, the second one when we add an attribute.

Google test is more of an integration test to not duplicate what is actually being tested. 

This is built on top of https://github.com/woocommerce/pinterest-for-woocommerce/pull/321

Check the Ci results for confirmation that it works.

### Changelog entry

> Add - Product attribute unit tests.

### DIFF
--- a/tests/Unit/FeedXMLGenerationTest.php
+++ b/tests/Unit/FeedXMLGenerationTest.php
@@ -407,7 +407,7 @@ class Pinterest_Test_Feed extends WC_Unit_Test_Case {
 		$method->setAccessible( true );
 
 		$product = WC_Helper_Product::create_simple_product();
-		$xml = $method->invoke( null, $product, '\t' );
+		$xml = $method->invoke( null, $product, '' );
 		// No attributes set, output should be empty.
 		$this->assertEquals( '', $xml );
 
@@ -431,7 +431,7 @@ class Pinterest_Test_Feed extends WC_Unit_Test_Case {
 		$method->setAccessible( true );
 
 		$product = WC_Helper_Product::create_simple_product();
-		$xml = $method->invoke( null, $product, '\t' );
+		$xml = $method->invoke( null, $product, '' );
 		// No attributes set, output should be empty.
 		$this->assertEquals( '', $xml );
 
--- a/tests/Unit/FeedXMLGenerationTest.php
+++ b/tests/Unit/FeedXMLGenerationTest.php
@@ -440,7 +440,7 @@ class Pinterest_Test_Feed extends WC_Unit_Test_Case {
 		$taxonomy                  = GoogleProductTaxonomy::TAXONOMY[502979]; // Randomly selected category - i just made sure that it has parent.
 		$full_taxonomy_name        = $full_category_name_method->invoke( new GoogleCategorySearch(), $taxonomy );
 		$condition                 = new GoogleCategory( $full_taxonomy_name );
-		$attribute_manager = AttributeManager::instance();
+		$attribute_manager         = AttributeManager::instance();
 		$attribute_manager->update( $product, $condition );
 
 		$xml = $method->invoke( null, $product, '' );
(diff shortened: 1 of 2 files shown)


tests/Unit/FeedXMLGenerationTest.php
```diff
@@ -6,7 +6,15 @@
...
 use \WC_Helper_Product;
 use \ReflectionClass;
+
 use Automattic\WooCommerce\Pinterest\ProductsXmlFeed;
+use Automattic\WooCommerce\Pinterest\Product\GoogleProductTaxonomy;
+use Automattic\WooCommerce\Pinterest\Product\GoogleCategorySearch;
+use Automattic\WooCommerce\Pinterest\Product\Attributes\Condition;
+use Automattic\WooCommerce\Pinterest\Product\Attributes\GoogleCategory;
+use Automattic\WooCommerce\Pinterest\Product\Attributes\AttributeManager;
+use ReflectionMethod;
```
mikkamp: Not strictly necessary, but in other extensions we have kept the import statements in alphabetical order to provide some kind of organisation.

tests/Unit/FeedXMLGenerationTest.php
```diff
@@ -389,6 +399,55 @@ public function testPropertyAdditionalImageLinkXML() {
...
 
+	/**
+	 * @group feed
+	 */
+	public function testAttributesConditionXML() {
+		$method = ( new ReflectionClass( ProductsXmlFeed::class ) )->getMethod( 'get_attributes_xml' );
+		$method->setAccessible( true );
+
+		$product = WC_Helper_Product::create_simple_product();
+		$xml = $method->invoke( null, $product, '\t' );
```
mikkamp: With single quotes the indent won't get interpreted as a tab. Although in this test we can probably leave it as an empty string.

mikkamp: With single quotes the indent won't get interpreted as a tab. Although in this test we can probably leave it as an empty string.

tests/Unit/FeedXMLGenerationTest.php
```diff
@@ -389,6 +399,55 @@ public function testPropertyAdditionalImageLinkXML() {
...
+	 *
+	 * Since GoogleCategory is not validated we would end up with yet another test like testAttributesConditionXML.
+	 * To make this a bit more useful this will be more of an integration test.
+	 */
+	public function testAttributesGoogleCategoryXML() {
+		$method = ( new ReflectionClass( ProductsXmlFeed::class ) )->getMethod( 'get_attributes_xml' );
+		$method->setAccessible( true );
+
+		$product = WC_Helper_Product::create_simple_product();
+		$xml = $method->invoke( null, $product, '\t' );
```
mikkamp: Same here, we can pass an empty string for the indent.

tests/Unit/FeedXMLGenerationTest.php
```diff
@@ -389,6 +399,55 @@ public function testPropertyAdditionalImageLinkXML() {
...
+		// No attributes set, output should be empty.
+		$this->assertEquals( '', $xml );
+
+		$full_category_name_method = new ReflectionMethod( GoogleCategorySearch::class, 'full_category_name' );
```
mikkamp, line 438: Bit of a mixed feeling on this. I think the main purpose of this unit test is to test the generation of the product feed. Converting the category name and confirming the search functionality works should have a unit test of it's own, maybe another PR. To generate the feed we just get the full category name which is saved in the product meta, so we can just simply set the product meta when creating the product and confirm it's output in the feed.

mikkamp (approved):
Thanks for adding more tests, it looks good and the tests run without issues.

I'll approve it since it's running well, but I did leave a few comments which we might want to consider.

tests/Unit/FeedXMLGenerationTest.php
budzanowski: I also thought about this - hence the comment. Having separate tests for this makes 100% sense, but I thought that having a bit more coverage, even a bit sideways is still good.

PR #331, opened 2022-01-19
Add shipping collumn.

budzanowski:

This PR implements a shipping column for the Pinterest XML feed file. Shipping column entry is defined as a comma-separated list of shipping information in the format `country:state:method_name:cost` like for example: `US:CA:Free shipping:0.00 USD` alternatively the state may be dropped ( if not configured in the zone ) resulting in `US::Free shipping:0.00 USD`. Check the screenshot for more advanced entry.

The basic logic is to take a product and try to figure out the best shipping option available as if the product would be alone inside the cart with the shipping destination set to the one defined by the shipping zone setting ( if multiple zones and multiple locations in the zone are provided we do that for each such location.

Not all of the shipping features present in WooCommerce could be mapped onto Pinterest feed shipping column specification. List of things that are supported:

  - Free shipping without additional settings.
  - Free shipping with a minimum order value. Minimum is tested over a single product. ( still, better than nothing )
  - Flat rate shipping without additional settings.
  - Flat rate with classes: no shipping class and regular classes.
  - Shipping zone locations: single country, country + state, continent ( zip codes are not allowed).
  - Locations mixing: continent + country + state.
  - Taxes for shipping global.
  - Taxes for shipping for a specific country.
  - Filtering of not supported countries.
  - Simple products.
  - Variable products.

Right now no extensibility ( filters ) has been introduced. We can add this when we will see requests for such features.

### Screenshots:
image

1. Connect the plugin to Pinterest.
2. Crate some products.
3. Create some shipping zone.
4. Add shipping methods to the zones
5. Observe the feed file ( `wp-content/uploads/pinterest-for-woocommerce-xxxxxx.xml` ) and check if the `<g:shipping>` tag has an entry that makes sense given the settings that you have created.
6. Additionally add some tax settings that should impact the shipping cost and verify the output again.

### Additional details:

The PR looks big but about one-half is the testing code and not the actual implementation.

Things that we will need to observe is performance. This is a calculation heavy implementation - but given the requirements presented for this task, I have not found any shortcuts that would allow performance improvements without severely impairing the accuracy.

### Changelog entry

> Add - Product shipping information in the feed file.

### DIFF
--- a/src/PinterestShippingZone.php
+++ b/src/PinterestShippingZone.php
@@ -176,6 +176,14 @@ class PinterestShippingZone extends WC_Shipping_Zone {
 	}
 
 
+	/**
+	 * Get shipping methods supported by the implementation.
+	 * Methods are filtered by allowed types and features.
+	 *
+	 * @since x.x.x
+	 *
+	 * @return array Supported shipping methods.
+	 */
 	private function get_supported_shipping_methods() {
 		if ( null !== $this->supported_shipping_methods ) {
 			return $this->supported_shipping_methods;
@@ -189,12 +197,19 @@ class PinterestShippingZone extends WC_Shipping_Zone {
 		return $this->supported_shipping_methods;
 	}
 
+	/**
+	 * Verify if shipping method is supported.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param WC_Shipping_Method $shipping_method Shipping rate to verify.
+	 * @return boolean
+	 */
 	private function is_shipping_method_supported( $shipping_method ) {
 		if ( ! in_array( $shipping_method->id, self::ALLOWED_SHIPPING_METHODS, true ) ) {
 			return false;
 		}
 
-		// We don't support for now free shipping with additional requirements options other than minimum.
 		if ( 'free_shipping' === $shipping_method->id && ! in_array( $shipping_method->requires, self::ALLOWED_FREE_SHIPPING_REQUIRED_SETTINGS, true ) ) {
 			return false;
 		}
--- a/src/PinterestShippingZone.php
+++ b/src/PinterestShippingZone.php
@@ -188,7 +188,8 @@ class PinterestShippingZone extends WC_Shipping_Zone {
 		if ( null !== $this->supported_shipping_methods ) {
 			return $this->supported_shipping_methods;
 		}
-		$active_shipping_methods = $this->get_shipping_methods( true );
+
+		$active_shipping_methods          = $this->get_shipping_methods( true );
 		$this->supported_shipping_methods = array_filter(
 			$active_shipping_methods,
 			array( $this, 'is_shipping_method_supported' )
--- a/src/PinterestShippingZone.php
+++ b/src/PinterestShippingZone.php
@@ -55,7 +55,7 @@ class PinterestShippingZone extends WC_Shipping_Zone {
 	 * @return array            List of filtered countries.
 	 */
 	private function filter_out_not_allowed_countries( $locations ) {
-		$allowed_countries = \Pinterest_For_Woocommerce_Admin::get_ads_supported_countries();
+		$allowed_countries = \Pinterest_For_Woocommerce_Ads_Supported_Countries::get_countries();
 		return array_filter(
 			$locations,
 			function ( $location ) use ( $allowed_countries ) {
--- a/src/PinterestShippingZone.php
+++ b/src/PinterestShippingZone.php
@@ -3,7 +3,9 @@
  * Represents a single Pinterest shipping zone
  *
  * @since   x.x.x
+ * @package Automattic\WooCommerce\Pinterest
  */
+
 namespace Automattic\WooCommerce\Pinterest;
 
 defined( 'ABSPATH' ) || exit;
@@ -13,15 +15,36 @@ use \WC_Shipping_Zone;
 /**
  * WC_Shipping_Zone class.
  */
-class ShippingZone extends WC_Shipping_Zone {
+class PinterestShippingZone extends WC_Shipping_Zone {
 
+
+	/**
+	 * Caching for internal structure of locations.
+	 *
+	 * @var $zone_countries_with_states
+	 */
 	private $zone_countries_with_states = null;
+
+	/**
+	 * Caching for supported shipping methods.
+	 *
+	 * @var $zone_countries_with_states
+	 */
 	private $supported_shipping_methods = null;
 
-	private function filter_out_not_allowed_countries( $location_objects ) {
+
+	/**
+	 * From the list of countries filter out those which are not supported right now.
+	 *
+	 * @since   x.x.x
+	 *
+	 * @param  array $locations List of countries to filter.
+	 * @return array            List of filtered countries.
+	 */
+	private function filter_out_not_allowed_countries( $locations ) {
 		$allowed_countries = \Pinterest_For_Woocommerce_Admin::get_ads_supported_countries();
 		return array_filter(
-			$location_objects,
+			$locations,
 			function ( $location ) use ( $allowed_countries ) {
 				return in_array( $location['country'], $allowed_countries, true );
 			}
--- a/src/PinterestShippingZone.php
+++ b/src/PinterestShippingZone.php
@@ -13,10 +13,14 @@ defined( 'ABSPATH' ) || exit;
 use \WC_Shipping_Zone;
 
 /**
- * WC_Shipping_Zone class.
+ * PinterestShippingZone class.
+ *
+ * By extending WC_Shipping_Zone we are able to add functionality necessary for Pinterest shipping column generation.
+ * This allows us to operate on familiar interface which will be useful when we will continue to expand the functionality.
+ *
+ * @since x.x.x
  */
 class PinterestShippingZone extends WC_Shipping_Zone {
-
 
 	/**
 	 * Caching for internal structure of locations.
@@ -32,6 +36,15 @@ class PinterestShippingZone extends WC_Shipping_Zone {
 	 */
 	private $supported_shipping_methods = null;
 
+	/**
+	 * Types of allowed shipping methods.
+	 */
+	const ALLOWED_SHIPPING_METHODS = array( 'free_shipping', 'flat_rate' );
+
+	/**
+	 * Type of settings in the required field that we support.
+	 */
+	const ALLOWED_FREE_SHIPPING_REQUIRED_SETTINGS = array( '', 'min_amount' );
 
 	/**
 	 * From the list of countries filter out those which are not supported right now.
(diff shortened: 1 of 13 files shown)


jorgemd24:
I was trying to see the changes in `wp-content/uploads/pinterest-for-woocommerce-xxxxxx.xml` but I do not see any update, I think it is because of my account did not pass the Merchant Guidelines.  

Do I need to get the approval first to make this work?

findingsimple:
@jorgemd24 I wonder if you hit https://github.com/woocommerce/pinterest-for-woocommerce/issues/338 ?

Otherwise, we have a slack channel where you can ask the Pinterest team to approval an account for you

jorgemd24:
@budzanowski the XML file should be updated automatically after adding some methods and shipping zones? Or is it updating every X time?

budzanowski:
@jorgemd24 

It is updated once a day. You can change that behavior by altering this line:

https://github.com/woocommerce/pinterest-for-woocommerce/blob/develop/src/ProductSync.php#L30

Just be sure that you have the latest code because I have pushed a fix something like 4h ago.

jorgemd24:
@budzanowski Most of the time (90%) I get the following error when the XML file is being created: 


```
Line 394 "Call to a member function get() on null" "/var/www/html/wp-content/plugins/woocommerce/includes/wc-cart-functions.php"
Line 189 "wc_get_chosen_shipping_method_ids" "/var/www/html/wp-content/plugins/woocommerce/includes/class-wc-customer.php"
Line 461 "get_taxable_address" "/var/www/html/wp-content/plugins/woocommerce/includes/class-wc-tax.php"
Line 565 "get_tax_location" "/var/www/html/wp-content/plugins/woocommerce/includes/class-wc-tax.php"
Line 294 "get_shipping_tax_rates" "/var/www/html/wp-content/plugins/woocommerce/includes/abstracts/abstract-wc-shipping-method.php"
Line 207 "add_rate" "/var/www/html/wp-content/plugins/woocommerce/includes/shipping/flat-rate/class-wc-shipping-flat-rate.php"
Line 233 "calculate_shipping" "/var/www/html/wp-content/plugins/woocommerce/includes/abstracts/abstract-wc-shipping-method.php"
Line 173 "get_rates_for_package" "/var/www/html/wp-content/plugins/pinterest-for-woocommerce/src/Shipping.php"
Line 70  "get_best_shipping_with_cost" "/var/www/html/wp-content/plugins/pinterest-for-woocommerce/src/Shipping.php"
Line 379 "prepare_shipping_column" "/var/www/html/wp-content/plugins/pinterest-for-woocommerce/src/ProductsXmlFeed.php"
```
This is not breaking the code because [we are catching the error and returning an empty string ](https://github.com/woocommerce/pinterest-for-woocommerce/blob/7073529865087bed130c8408122a67a8b5bd4f0b/src/Shipping.php#L84) but it is overriding the XML file and removing the `g:shipping` tag if it was created from before. 

For the other 10%, I can see that the new tag `g:shipping` is added in the XML with the correct shipping methods. 

I narrow the issue down to WC()->session, sometimes the session is not set, causing `$chosen_methods = WC()->session->get( 'chosen_shipping_methods', array() );` to throw an error. Could it be that the scheduled jobs are not loading the session?

budzanowski:
@jorgemd24 Nicely spotted. This was quite strange. UT and my local testing was 100% correct. Only when I have created a fresh site on JN I have start to see the problems. I think that I have fixed this by ensuring that things that are needed are always loaded. Please check again.

src/Shipping.php
```diff
@@ -0,0 +1,253 @@
...
+	 * Local cache of shipping zones defined in WC settings.
+	 *
+	 * @var array|null $shipping_zones
+	 */
+	private static $shipping_zones = null;
+
+	/**
+	 * Prepare content of a shipping column entry for $product.
+	 * Entry is a comma separated string with values in the following format:
+	 *   COUNTRY:STATE:POST_CODE:SHIPPING_COST
```
jorgemd24: ❔ 

Maybe we can mention that this format is use for the XML feed.

src/Shipping.php
```diff
@@ -0,0 +1,253 @@
...
+						// No valid shipping cost for $location. Skip to the next shipping destination.
+						continue;
+					}
+					$shipping_name    = $best_shipping['name'];
+					$shipping_cost    = $best_shipping['cost'];
+					$shipping_country = $location['country'];
+					$shipping_state   = $location['state'];
+
+					// Build shipping entry.
+					$lines[] = "$shipping_country:$shipping_state:$shipping_name:$shipping_cost $currency";
```
jorgemd24: ❓

What do you think if this method returns the shipping data instead of the data formatted?  

Like this, the Shipping Class will only return data about the shipping independently of the format. The formatting could be done in ProductsXMLFeed.

src/PinterestShippingZone.php
```diff
@@ -0,0 +1,261 @@
...
+		$zone_locations = $this->get_zone_locations();
+		$continents     = array_filter( $zone_locations, array( $this, 'location_is_continent' ) );
+		$countries      = array_filter( $zone_locations, array( $this, 'location_is_country' ) );
+		$states         = array_filter( $zone_locations, array( $this, 'location_is_state' ) );
+		$postcodes      = array_filter( $zone_locations, array( $this, 'location_is_postcode' ) );
+		$locations      = array();
+
+		if ( ! empty( $postcodes ) ) {
+			/**
+			 * W don't process zones with postcodes because Pinterest does not support postcode locations.
```
jorgemd24: 💅 

Small typo W don't -> We don't

src/PinterestShippingZone.php
```diff
@@ -0,0 +1,261 @@
...
+		if ( null !== $this->zone_countries_with_states ) {
+			return $this->zone_countries_with_states;
+		}
+
+		$all_continents = WC()->countries->get_continents();
+		$zone_locations = $this->get_zone_locations();
+		$continents     = array_filter( $zone_locations, array( $this, 'location_is_continent' ) );
+		$countries      = array_filter( $zone_locations, array( $this, 'location_is_country' ) );
+		$states         = array_filter( $zone_locations, array( $this, 'location_is_state' ) );
+		$postcodes      = array_filter( $zone_locations, array( $this, 'location_is_postcode' ) );
```
jorgemd24, line 86: ❓ 

Could we not group the zone locations like this: 

```
$type_areas = [];

foreach ( $zone_locations as $zone ) {
      $type_areas[ $zone->type ][] = $zone;
}
```

Instead of iterating the  `$zone_locations` for each type?

src/ProductsXmlFeed.php
```diff
@@ -377,4 +401,24 @@ private static function get_taxonomies( $product_id ) {
...
+	 * @return Shipping
+	 */
+	private static function get_shipping() {
+		if ( null === self::$shipping ) {
+			self::$shipping = new Shipping();
+			/**
+			 * When we start generating lets make sure that the cart is loaded.
+			 * Various shipping and tax functions are using elements of cart.
+			 */
+			wc_load_cart();
```
jorgemd24, line 430: 🤩  

This fixed the issue with the WC session.

src/Shipping.php
```diff
@@ -0,0 +1,253 @@
...
+		return $best_rate;
+	}
+
+	/**
+	 * Pick the best rate from an array of rates.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param  array $rates List of shipping rates.
+	 * @return array        Name and cost for the best found rate or null in case nothing was found.
```
jorgemd24, line 191: ❓ 

In case nothing was found it would return an empty name and a cost of  "INF",  should it not return null?

src/Shipping.php
```diff
@@ -0,0 +1,253 @@
...
+	 * @param WC_Product $product  Product to package.
+	 * @param array      $location Product destination location.
+	 * @return array               Product packed into a package for use by shipping methods.
+	 */
+	public static function put_product_into_a_shipping_package( $product, $location ) {
+		$cart_item = array(
+			'key'          => 0,
+			'product_id'   => $product->get_id(),
+			'variation_id' => null,
+			'variation'    => null,
```
jorgemd24, line 230: ❔ 

Just curious that the variation_id is null but we are still getting the correct shipping if the variation has a shipping class?

jorgemd24 (commented):
I have been doing some testing with different shipping zones such as countries, states, continents and with different taxes and I am getting the expected results. 

The problems with the WC->session has been resolved, I can generate the XML without any issues. 👍 

I left just some comments about the code.

Thanks for your work!

src/PinterestShippingZone.php
budzanowski: We could. Both approaches would work.
I like the direct approach of having a separate variable for each type - this makes the later code easier to write.

What needs to happen in your example is that we need to prepopulate the `$type_areas` array with empty arrays for each type before we append any item. Or detect in the loop if the key exist and add it if not.

For me a flat structure is easier to grasp. Iterating over zone locations for each type is cheap so we don't have to worry about it.

src/Shipping.php
budzanowski: Good idea - implemented.

src/Shipping.php
budzanowski: Corrected.

src/Shipping.php
budzanowski: The shipping classes are only relevant for the Flat Rate Shipping. In that shipping method, the `find_shipping_classes` is used for finding classes. Inside we have the `get_shipping_class` being called on the `$product` that is under the `data` key in the `$cart_item`. The `variation_id` plays no role in this process - it is used elsewhere. So as long as we have the correct item under `data` we are OK.

src/Shipping.php
budzanowski: Not needed here after the refactor.

src/PinterestShippingZone.php
budzanowski: Fixed.

src/ProductsXmlFeed.php
budzanowski: 👍🏻

budzanowski:
@jorgemd24 Thank you for the review. All comments addressed and fix made.

tests/Unit/FeedXMLGenerationShippingTest.php
```diff
@@ -0,0 +1,530 @@
...
+	 * Single shipping zone that has multiple countries and free shipping method.
+	 *
+	 * @group feed
+	 * @group shipping
+	 */
+	public function testZoneWithMultipleCountriesAndFreeShipping(){
+		$zone = ShippingHelpers::createZoneWithLocations(
+			[
+				['US', 'country'],
+				['UK', 'country'],
```
jorgemd24: 💅 

UK country code is incorrect, should be GB. From this test, I would expect free shipping for the 3 countries, US, GB and IT.

jorgemd24 (approved):
LGTM 👍 

Thanks for the changes! All is working as expected.

💅  I have just left a minor comment about a test.

tests/Unit/FeedXMLGenerationShippingTest.php
budzanowski: The code works correctly - the test is misleading though. I have fixed it in https://github.com/woocommerce/pinterest-for-woocommerce/pull/331/commits/abd1bed82cbe6106561ea275dc85869a3b79e12a The `UK` was not recognized by the code and filtred out. This is why in the results we only had `US` and `IT`. After changing `UK` to `GB` we now had to add it to the results because this is a valid country.

Nice catch.

findingsimple:
Thanks @jorgemd24 and @budzanowski !

@budzanowski would you mind creating a pre-release build of this branch for @niketpins and team to test out

Thanks in advance

budzanowski:
Zip for testing.
[pinterest-for-woocommerce.zip](https://github.com/woocommerce/pinterest-for-woocommerce/files/7979570/pinterest-for-woocommerce.zip)

PR #335, opened 2022-01-21
Fix/228 wrong feed registration status multiple feed profiles

### LINKED ISSUE #228: Feed registration status is incorrect when user has more than one feed profile

ksere:

Closes #228 .

**Replaces the already approved #293 as due to some issues in the original repo, the change-set was incorrect**

**The original PR which also shows the original file changes is [here](https://github.com/saucal/pinterest-for-woocommerce/pull/6)**

---

When a merchant has multiple feeds, we should get the feed report, in order to populate the status of the feed, using the stored feed_id. 

### Screenshots:
image

1. Create a feed profile using https://www.pinterest.com/product-catalogs/ directly and then disable it.
2. Install & finish onboarding of the plugin so that a new feed profile is registered by the plugin.
3. Verify that the status shown in the product catalog page is of the expected feed (the one created by the plugin).

### Changelog entry

> Fix - Get feed report for the stored Feed ID instead for the first feed returned by the API.
> Add - get_merchant_feed_report() method that fetches the feed_report
> Tweak - get_merchant_feed() now fetches the feed instead of the feed_report.

### DIFF
--- a/src/API/Base.php
+++ b/src/API/Base.php
@@ -516,7 +516,7 @@ class Base {
 	 * @param string $merchant_id The merchant ID the feed belongs to.
 	 * @param string $feed_id     The ID of the feed.
 	 *
-	 * @return mixed
+	 * @return object
 	 *
 	 * @throws \Exception PHP Exception.
 	 */
--- a/src/API/Base.php
+++ b/src/API/Base.php
@@ -493,48 +493,6 @@ class Base {
 
 
 	/**
-	 * Get a specific merchant feed using the given arguments.
-	 *
-	 * @param string $merchant_id The merchant ID the feed belongs to.
-	 * @param string $feed_id     The ID of the feed.
-	 *
-	 * @return object
-	 *
-	 * @throws \Exception PHP Exception.
-	 */
-	public static function get_merchant_feed( $merchant_id, $feed_id ) {
-		try {
-
-			$feeds = self::get_merchant_feeds( $merchant_id );
-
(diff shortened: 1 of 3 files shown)


budzanowski:
Since it was already approved I am merging it.

PR #337, opened 2022-01-25
Exclude products with zero price on feed generation

### LINKED ISSUE #265: Request to omit zero cost products from Pinterest feed
(https://wordpress.org/support/topic/request-to-omit-zero-cost-products-from-pinterest-feed/)

 > I have about 100 products in my shop that are free downloads, so the product price is $0.00. This is a lead magnet to get my customers to my online shop to buy digital products. On Pinterest side, it always shows an Error 113 for the 101 free products, as Pinterest does not accept 0 values for price.
> 
> Error 113: “Some items have price value formatting errors in product metadata, so those items will not be published”

Add products to WooCommerce with price = 0.00

> If possible, would you make a change to exclude any products with zero cost from the feed in a future release? Many in my industry follow these same practices of offering zero cost items. So, this is more of an enhancement request. Perhaps an option to exclude products in the future using the settings?

alexiglesias31:

Closes #265 .

Exclude $0.00 price products on feed generation.
---
**Pinterest** does not support ingesting $0.00 price products.

1. Add a product with 0.00 price
2. Verify that is not included in the feed

### DIFF
--- a/src/FeedGenerator.php
+++ b/src/FeedGenerator.php
@@ -243,6 +243,20 @@ class FeedGenerator extends AbstractChainedJob {
 
 			$products = wc_get_products( $products_query_args );
 
+			// Exclude products with zero price.
+			$products = array_filter(
+				$products,
+				function ( $product ) {
+					if ( ! $product->get_parent_id() && method_exists( $product, 'get_variation_price' ) ) {
+						$price = $product->get_variation_regular_price();
+					} else {
+						$price = $product->get_regular_price();
+					}
+
+					return ! ( empty( $price ) || empty( floatval( $price ) ) );
+				}
+			);
+
 			$this->prepare_feed_buffers();
 
 			array_walk(


findingsimple:
Thanks @alexiglesias31 can you double-check the coding standards used here please

alexiglesias31:
> Thanks @alexiglesias31 can you double-check the coding standards used here please

Hi @jconroy. The branch was branched off from the refactor branch and will be merged to the refactor branch, as per your suggestion [here](https://github.com/woocommerce/pinterest-for-woocommerce/issues/265#issuecomment-1018435374). The linting problems are beyond of the changes introduced by this PR.

findingsimple:
Aha! Thanks @alexiglesias31

ksere (approved):
The added code looks code. Should do the trick. 

Not sure if it will cause any issues with the iterations of the feed creation job by removing items from the pool at this stage, since it looks like that part isn't finalized yet in the source branch. 

Because of that, this PR cannot be tested at the moment.

ksere:
> Because of that, this PR cannot be tested at the moment.

We can either keep this on hold till the source branch is ready, or redo it based on `develop`.

budzanowski:
It would be cleaner if we would implement it at an individual XML item level.
IMO the `get_xml_item` should return an empty string if the code decides that the $product is not fit for the feed file.

This way we will have the logic deciding if the feed file should be a part as close to XML logic as possible. In the future, we will most probably add other criteria to the logic so it will be nice to have them in a centralized location.

Besides, such logic should be extracted into a separate method. If you look at `ProductsXmlFeed.php` you will see that you already have basically the same code for fetching the price inside `get_property_g_price` which means that you can extract that and reuse it.

This will also allow the addition of UT for no price code much easier.

If you would move your solution using my suggestions to the `add/shipping` branch you will have the scaffolding for tests already there so adding UT should be quite simple. This PR https://github.com/woocommerce/pinterest-for-woocommerce/pull/331 with that branch will be merged after the next Pinterest release (which will happen on 02.02.2022 ) so the code will be available from the develop branch in the next two days.

alexiglesias31:
> It would be cleaner if we would implement it at an individual XML item level. IMO the `get_xml_item` should return an empty string if the code decides that the $product is not fit for the feed file.
> 
> This way we will have the logic deciding if the feed file should be a part as close to XML logic as possible. In the future, we will most probably add other criteria to the logic so it will be nice to have them in a centralized location.

Perfect.
@budzanowski, I have a little question here, when the $product is not fit for the feed file. Placing the logic in `get_xml_item` will not affect the `product_count` variable. Is this the expected behaviour? Let's say, in a 5 products batch (1 product is not fit for the feed), then `product_count` is still 5 right? (Assuming that the logic is to hold in `product_count` the number of analysed products)

budzanowski:
@alexiglesias31 

That is right, it would not be correct. This is why you need to count the number of products that the XML generator finds fit for the file.

For this actually, it would be better if `get_xml_item` would return `null` in case the product is no good. Then you just need to check the output of `get_xml_item` here https://github.com/woocommerce/pinterest-for-woocommerce/pull/337/files#diff-a15b867c7ebc5a8cbe9e72d3200409204b0d65b56a45bb9ba5d5e427bc2d8decR266 and count and append if it is not null.

alexiglesias31:
> If you would move your solution using my suggestions to the `add/shipping` branch you will have the scaffolding for tests already there so adding UT should be quite simple. This PR #331 with that branch will be merged after the next Pinterest release (which will happen on 02.02.2022 ) so the code will be available from the develop branch in the next two days.

@budzanowski, related to move the solution to the `add/shipping` branch, I was following the [suggestion](https://github.com/woocommerce/pinterest-for-woocommerce/issues/265#issuecomment-1018435374) of tackling this issue in this refactor branch.
However, we can move the solution to the `add/shipping` branch. Just to be sure, do you prefer to move it now branching off `add/shipping` and merge it back to `add/shipping`, or in the other hand, do you prefer to put this on hold till you branch be merged to `develop` and branch off this solution from `develop`?

budzanowski:
@alexiglesias31 

It should be OK to move it to `add/shipping`. After I will be merged the Github will recognize that and it will retarget your branch to develop automatically. You can do that manually also. The `add/shipping` was rebased with `develop` so after the merge you should not have any conflicts.

PR #345, opened 2022-01-31
Fix/325 consistent price format

### LINKED ISSUE #325: Ensure consistent price format for different product types.
### Describe the bug:

Right now some of the prices are added to the feed in non-decimal format like 10USD ( simple product ) and 10.00USD ( variable product ). Accodording to the documentation it should always be the second format:

image

We probably just need to ensure to run `wc_format_decimal( $price, wc_get_price_decimals() );` over all of the functions.

#### Expected behavior
All prices are in format according to the documentation.

# Todo

- [ ] Update the code to conform to the specification
- [ ] Update UT ( they should fail after the code change )

alexiglesias31:

Closes #325 .

Add the prices to the feed with the proper decimal format like 10.00USD (second format of the screenshot).
UT was updated to success with the new changes.

### Screenshots:

149070840-9fa2244d-f306-4b30-8626-73ecca462068

1. Create a product with a non-decimal price (like 10USD).
2. Verify that the price in the feed is in decimal format (like 10.00USD).
3. Verify that UT are success with the new changes.

### Changelog entry

> Update - Ensure that the prices are formatted in a consistent way across the feed file.

### DIFF
--- a/src/ProductsXmlFeed.php
+++ b/src/ProductsXmlFeed.php
@@ -418,9 +418,15 @@ class ProductsXmlFeed {
 	 */
 	private static function get_currency_decimals() {
 		$locale_info = include WC()->plugin_path() . '/i18n/locale-info.php';
-		$country     = Pinterest_For_Woocommerce()::get_base_country() ?? 'US';
 
-		return $locale_info[ $country ]['num_decimals'] ?? wc_get_price_decimals();
+		$currencies = get_transient( PINTEREST_FOR_WOOCOMMERCE_PREFIX . '_currencies_list' );
+
+		if ( ! $currencies ) {
+			$currencies = wp_list_pluck( $locale_info, 'num_decimals', 'currency_code' );
+			set_transient( PINTEREST_FOR_WOOCOMMERCE_PREFIX . '_currencies_list', $currencies, DAY_IN_SECONDS );
+		}
+
+		return $currencies[ get_woocommerce_currency() ];
 	}
 
 	/** Fetch shipping object.
--- a/src/ProductsXmlFeed.php
+++ b/src/ProductsXmlFeed.php
@@ -426,7 +426,7 @@ class ProductsXmlFeed {
 			set_transient( PINTEREST_FOR_WOOCOMMERCE_PREFIX . '_currencies_list', $currencies, DAY_IN_SECONDS );
 		}
 
-		return $currencies[ get_woocommerce_currency() ];
+		return $currencies[ get_woocommerce_currency() ] ?? 2;
 	}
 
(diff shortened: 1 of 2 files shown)


budzanowski:
In WooCommerce the number of decimal points is configurable in the settings. Since the Feed has specific requirements the code may be forced to be indifferent to the system settings. The number of decimal points settings is available at WooCommerce > Settings > General ( bottom of the tab )
image 

We should confirm with Pinterest if a variable precision is acceptable. I am mentioning this because we have seen sites that are using for example USD ( so they should use 2 decimal points according to ISO-4271 ) but site owners want to use 0 numbers for precision - for example, because they sell higher-priced products and they want to just show 1000 USD as a price sitewide. If we would be forced to use 2 number precision in such cases then this, in turn, would mean that we can't really rely on the `wc_get_price_decimals` function.

ksere (changes requested):
Great feedback, thanks Bartosz. 

I 'd say since the document is clear enough we go with the default decimal points for the given currency. We should be able to get that from [locale-info.php](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/i18n/locale-info.php).

src/ProductsXmlFeed.php
```diff
@@ -414,7 +414,16 @@ private static function get_taxonomies( $product_id ) {
 	}
 
 	/**
-	 * Fetch shipping object.
+	 * Get locale currency decimals
+	 */
+	private static function get_currency_decimals() {
```
ksere, line 419: This won't produce the desired output if for example my base country is US and my currency is JPY.
We need to get the decimals from that array based on the currency returned by `get_woocommerce_currency()`

ksere (changes requested):

src/ProductsXmlFeed.php
```diff
@@ -414,7 +414,22 @@ private static function get_taxonomies( $product_id ) {
...
+		$locale_info = include WC()->plugin_path() . '/i18n/locale-info.php';
```
ksere: Maybe move the `include` in the conditional as well so that it doesn't load up the array when not needed.

ksere (approved):

budzanowski:
This looks great, approving. Could you please add one additional test where you use different currency - to highlight that the formatting for different decimal points works?

budzanowski (approved):

alexiglesias31:
Done @budzanowski. Added the unit test with JPY (that has 0 decimals as default).

budzanowski:
I have updated the changelog entry in the PR description. Please make sure to have it as it is used by our release tool to automatically populate the changelog.

PR #350, opened 2022-02-01
Omit zero price products

### LINKED ISSUE #265: Request to omit zero cost products from Pinterest feed

alexiglesias31:

Closes #265 .

Exclude $0.00 price products on feed generation.
---
**Pinterest** does not support ingesting $0.00 price products.

1. Add a product with 0.00 price
2. Verify that is not included in the feed

### Changelog entry

> Update - Omit zero price products from the feed file.

### DIFF
--- a/src/ProductSync.php
+++ b/src/ProductSync.php
@@ -41,7 +41,7 @@ class ProductSync {
 	 *
 	 * @var integer
 	 */
-	private static $products_per_step = 500;
+	private static $products_per_step = 50;
 
 	/**
 	 * The number of products to hold in a buffer variable before writing their XML output to a file.
@@ -49,7 +49,7 @@ class ProductSync {
 	 *
 	 * @var integer
 	 */
-	private static $products_per_write = 100;
+	private static $products_per_write = 25;
 
 	/**
 	 * The current index of the loop iterating through feed creation steps.
@@ -57,6 +57,14 @@ class ProductSync {
 	 * @var integer
 	 */
 	private static $current_index = 0;
+
+	/**
+	 * The current number of products already added to the feed.
+	 * Using current_index to track this may lead to errors if some products not fit on the feed.
+	 *
+	 * @var integer
+	 */
+	private static $current_products = 0;
 
 	/**
 	 * The buffer used to hold XML markup to be printed to the XML file.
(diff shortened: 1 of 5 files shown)


src/ProductsXmlFeed.php
```diff
@@ -73,12 +73,16 @@ public static function get_xml_footer() {
 	/**
 	 * Returns the Item's XML for the given product.
 	 *
-	 * @param WC_Product $product The product to print the XML for.
+	 * @param WC_Product $product  The product to print the XML for.
 	 *
-	 * @return string
+	 * @return string|null
```
ksere: Prefer keeping 1 single return type. 
And check for empty string where needed

ksere (changes requested):
Small comment on a method's return type

src/ProductSync.php
```diff
@@ -362,15 +372,23 @@ public static function handle_feed_generation() {
...
 
 			for ( self::$current_index; ( self::$current_index < $products_count ); self::$current_index++ ) {
 
 				$product_id = $product_ids[ self::$current_index ];
 
-				self::$iteration_buffer .= ProductsXmlFeed::get_xml_item( wc_get_product( $product_id ) );
-				self::$iteration_buffer_size++;
+				$xml_item = ProductsXmlFeed::get_xml_item( wc_get_product( $product_id ) );
+
+				if ( $xml_item ) {
```
ksere:
```suggestion
				if ( '' !== $xml_item ) {
```

ksere (approved):

budzanowski:
Rebased with `develop` to fix merge conflicts.

budzanowski (approved):

PR #351, opened 2022-02-02
Fix/323 strip html shortcodes from feed

### LINKED ISSUE #323: ProductsXmlFeed::description does not strip the HTML tags.
### Describe the bug:
According to the documentation, the description filed does not support the HTML tags and sets no limit on characters and the length of the description.

image
from https://help.pinterest.com/en/business/article/before-you-get-started-with-catalogs

The function that is creating the feed file XML should be aware of these limitations. Right now it is not doing any formatting/stripping. 

#### Expected behavior
`ProductsXmlFeed::description` should create XML that conforms the requirements in the documentation.

Requirements to finish this item:

- [ ] Modify the method in such a way that it produces output according to the requirements in the documentation.
- [ ] Add unit tests that test the new functionality.

### LINKED ISSUE #344: Check shortcode handling and update if necessary
Merchant reported in https://github.com/woocommerce/pinterest-for-woocommerce/issues/247 that html and shortcodes were NOT being stripped.

Original functionality https://github.com/woocommerce/pinterest-for-woocommerce/pull/85 indicates that shortcodes and HTML should be being stripped.

With the refactoring taking place a subsequent issue was created to add support for stripping HTML https://github.com/woocommerce/pinterest-for-woocommerce/issues/323

So we also need to double check shortcode handling in the extension. 

See https://github.com/woocommerce/google-listings-and-ads/pull/911 for some inspiration for prior art https://github.com/woocommerce/google-listings-and-ads/pull/911 in the GLA extension. Ideally we want to strip shortcodes by default - but provide a filter for merchants to use to override if they choose to allow it.

* [ ] Confirm if shortcodes are being stripped or not (i.e. need investigation)
* [ ] Strip shortcodes by default
* [ ] Add filter to allow merchants to enable shortcodes to be applied if they wish
* [ ] Add unit tests

alexiglesias31:

Closes #323 
Closes #344 

The PR aims to strip the HTML tags and shortcodes from feed file.

- ProductsXmlFeed::description creates an XML that conforms the requirements in the documentation.
- The shortcodes are stripped from description field by default.
- Added a filter to allow merchants to enable shortcodes to be applied if they wish.
- Added unit tests.

### Screenshots:

image

1. Create a simple product.
2. Add HTML tags on the title.
3. Add a shortcode and/or HTML tags on the product's description.
4. Verify that HTML tags were stripped in title and description fields.
5. Verify that shortcodes are stripped by default from the description field.
6. Shortcode in product's description should be allowed if the filter `pinterest_for_woocommerce_product_description_apply_shortcodes` returns `true`.

### DIFF
--- a/src/ProductsXmlFeed.php
+++ b/src/ProductsXmlFeed.php
@@ -11,6 +11,7 @@ namespace Automattic\WooCommerce\Pinterest;
 use Automattic\WooCommerce\Pinterest\Product\Attributes\AttributeManager;
 use WC_Product_Variation;
 use WC_Product;
+use Automattic\WooCommerce\Pinterest\Logger;
 
 if ( ! defined( 'ABSPATH' ) ) {
 	exit;
@@ -42,6 +43,14 @@ class ProductsXmlFeed {
 		'g:shipping',
 		'g:additional_image_link',
 	);
+
+
+	/**
+	 * Limit of characters allowed by Pinterest in the product description.
+	 *
+	 * @var int
+	 */
+	private static $limit_chars_product_description = 10000;
 
 
 	/**
@@ -204,7 +213,12 @@ class ProductsXmlFeed {
 		$description = str_replace( '[&hellip;]', '...', $description );
 
 		// Limit the number of characters in the description to 10000.
-		return '<' . $property . '><![CDATA[' . substr( $description, 0, 10000 ) . ']]></' . $property . '>';
+		if ( strlen( $description ) > 10000 ) {
+			Logger::log( sprintf( esc_html__( 'The product [%s] has a description longer than the allowed limit.', 'pinterest-for-woocommerce' ), $product->get_id() ) );
+		}
+		$description = substr( $description, 0, self::$limit_chars_product_description );
+
+		return '<' . $property . '><![CDATA[' . $description . ']]></' . $property . '>';
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


findingsimple:
Thanks for adding unit tests - keep it up!

ksere (approved):

ksere (approved):

src/ProductsXmlFeed.php
```diff
@@ -182,7 +182,29 @@ private static function get_property_description( $product, $property ) {
...
+		}
+
+		// Strip HTML tags from description.
+		$description = wp_strip_all_tags( $description );
+
+		// Strip [&hellip] character from description.
+		$description = str_replace( '[&hellip;]', '...', $description );
+
+		// Limit the number of characters in the description to 10000.
+		return '<' . $property . '><![CDATA[' . substr( $description, 0, 10000 ) . ']]></' . $property . '>';
```
budzanowski: We should log if we have encountered descriptions longer than 10K chars.

budzanowski: Please turn 10000 into class const and let it have a descriptive name.

budzanowski:
I have left two very small suggestions, besides that this looks very nice :)

alexiglesias31:
Done @budzanowski.

budzanowski:
I have moved the limit value to a const. And I have added a UT to test the description length clipping functionality.

budzanowski (approved):

PR #349, opened 2022-02-01
Use Pinterest dasicon for the Pinterest Attribues tab.

budzanowski:
### Changes proposed in this Pull Request:

Use Pinterest Dashicon for product attributes tab.

### Screenshots:

From this:
image

to this:
image

### Detailed test instructions:

0. Build the assets
1. Open then wp-admin
2. Open products.
3. Got to a product.
4. Observe the icon :)

PR #352, opened 2022-02-03
Remove unused css copy from GLA.

budzanowski:
Per this commit note from a different pull request https://github.com/woocommerce/pinterest-for-woocommerce/pull/349#pullrequestreview-870449178

This change has no effect. Removed CSS was not used.


mikkamp (approved):
Looks good, thanks for fixing this.

PR #343, opened 2022-01-26
Fetch and store merchant id during setup.

budzanowski:
### Changes proposed in this Pull Request:

Closes #342  .

This adjusts the logic introduced in https://github.com/woocommerce/pinterest-for-woocommerce/pull/335 and in https://github.com/woocommerce/pinterest-for-woocommerce/pull/305
Before this change the code was setting up the feed but was failing to store the merchant id. This resulted in the scenario where the code did not recognize that the feed and merchant were already set up and was redoing the operation. 

### Detailed test instructions:

Please check #342 for testing and reproduction instructions.

### Additional details:

I don't really understand why this scenario is different compared to the scenario without a linked business account - If someone knows please share, if not then I will need to dig deeper.

### Changelog entry

> Fix - Store merchant id during the account creatioin.

Issue #342, opened 2022-01-26
Pinterest Personal account linked to Business account, feed is not synched correctly


ChristianAF:
### Describe the bug:
When merchant has a personal account linked to a Business account, Feed creation and sync with Pinterest is not working ok.
On Pinterest End the feed is being created several times with same source

https://user-images.githubusercontent.com/43881685/151174629-fdacd648-ace8-4fbe-abb8-a6a28f9ceace.mp4

On the site:
![image](https://user-images.githubusercontent.com/43881685/151174774-f56c0d02-23ab-4ab8-b341-8886c028c975.png)

Logs:
```
2022-01-26T13:42:19+00:00 DEBUG GET Request: https://api.pinterest.com/v3/commerce/product_pin_merchants//

--- EMPTY STRING ---

2022-01-26T13:42:19+00:00 DEBUG Response: 

HTTP/1.1 405 Method Not Allowed
Connection: close
Content-Length: 164
content-type: application/json
pinterest-version: 9d2c094
x-content-type-options: nosniff
x-frame-options: DENY
x-envoy-upstream-service-time: 8
x-pinterest-direct: true
cache-control: no-cache, no-store, must-revalidate
age: 0
set-cookie: _ir=0; Max-Age=1800; HttpOnly; Path=/; Secure
x-pinterest-rid: 7946711880453185
Date: Wed, 26 Jan 2022 13:42:19 GMT
X-CDN: fastly
Pinterest-Generated-By: coreapp-ngapi-prod-0a018829

{"code":5,"data":"405 Method Not Allowed: The method is not allowed for the requested URL.","message":"Niedozwolona metoda","endpoint_name":null,"status":"failure"}

2022-01-26T13:42:19+00:00 ERROR Niedozwolona metoda
```

### Steps to reproduce:
1. on new site install plugin 1.0.3
2. Do the onboarding with a PErsonal account linked with a Business account

### Expected behavior:
Feed is created and synched correctly

### Actual behavior:
feed is synched several times with Pinterest business account, and on the admin site is with errors

PR #353, opened 2022-02-03
Do not create merchant on get_feed_state

### LINKED ISSUE #341: Create merchant endpoint shouldnt be called repeatedly if there is an error while creating merchants
### Describe the bug:
Recently on pinterest we noticed that as the error rate increased in create merchant (due to rate limiting, which was fixed by increasing the rate limit), RPS also increased. See graph below

From initial research the root cause is mostly due to [healthcheck](https://github.com/woocommerce/pinterest-for-woocommerce/blob/296316e9be7aabe1afdfd5110d19c072c1003495/src/API/HealthCheck.php#L53) or [product sync](https://github.com/woocommerce/pinterest-for-woocommerce/blob/ff9d7ac7f9eab7d45e234ac920eb0dd45fb0cb0b/src/ProductSync.php#L514)

### Steps to reproduce:
1. Create a new shop and add it to a new pinterest account
2. Make sure merchant creation fails on pinterest side
3. If you have multiple such failed merchant creations, then you will notice that create merchant end point will get called repeatedly

### Expected behavior:
When there is an error in API call, e.g in merchant creation it is better to exponentially back off rather than repeatedly try it

### Actual behavior:
If there is an error in merchant creation, the create merchant is called repeatedly during health checks and product sync

### Additional details:
Screen Shot 2022-01-19 at 3 15 11 PM

alexiglesias31:

Fix the repeatedly calls to the create merchant endpoint.

1. Create a new shop and make the onboarding process.
2. Verify on the logs that `catalogs/partner/connect/` endpoint is hit only once.

### DIFF
--- a/src/API/FeedState.php
+++ b/src/API/FeedState.php
@@ -246,25 +246,17 @@ class FeedState extends VendorAPI {
 	 */
 	private function add_feed_registration_state( $result ) {
 
-		$merchant    = null;
 		$merchant_id = Pinterest_For_Woocommerce()::get_data( 'merchant_id' );
 		$feed_id     = Pinterest_For_Woocommerce()::get_data( 'feed_registered' );
 		$extra_info  = '';
 
 		try {
 
-			if ( empty( $merchant_id ) ) {
-				$merchant    = Pinterest\Merchants::get_merchant();
-				$merchant_id = Pinterest_For_Woocommerce()::get_data( 'merchant_id' );
-			}
-
 			if ( empty( $merchant_id ) || empty( $feed_id ) ) {
 				throw new \Exception( esc_html__( 'Product feed not yet configured on Pinterest.', 'pinterest-for-woocommerce' ), 200 );
 			}
 
-			if ( empty( $merchant ) ) {
-				$merchant = Base::get_merchant( $merchant_id );
-			}
+			$merchant = Base::get_merchant( $merchant_id );
 
 			if ( 'success' !== $merchant['status'] ) {
 				throw new \Exception( esc_html__( 'Could not get merchant info.', 'pinterest-for-woocommerce' ) );


findingsimple:
Let us know once the PR description is cleaned up and then request a review from @woocommerce/ventures

ksere (approved):

ksere:
> Let us know once the PR description is cleaned up and then request a review from @woocommerce/ventures

@jconroy I am not able to assign to that team.

findingsimple:
Oh interesting and sorry @ksere I thought it was available to anyone, I've just requested

budzanowski:
Are we sure that this will not reintroduce https://github.com/woocommerce/pinterest-for-woocommerce/issues/342?

alexiglesias31:
We tested and the changes don't re-introduce the [#342](https://github.com/woocommerce/pinterest-for-woocommerce/issues/342) issue.
The [#342](https://github.com/woocommerce/pinterest-for-woocommerce/issues/342) issue was originated because the merchant_id was not being properly saved, (which was corrected [here](https://github.com/woocommerce/pinterest-for-woocommerce/pull/343/files#diff-165dc486d272cdd52e9f8f67ea2fd88e3dacf5e7600d6e2e398303fb46cf04d0R74)), therefore the request to the API didn't contain the `merchant_id` as can be observed in the Logs on the [#342](https://github.com/woocommerce/pinterest-for-woocommerce/issues/342) issue:
https://api.pinterest.com/v3/commerce/product_pin_merchants//
In the other hand, calling twice the method `Merchants::get_merchant()` on every request (`FeedState` and `HealthCheck`) can hammer the API as spotted in the [#341](https://github.com/woocommerce/pinterest-for-woocommerce/issues/341) issue. So, we are reverting the changes of trying to create the merchant again on `FeedState` requests, since that will be handled by the `HealthCheck` script.

mikkamp:
Thanks for the additional clarification on what this PR does. However I'm still missing how it will fully resolve the scenario described in #341

The expected behaviour is:
> When there is an error in API call, e.g in merchant creation it is better to exponentially back off rather than repeatedly try it

And in the issue it describes the creation mainly being triggered from HealthCheck and ProductSync. From what I can see HealthCheck is triggered on every Pinterest page view and the ProductSync is triggered every 10 minutes. While that is not exactly hammering the API, if there is a persistent error it will still try to create the merchant every 10 minutes without backing off.

I'm not sure how I can best reproduce the following point:
> Make sure merchant creation fails on pinterest side

However I'd expect to be seeing some kind of failure count being recorded and it either backing off or stopping if it's unable to create the merchant. Removing the creation being triggered from the FeedState request might be helpful, but from my testing that request was called the least.

ksere:
Hey @mikkamp 

> And in the issue it describes the creation mainly being triggered from HealthCheck and ProductSync. From what I can see HealthCheck is triggered on every Pinterest page view and the ProductSync is triggered every 10 minutes. While that is not exactly hammering the API, if there is a persistent error it will still try to create the merchant every 10 minutes without backing off.

Indeed HealthCheck was calling that on every pageview of the plugins, settings, and each time the user switched between the settings tabs. Which I assume for a user who sees "Merchant / Feed will be configured shortly" type of message we are throwing at that point it can trigger a response of clicking refresh or switching between tabs till that process is done. 

So attempting to create the merchant in that case can probably result in dozens of requests per minute. Leaving that only to the scheduled task (once every 10mins) should result in much fewer attempts. AFAIK We don't have exact info on what "hammering the API" meant but it was probably in the context of multiple users trying to set-up the plugin while there was another issue at play (related to Pinterest's API) causing merchant creation failures. 

>However I'd expect to be seeing some kind of failure count being recorded and it either backing off or stopping if it's unable to create the merchant. Removing the creation being triggered from the FeedState request might be helpful, but from my testing that request was called the least.

Imho this is too complex for the issue its trying to solve, given that probably very few users get merchant creation errors after those API issues were resolved. Changes in this PR should be sufficient. 
If you still think we should implement this, let us know and we will re-scope the task :)

mikkamp (approved):
Thanks for the additional clarification. I'd agree that it looks like hammering the API was by many users and not an individual site. I still think there is value in leaving the original issue #341 open as this PR doesn't fully resolve that scenario, but I've removed the bug label and added it as technical debt since it's not a priority.

I think this PR still adds value, so it looks ready to merge. If we can just update the PR description so it doesn't auto close #341.

budzanowski:
I am letting myself update the description to remove the #341 reference. This partially fixes the problem but as @mikkamp has noted it does not fully resolve the issue.

PR #355, opened 2022-02-08
Force logs when not connected.

### LINKED ISSUE #116: Enable debug logging based on WP_DEBUG

As by default debug logging is [disabled](https://github.com/saucal/pinterest-for-woocommerce/blob/3598d1886e0baa6583b4c2d2b00fc4978727d375/class-pinterest-for-woocommerce.php#L64) any potential issues during the onboarding - for example on API requests to Pinterest - are not caught.
The user needs to complete onboarding and enable debug logging in the settings, which may not be possible if there is an error during onboarding. 

1. Start the onboarding process.
2. Connect your account.
3. Revoke auth of "App" through your Pinterest dashboard or using any other way make the API calls fail.

This could be solved by always logging if WP_DEBUG is true, as if the `enable_debug_logging` was enabled. [Log method](https://github.com/saucal/pinterest-for-woocommerce/blob/feature/refactor-plugin-navigation/src/Logger.php#L46)

We could also 
- just have `enable_debug_logging` enabled by default, but it will result in the whole userbase logging API calls for no reason. 
- Enable logging regardless of the plugin setting when not yet fully onboarded.

budzanowski:

By default the logs are disabled. This means that when there are issues during the connection process we lose the information about the errors. With this PR the connection logs will be preserved.

Closes #116  .

Follow the steps from #116

### Changelog entry

> Add - Enable debug logs for the connection process.

### DIFF
--- a/src/Logger.php
+++ b/src/Logger.php
@@ -48,6 +48,14 @@ class Logger {
 			$allow_logging = Pinterest_For_WooCommerce()::get_setting( 'enable_debug_logging' );
 		}
 
+		/*
+		 * When the integration is not connected force the logs.
+		 * This will allow to debug potential connection issues.
+		 */
+		if ( ! Pinterest_For_Woocommerce()::is_setup_complete() ) {
+			$allow_logging = true;
+		}
+
 		if ( empty( $allow_logging ) || ! function_exists( 'wc_get_logger' ) ) {
 			return;
 		}


findingsimple (approved):
LGTM

Note the changelog entry in the PR description looks like it has a stray "oh" on the end

budzanowski:
> Note the changelog entry in the PR description looks like it has a stray "oh" on the end

image

PR #357, opened 2022-02-08
Avoid to constantly create transient options

### LINKED ISSUE #336: Constantly generation transients on wp-options
### Describe the bug:
When tracking conversion is activated, transients are constantly being generated on wp_options. it's a problem for site with high traffic.

### Steps to reproduce:
1. Install Pinterest for WooCommerce
2. Navigate on the site 
3. go to the table wp_options

### Expected behavior:
no transients created

### Actual behavior:
transients are created on wp_options

### Additional details:
https://saucal.zendesk.com/agent/tickets/8792

### LINKED ISSUE #359: [p3] Sometimes two page view events are fired for tag
### Describe the bug:

One one page view event should be fired per page view. Sometimes there are two generated. However its not consistently reproducable

### Steps to reproduce:
1. e.g Store where the issue was seen https://due-squid.jurassic.ninja/shop/
2. Install pinterest tag plugin for help with testing the tag https://chrome.google.com/webstore/detail/pinterest-tag-helper/gmlcbajhgoaaegmlbaclmmmhpmfdajmp?hl=en
3.  Click add to cart in the website. (Try other actions too since this bug was only happening intermittently)

### Expected behavior:
One page view event per page view

### Actual behavior:
Double page view events

### Additional details:
1-BKN-M3YQD6Ugs8b9ohqThlJUf2OWNb-H1Gm5lYm

alexiglesias31:

Closes #336 .
Closes #359 .

The plugin is creating some uncaught transient events (for the event `pagevisit`).

---
The name of the transients are based on the customer_id of the user, when a user that is not logged nor registered visits a page a transient is generated and since the customer_id is not set, the transient can't be caught on `shutdown` to print it and remove it.
We need to change the logic that captures the events in order to print all events in the same runtime and avoid storing transients for anonymous users. We are only storing the transient for `add_to_cart` events when redirect to cart is enabled.

1. Install Pinterest for WooCommerce.
2. Navigate on the site.
3. Transient options for pagevisit event should not remain on wp_options table.

### DIFF
--- a/class-pinterest-for-woocommerce.php
+++ b/class-pinterest-for-woocommerce.php
@@ -247,6 +247,9 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 
 			// Disconnect advertiser if advertiser or tag change.
 			add_action( 'update_option_pinterest_for_woocommerce', array( $this, 'maybe_disconnect_advertiser' ), 10, 2 );
+
+			// Enable customer session if tracking is enabled.
+			add_action( 'woocommerce_init', array( $this, 'enable_woocommerce_session' ) );
 		}
 
 
@@ -276,6 +279,21 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 			$admin->register();
 			$attributes_tab->register();
 			$variation_attributes->register();
+		}
+
+		/**
+		 * Enable WC_Sessions on woocommerce_init to track events for unauthorized users.
+		 *
+		 * @return void
+		 */
+		public function enable_woocommerce_session() {
+			if ( is_user_logged_in() || is_admin() || ! self::is_tracking_configured() ) {
+				return;
+			}
+
+			if ( isset( WC()->session ) && ! WC()->session->has_session() ) {
+				WC()->session->set_customer_session_cookie( true );
+			}
 		}
 
 		/**
--- a/class-pinterest-for-woocommerce.php
+++ b/class-pinterest-for-woocommerce.php
@@ -248,8 +248,6 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 			// Disconnect advertiser if advertiser or tag change.
 			add_action( 'update_option_pinterest_for_woocommerce', array( $this, 'maybe_disconnect_advertiser' ), 10, 2 );
 
-			// Enable customer session if tracking is enabled.
-			add_action( 'woocommerce_init', array( $this, 'enable_woocommerce_session' ) );
 		}
 
 
@@ -279,21 +277,6 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 			$admin->register();
 			$attributes_tab->register();
 			$variation_attributes->register();
-		}
-
-		/**
-		 * Enable WC_Sessions on woocommerce_init to track events for unauthorized users.
-		 *
-		 * @return void
-		 */
-		public function enable_woocommerce_session() {
-			if ( is_user_logged_in() || is_admin() || ! self::is_tracking_configured() ) {
-				return;
-			}
-
-			if ( isset( WC()->session ) && ! WC()->session->has_session() ) {
-				WC()->session->set_customer_session_cookie( true );
-			}
 		}
 
 		/**
--- a/src/Tracking.php
+++ b/src/Tracking.php
@@ -93,7 +93,7 @@ class Tracking {
 			}
 
 			// AddToCart - non-ajax.
-			add_action( 'woocommerce_add_to_cart', array( __CLASS__, 'hook_add_to_cart_event' ), 10, 4 );
+			add_action( 'woocommerce_add_to_cart', array( __CLASS__, 'hook_add_to_cart_event' ), 20, 4 );
 
 			// Checkout.
 			add_action( 'woocommerce_before_thankyou', array( __CLASS__, 'hook_checkout_event' ), 10, 1 );
--- a/src/Tracking.php
+++ b/src/Tracking.php
@@ -105,7 +105,7 @@ class Tracking {
 		add_action( 'shutdown', array( __CLASS__, 'save_async_events' ) );
 
 		// Print to head.
-		add_action( 'wp_head', array( __CLASS__, 'print_script' ), 10 );
+		add_action( 'wp_head', array( __CLASS__, 'print_script' ) );
 
 		add_action( 'admin_init', array( __CLASS__, 'verify_advertiser_connection' ) );
 	}
--- a/src/Tracking.php
+++ b/src/Tracking.php
@@ -84,7 +84,7 @@ class Tracking {
 		if ( function_exists( 'WC' ) ) {
 
 			if ( ! wp_doing_ajax() ) {
-				add_action( 'wp', array( __CLASS__, 'late_events_handling' ) );
+				add_action( 'wp_head', array( __CLASS__, 'late_events_handling' ), 20 );
 			}
 
 			// AddToCart - ajax.
@@ -105,7 +105,7 @@ class Tracking {
 		add_action( 'shutdown', array( __CLASS__, 'save_async_events' ) );
 
 		// Print to head.
-		add_action( 'wp_head', array( __CLASS__, 'print_script' ) );
+		add_action( 'wp_head', array( __CLASS__, 'print_script' ), 10 );
 
 		add_action( 'admin_init', array( __CLASS__, 'verify_advertiser_connection' ) );
 	}


ksere (approved):

ksere (approved):

findingsimple:
Checking if this is still draft?

alexiglesias31:
> Checking if this is still draft?

@jconroy, the PR is ready for review now.

src/Tracking.php
```diff
@@ -84,7 +84,7 @@ public static function maybe_init() {
 		if ( function_exists( 'WC' ) ) {
 
 			if ( ! wp_doing_ajax() ) {
-				add_action( 'wp', array( __CLASS__, 'late_events_handling' ) );
+				add_action( 'wp_head', array( __CLASS__, 'late_events_handling' ), 20 );
```
mikkamp, line 87: This function tracks the page visit and the product category visit event, both of which use `add_event`. Since we change the hook `did_action( 'wp_head' )` is always going to be truthy. Which means we are always going to delay the tracking of the page visit till the next request from the same user (transient is tied to customer cookie).

That looks like we'd be making the tracking a lot more unreliable since we can't guarantee that there will be a next request from the same user.

class-pinterest-for-woocommerce.php
```diff
@@ -278,6 +281,21 @@ public function admin_init() {
...
+		 *
+		 * @return void
+		 */
+		public function enable_woocommerce_session() {
+			if ( is_user_logged_in() || is_admin() || ! self::is_tracking_configured() ) {
+				return;
+			}
+
+			if ( isset( WC()->session ) && ! WC()->session->has_session() ) {
+				WC()->session->set_customer_session_cookie( true );
```
mikkamp: WooCommerce explicitly sets the customer cookie only when a cart event happens. Before that the customer is considered to be anonymous without any cookies. We are changing that behaviour here and attaching a customer cookie to everyone. I'm not convinced we should be changing that here.

mikkamp (commented):
I haven't followed the full history of how the tracking code was implemented, so maybe I'm missing something.

From my understanding this tracking code works on the premise that some events are caught very late (after wp_head), when this happens we save them in a transient (attached to the customer session). On the subsequent request from the same user it loads any saved events and calls the tracking code for them. Can you clarify if that's correct?

This logic seems flawed as we can't guarantee there is going to be a subsequent request from the same user. Are there some specific scenarios where a page visit can't be reliably tracked, and we specifically need to save it to be tracked later? There is still the `wp_footer` hook, which could provide an opportunity to print a tracking script.

I did some further digging and it seems it was PR #67 where the tracking was delayed for redirect after add to cart. In that scenario we would have a session cookie already set by WooCommerce. So I think we should be stricter and only allow saving delayed tracking for that specific scenario.

I can reproduce a redirect scenario where the transient gets saved but never cleaned up, by visiting the checkout page (with an empty cart in an incognito window). This redirects to the cart page.

Also an unrelated thing I noticed is the `verify_advertiser_connection` function. It's hooked into `admin_init` but the `maybe_init` returns early when `is_admin` is true, so it looks like that code will never get called.

src/Tracking.php
ksere: > Since we change the hook did_action( 'wp_head' ) is always going to be truthy

Correct.

> Which means we are always going to delay the tracking of the page visit till the next request from the same user
Not true. I will try to shed some light here:
```
$action = did_action( 'wp_head' ) ? 'print_event' : 'enqueue_event';
```
In `add_event()` the action would be `print_event` which will result in those events being printed immediately. Since we (now) are already in `wp_head` for these events, they will be printed in the `<head>`. 

As implemented, only the events that need to be printed on the next request are stored to the transient. 

The `add_event()` method will either have a `print_event` action which will print the event code on the spot, or have a `enqueue_event` action which will put it in a runtime variable. Events will get stored in that transient if they haven't been printed (so that runtime variable isn't flushed) when reaching `shutdown`. 
So a call to `add_event()` that runs before `wp_head` will enqueue an event, but when `wp_head` does run, it should print the event and clear the queue. 
The only case where an event should end up in the transient is when `add_event()` is called before `wp_head` and `wp_head` is never fired, as in the "Add2Cart with redirect to cart enabled" flow.

mikkamp: Thanks for the clarification, I misinterpreted that logic. Because that means with it previously hooking on `wp` it should have always been saving the transient (as long as there was a session). I wasn't seeing that behaviour in my tests, but I suppose that was because `$deferred_conversion_events_transient_key` was still null.

It still leaves me with some questions though.

For an add to cart event we hook into `woocommerce_add_to_cart`, which will always be called before `wp_head`. How are we recording that if we don't have redirect to cart enabled? Does that mean we are still scheduling it for the next request? For example on a single product page when redirect is not enabled (or on a shop page if ajax and redirect are disabled). From my testing it is still being scheduled which makes it seem like the `did_action( 'wp_head' )` is not the right way to detect if we should enqueue or not.

Also why do we need to be setting the WC session to "track events for unauthorized users"? If the only case we need it is when redirect to cart is enabled, then can't we just set `$deferred_conversion_events_transient_key` at a later point (WC will have set the session before `woocommerce_add_to_cart` is called)? There is a reason why WooCommerce doesn't always set the tracking cookies, and I think we should respect that, and not change this behaviour.

alexiglesias31: > Also why do we need to be setting the WC session to "track events for unauthorized users"? If the only case we need it is when redirect to cart is enabled, then can't we just set $deferred_conversion_events_transient_key at a later point (WC will have set the session before woocommerce_add_to_cart is called)? There is a reason why WooCommerce doesn't always set the tracking cookies, and I think we should respect that, and not change this behaviour.

alexiglesias31: > For an add to cart event we hook into `woocommerce_add_to_cart`, which will always be called before `wp_head`. How are we recording that if we don't have redirect to cart enabled? Does that mean we are still scheduling it for the next request? For example on a single product page when redirect is not enabled (or on a shop page if ajax and redirect are disabled). From my testing it is still being scheduled which makes it seem like the `did_action( 'wp_head' )` is not the right way to detect if we should enqueue or not.

The `woocommerce_add_to_cart` [enqueues the event](https://github.com/woocommerce/pinterest-for-woocommerce/blob/7d7f82c8bbf27c259bcf6f47e41d0443c80bf866/src/Tracking.php#L213), and since it is going to be called before `wp_head`, the enqueued event is always printed when `wp_head` is called.
If ajax is enabled and redirect is disabled we are [using JS to track an async ajax event](https://github.com/woocommerce/pinterest-for-woocommerce/blob/7d7f82c8bbf27c259bcf6f47e41d0443c80bf866/src/Tracking.php#L92).
If ajax and redirect are disabled, then the event is enqueued and printed in the same runtime (on `wp_head`).


> Also why do we need to be setting the WC session to "track events for unauthorized users"? If the only case we need it is when redirect to cart is enabled, then can't we just set `$deferred_conversion_events_transient_key` at a later point (WC will have set the session before `woocommerce_add_to_cart` is called)? There is a reason why WooCommerce doesn't always set the tracking cookies, and I think we should respect that, and not change this behaviour.

Correct. 
We included a change to set the `$deferred_conversion_events_transient_key` at a later point, since the only case is when redirect to cart is enabled.

mikkamp: Thanks for the additional changes and clarifications. I still found one scenario where it doesn't look like it's handling it correctly. Can you clarify what the expected behaviour should be there?

##### Single product page add to cart (without redirect)
Regardless of the ajax setting, on the single product page it adds to the cart using a POST request to the current page. If I set a breakpoint in the function `hook_add_to_cart_event` and then add a product to the cart, I get the following values:
image

Since `wp_head` hasn't run yet it will queue the event and I'm seeing the transient being saved. So the event won't get tracked till I view another page.

Same scenario occurs when I disable ajax add to cart and try to add to cart from an archive/shop page. The event is also queued.

alexiglesias31: > Single product page add to cart (without redirect)
Regardless of the ajax setting, on the single product page it adds to the cart using a POST request to the current page. If I set a breakpoint in the function hook_add_to_cart_event and then add a product to the cart, I get the following values:

Correct.
In this case the transient is saved in the function `hook_add_to_cart_event`. 

> So the event won't get tracked till I view another page.

When the product is added to the cart the page is reloaded and the event saved in the transient is printed and cleaned on `wp_head`. So, the event is tracked in the same runtime.

> Same scenario occurs when I disable ajax add to cart and try to add to cart from an archive/shop page. The event is also queued.

Same happens here. The event is captured in `hook_add_to_cart_event` and printed and cleaned on `wp_head` in the same runtime.

mikkamp (approved):
Thanks for the all the clarifications on the various tracking scenarios. With the adjusted priorities I'm now seeing the hooks and the events firing at the expected time.

In the cases where redirect to cart is not enabled it enqueue's the event early, but it's printed by `print_script` so when we reach shutdown, all events have been printed.

PR #358, opened 2022-02-09
Use website verification instead of domains

### LINKED ISSUE #354: Few api endpoint /v3/catalog/partner/connect/ requests are failing
### Describe the bug:
Reason: Pinterest allow users to claim domain with paths (e.g. https://xxz.com/en-US/ might be claimed by one user and  https://xxz.com/en-en_GB/ claimed by another user)
When merchant domains are passed with path one verification step is failing 

### Steps to reproduce:

1. invoke /v3/catalog/partner/connect/ with merchant domain as domain + path

### Expected behavior:
merchant_domain parameter in request should only contain domain from Woo not with path

### Actual behavior:
merchant_domain parameter in request contain domain with path

Currently during domain verification plugin invokes following APIs 
1. https://developers.pinterest.com/docs/redoc/#operation/v3_verification_code_GET to get verification code
2. https://developers.pinterest.com/docs/redoc/#operation/v3_verify_domain_POST to verify domain

This restricts verification to only domain level not with domain+path.

Please switch to following APIs
1. https://developers.pinterest.com/docs/redoc/#operation/v3_verification_code_for_website_GET to get verification code
2. https://developers.pinterest.com/docs/redoc/#operation/v3_verify_website_POST to verify website

Differences in domain and website API: 
- Instead of hostname you need to pass get_home_url() which is passed in merchant connect. 
- pass website in query parameter instead of path as per documentation
- can_claim_multiple param is not needed. can_claim_multiple is True for requests coming for website apis.

alexiglesias31:

Closes #354 .

Switch from domain to websites verification API. This will allow users domain + path level verification.

### Screenshots:

image

1. Setup a WordPress installation that includes a path (e.g: https://example.com/us-site).
2. Install the plugin in the site.
3. The verified domain should include also the path.

### DIFF
--- a/assets/source/setup-guide/app/data/settings/selectors.js
+++ b/assets/source/setup-guide/app/data/settings/selectors.js
@@ -59,11 +59,15 @@ export const isDomainVerified = ( state ) => {
 		return false;
 	}
 
-	const { hostname } = new URL(
+	const { hostname, pathname } = new URL(
 		wcSettings.pinterest_for_woocommerce.homeUrlToVerify
 	);
+
+	// Build url for single site and multisite.
+	const urlToVerify = pathname !== '/' ? hostname + pathname : hostname;
+
 	return state?.settings?.account_data?.verified_user_websites.includes(
-		hostname
+		urlToVerify
 	);
 };
 
--- a/assets/source/setup-guide/app/data/settings/selectors.js
+++ b/assets/source/setup-guide/app/data/settings/selectors.js
@@ -55,14 +55,16 @@ export const isDomainVerified = ( state ) => {
 		return;
 	}
 
-	if ( undefined === state?.settings?.account_data?.verified_domains ) {
+	if ( undefined === state?.settings?.account_data?.verified_user_websites ) {
 		return false;
 	}
 
 	const { hostname } = new URL(
 		wcSettings.pinterest_for_woocommerce.homeUrlToVerify
 	);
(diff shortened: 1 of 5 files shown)


ksere (approved):

findingsimple:
Checking if this is still draft?

budzanowski:
From what I was able to observe the `is_domain_verified()` stops working if the original verification was done using the old code and the user updates to this code. This basically breaks the site. The user would be required to reconnect so the new data would be saved with `update_account_data()`. It looks like `isDomainVerified` JS method may also experience problems on update.

alexiglesias31:
@budzanowski 

> From what I was able to observe the `is_domain_verified()` stops working if the original verification was done using the old code and the user updates to this code. This basically breaks the site. The user would be required to reconnect so the new data would be saved with `update_account_data()`. It looks like `isDomainVerified` JS method may also experience problems on update.

Correct. A migration routine was added in order to prevent the break for existing users.

budzanowski:
@alexiglesias31 thank you for adding the update procedure.
We will need to perform updates more often in different scenarios. This is why I have decided that we need an update framework https://github.com/woocommerce/pinterest-for-woocommerce/pull/390.

Using that PR I have redone the update procedure and copied the commits to https://github.com/woocommerce/pinterest-for-woocommerce/pull/391 Please review. It will replace this PR.

budzanowski:
Closing in favour of #391

PR #368, opened 2022-02-17
Feed Refactor with Action Scheduler Framework.

### LINKED ISSUE #185: Product sync architecture review (supporting sync problems, scaling to many products)

### LINKED ISSUE #242: feed xml: Remove parent item node when child items (variants are present)

### LINKED ISSUE #257: Implement multiple feed files generation ( server-side )

budzanowski:
This supersedes the #271. There was a bad merge with `develop`. I wanted this branch to start with a new develop merge - this time not botched. 

This PR updates the feed generation process.
The main goals for the PR:

- simplified and streamlined generation process
- better error handling
- multi-feed support on the server-side
- predictable scalability
- extensibility of the process

Bonus:
- removes variable parents from XML ( bug #242 )
- if https://github.com/woocommerce/pinterest-for-woocommerce/pull/284 will be merged then additionally #206 will be fixed.

There are very small changes in the UI compared to the original implementation. The feed status no longer shows the total number of products for processing. The rationale for this change comes from two sources:
1. We no longer precalculate the products list and store it in a DB option. Instead, the implementation fetches pages of product ids during each generation cycle. The presence of a large DB option containing hundreds of thousands of items that would be read and updated during batches generation was flagged as a problem by merchants that have used Facebook For WooCommerce plugin. This plugin was using the same method. After changing to the pagination method the merchants have observed better performance. Learning from that experience I have used the same approach here.
2. When we will have multiple feeds the total number of products may not be a very useful number bc each feed may have a different total number.

Usage of Action Scheduler Jobs Framework ( https://github.com/woocommerce/action-scheduler-job-framework ) will be highly beneficial when we will introduce multiple feed generation. During each generation cycle, we will be able to determine if a particular product should be a part of a particular feed and the `ProductsXmlFeed` class will understand the location parameter passed ( currently it does nothing ) to determine the shape of the feed element for that feed.

Most of the parts of the code are ready for multiple feeds. During the refactor it has become obvious that if this capability would not be implemented alongside the refactoring it would be a wasted effort. There are places that still assume that we have a singular feed. Those places need to wait for UI changes part of the Multiple-feeds feature.

This is based on improvements from https://github.com/woocommerce/pinterest-for-woocommerce/pull/240 ( now in develop )

 =

Closes #257
Closes #242
Closes #185 

Please reinstall composer dependencies - this uses a new library and if not installed this will fail ( easiest way would be to just delete vendors folder and then composer install )

There are two distinct scenarios:

- First: fresh install
1. Install the plugin.
2. Make the connection.

- Second: updating over an existing version
0. Install from th developer branch, make the connection.
1. Update the plugin, remember to reinstall composer deps.
2. Old actions should be removed
3. Everything should work as usual except that the feed generation is now controlled by the Action Scheduled framework.

### Changelog entry
Fix - Remove parent variable products from the feed file. ( #242 )
Update - Feed generation process refactor. ( #185 )
Add - Multiple feed backend.
Tweak - Additional logging of the feed generation process.
>


budzanowski:
Tests should pass. There was a change in the WooCommerce core that alters the structure of the code. We need to figure out how to fix our tests after this change.

budzanowski:
Tests fixed upstream.

src/Feeds.php
```diff
@@ -118,8 +118,8 @@ public static function get_merchant_feed_by_location( $merchant_id, $feed_locati
 	 * @return string Returns the ID of the feed if properly registered or an empty string otherwise.
 	 */
 	public static function is_local_feed_registered( $merchant_id ) {
-
-		$local_feed = ProductFeedStatus::get_local_feed();
+		$configs    = LocalFeedConfigs::get_instance()->get_configurations();
+		$local_feed = reset( $configs );
```
alexiglesias31: Minor comment: in [Merchants.php](https://github.com/woocommerce/pinterest-for-woocommerce/pull/368/files#diff-165dc486d272cdd52e9f8f67ea2fd88e3dacf5e7600d6e2e398303fb46cf04d0R131) and [FeedState.php](https://github.com/woocommerce/pinterest-for-woocommerce/pull/368/files#diff-d7374912c385ac51b92ce25d055029bdb751eeb9f7e1cddd17015e182ee015f9R531) we are using `$config` instead of `$local_feed`

src/PluginUpdate.php
```diff
@@ -94,6 +95,19 @@ public function update_plugin_update_version_option(): void {
 		);
 	}
 
+	/**
+	 * List of update procedures
+	 *
+	 * @since x.x.x
+	 * @return array List of update procedures names.
+	 */
+	private function update_procedures() {
```
alexiglesias31, line 104: Minor comment: If the procedures grows in a future, probably we can store them in an array by versions, in order to have a clear view at this method and avoid looking into each procedure.

src/ProductFeedStatus.php
```diff
@@ -18,20 +18,13 @@
...
+		'error_message' => '',
 	);
 
-	/**
-	 * The array that holds the parameters of the feed
-	 *
-	 * @var array
-	 */
-	private static $local_feed = array();
+	const PINTEREST_FOR_WOOCOMMERCE_DATA_PREFIX = PINTEREST_FOR_WOOCOMMERCE_PREFIX . '_feeds_';
```
alexiglesias31: Minor comment: Probably `PINTEREST_FOR_WOOCOMMERCE_FEEDS_DATA_PREFIX` is a bit more clear

src/LocalFeedConfigs.php
```diff
@@ -0,0 +1,119 @@
...
+	private static $instance = null;
+
+	/**
+	 * Singleton initialization and instance fetching method.
+	 *
+	 * @since x.x.x
+	 * @return LocalFeedConfigs Singleton instance.
+	 */
+	public static function get_instance(): LocalFeedConfigs {
+		$cls = static::class;
```
alexiglesias31: Can `$cls` be removed?

src/LocalFeedConfigs.php
```diff
@@ -0,0 +1,119 @@
...
+		return self::$instance;
+	}
+
+	/**
+	 * Class responsible for local feed configurations and handling.
+	 *
+	 * @since x.x.x
+	 */
+	protected function __construct() {
+		$locations = array( Pinterest_For_Woocommerce()::get_base_country() ?? 'US' ); // Replace with multiple countries array for multiple feed config.
```
alexiglesias31, line 58: Imho we should have a filter here. Assuming that the idea is to support multiple countries.

src/LocalFeedConfigs.php
```diff
@@ -0,0 +1,119 @@
...
+			}
+		);
+		$this->feeds_configurations = $feed_ids;
+	}
+
+	/**
+	 * Cleanup local feed configs.
+	 */
+	public function deregister() {
+		Pinterest_For_Woocommerce()::save_data( 'local_feed_ids', false );
```
alexiglesias31, line 102: Probably would be a good idea to clean `feeds_configurations variable` here.

src/FeedRegistration.php
```diff
@@ -0,0 +1,177 @@
...
+	 */
+	private static function register_feed() {
+
+		// Get merchant object.
+		$merchant   = Merchants::get_merchant();
+		$registered = false;
+
+		if ( ! empty( $merchant['data']->id ) && 'declined' === $merchant['data']->product_pin_approval_status ) {
+
+			$registered = false;
```
alexiglesias31: Do we need to set again to false here?

src/LocalFeedConfigs.php
```diff
@@ -0,0 +1,119 @@
...
+
+		$file_name_base = trailingslashit( wp_get_upload_dir()['basedir'] ) . PINTEREST_FOR_WOOCOMMERCE_LOG_PREFIX . '-';
+		$url_base       = trailingslashit( wp_get_upload_dir()['baseurl'] ) . PINTEREST_FOR_WOOCOMMERCE_LOG_PREFIX . '-';
+		array_walk(
+			$feed_ids,
+			function ( &$id ) use ( $file_name_base, $url_base ) {
+				$id = array(
+					'feed_id'   => $id,
+					'feed_file' => "{$file_name_base}{$id}.xml",
+					'tmp_file'  => "{$file_name_base}{$id}.xml",
```
alexiglesias31: Is it intentional that tmp file has the same name of feed? Perhaps would be a good idea to have separate files imho.

alexiglesias31 (commented):
Seems pretty good. Good job @budzanowski. 
Left a few comments here.
Probably could be a good idea send this to our QA for some testing.

pinterest-for-woocommerce.php
```diff
@@ -44,7 +44,7 @@
 }
 
 define( 'PINTEREST_FOR_WOOCOMMERCE_PLUGIN_FILE', __FILE__ );
-define( 'PINTEREST_FOR_WOOCOMMERCE_VERSION', '1.0.8' ); // WRCS: DEFINED_VERSION.
```
JPry: Why is this PR updating the versions in this file?

src/API/FeedIssues.php
```diff
@@ -56,7 +57,7 @@ public function __construct() {
 	public function get_feed_issues( WP_REST_Request $request ) {
 
 		try {
-			$feed_id = Pinterest\ProductSync::get_registered_feed_id();
+			$feed_id = FeedRegistration::get_registered_feed_id();
```
JPry, line 60: ❤️

src/API/FeedState.php
```diff
@@ -524,4 +518,17 @@ private function get_global_error_from_workflow( $workflow ) {
...
+	 * Helper function used for fetching local feed config file.
+	 *
+	 * This is temporary as we will need to operate on multiple feed files in the future.
+	 *
+	 * @return string
+	 */
+	private function get_feed_url() {
+		$configs = LocalFeedConfigs::get_instance()->get_configurations();
+		$config  = reset( $configs );
+		return $config['feed_url'];
```
JPry, line 532: This seems like a good place to use the null coalescing operator rather than `reset()` as a way to grab the first element, especially since `$configs` could be an empty array.

```suggestion
		return $configs[0]['feed_url'] ?? '';
```

src/API/FeedState.php
```diff
@@ -163,26 +164,20 @@ public function get_feed_state() {
...
 				$extra_info   = sprintf(
-					/* Translators: %1$s Time string, %2$s current index of written products, %3$s total number of products, %4$s Opening link tag, %5$s closing link tag */
-					esc_html__( 'Last activity: %1$s ago - Wrote %2$s out of %3$s products to %4$sfeed file%5$s.', 'pinterest-for-woocommerce' ),
+					/* Translators: %1$s Time string, %2$s number of products */
+					esc_html__( 'Last activity: %1$s ago - Wrote %2$s products to %3$sfeed file%4$s.', 'pinterest-for-woocommerce' ),
 					human_time_diff( $state['last_activity'] ),
-					$state['current_products'],
 					$state['product_count'],
-					'<a href="' . $local_feed['feed_url'] . '" target="_blank">',
+					'<a href="' . $this->get_feed_url() . '" target="_blank">',
```
JPry: For readability, I think `sprintf()` would be preferable here, and we should also be escaping the URL.

```suggestion
					sprintf( '<a href="%s" target="_blank">', esc_url( $this->get_feed_url() ) ),
```

Also, do we want to have any kind of handling (either here or in the `get_feed_url()` method) for getting no result back for the local feed?

src/API/FeedState.php
```diff
@@ -191,12 +186,11 @@ private function add_local_feed_state( $result ) {
...
 				$extra_info   = sprintf(
-					/* Translators: %1$s Time string, %2$s current index of written products, %3$s total number of products, %4$s Opening link tag, %5$s closing link tag */
-					esc_html__( 'Successfully generated %1$s ago - Wrote %2$s out of %3$s products to %4$sfeed file%5$s.', 'pinterest-for-woocommerce' ),
+					/* Translators: %1$s Time string, %2$s total number of products */
+					esc_html__( 'Successfully generated %1$s ago - Wrote %2$s products to %3$sfeed file%4$s', 'pinterest-for-woocommerce' ),
 					human_time_diff( $state['last_activity'] ),
-					$state['current_products'],
 					$state['product_count'],
-					'<a href="' . $local_feed['feed_url'] . '" target="_blank">',
+					'<a href="' . $this->get_feed_url() . '" target="_blank">',
```
JPry: Same as above.

```suggestion
					sprintf( '<a href="%s" target="_blank">', esc_url( $this->get_feed_url() ) ),
```

src/Utilities/FeedLogger.php
```diff
@@ -0,0 +1,34 @@
...
+namespace Automattic\WooCommerce\Pinterest\Utilities;
+
+use Automattic\WooCommerce\Pinterest\Logger;
+
+/**
+ * Trait FeedLogger
+ *
+ * @since x.x.x
+ */
+trait FeedLogger {
```
JPry: One minor thing: because the `$level` used for the logger is `product-sync`, I feel like this should be `ProductSyncLogger` rather than `FeedLogger`.

src/ProductsXmlFeed.php
```diff
@@ -84,18 +84,25 @@ public static function get_xml_footer() {
...
 		if ( ! self::is_product_fit_for_feed( $product ) ) {
 			return '';
 		}
 
 		$xml = "\t\t<item>" . PHP_EOL;
 
+		/**
+		 * Filter that controls the attributes that will be added to the product XML file.
+		 *
+		 * @since x.x.x
```
JPry: Since we're just now adding this comment and the next one, I think we should include the actual version where these were introduced.

src/ProductsXmlFeed.php
```diff
@@ -84,18 +84,25 @@ public static function get_xml_footer() {
...
 	 *
-	 * @param WC_Product $product  The product to print the XML for.
-	 *
-	 * @return string XML string
+	 * @param  WC_Product $product  The product to print the XML for.
+	 * @param  string     $location The location to print the XML for.
+	 * @return string     XML string.
 	 */
-	public static function get_xml_item( $product ) {
+	public static function get_xml_item( $product, $location ) {
```
JPry, line 91: The `$location` variable isn't used. Is this something that is meant for a future change?

src/FeedRegistration.php
```diff
@@ -0,0 +1,177 @@
...
+		if ( ! self::feed_file_exists() ) {
+			self::log( 'Feed didn\'t fully generate yet. Retrying later.', 'debug' );
+			// Feed is not generated yet, lets wait a bit longer.
+			return true;
+		}
+
+		try {
+			$registered = self::register_feed();
+
+			if ( $registered ) {
```
JPry: It looks like we don't need the `$registered` variable here.

```suggestion
			if ( self::register_feed() ) {
```

src/LocalFeedConfigs.php
```diff
@@ -0,0 +1,119 @@
...
+ * At the same time allow distributed use of configurations.
+ */
+class LocalFeedConfigs {
+
+	/**
+	 * Array of local feed configurations.
+	 *
+	 * @var array $feeds_configurations
+	 */
+	private $feeds_configurations = null;
```
JPry: Since this variable is marked as type `array`, it should be instantiated as an empty array.

```suggestion
	private $feeds_configurations = [];
```

src/LocalFeedConfigs.php
```diff
@@ -0,0 +1,119 @@
...
+	 * Prepare feed configurations.
+	 *
+	 * @since x.x.x
+	 * @param array $locations Array of location to generate the feed files for.
+	 */
+	private function initialize_local_feeds_config( $locations ) {
+
+		if ( $this->feeds_configurations ) {
+			return;
+		}
```
JPry: This method is only called from the constructor, and is `private`. I don't think we need this check here.

```suggestion
```

src/LocalFeedConfigs.php
```diff
@@ -0,0 +1,119 @@
...
+
+		$feed_ids = Pinterest_For_Woocommerce()::get_data( 'local_feed_ids' );
+		if ( ! $feed_ids ) {
+			$feed_ids = array();
+		}
```
JPry:
```suggestion
		$feed_ids = Pinterest_For_Woocommerce()::get_data( 'local_feed_ids' ) ?? array();
```

src/Merchants.php
```diff
@@ -127,13 +127,14 @@ private static function get_merchant_id_from_advertiser() {
...
-		$local_feed = ProductFeedStatus::get_local_feed();
+		$configs = LocalFeedConfigs::get_instance()->get_configurations();
+		$config  = reset( $configs );
 
 		$merchant_name = apply_filters( 'pinterest_for_woocommerce_default_merchant_name', esc_html__( 'Auto-created by Pinterest for WooCommerce', 'pinterest-for-woocommerce' ) );
 
 		$args = array(
 			'merchant_domains' => get_home_url(),
-			'feed_location'    => $local_feed['feed_url'],
+			'feed_location'    => $config['feed_url'],
```
JPry, line 137: What happens when for some reason there is no data returned and this is unset, or `feed_url` doesn't exist?

src/PluginUpdate.php
JPry: +1 to this

JPry (changes requested):

src/Merchants.php
budzanowski: We always have configs, at least one. The LocalFeedConfigs is created this way. There feed file may not be generated yet and there can be nothing to fetch under `feed_url` but the URL already exists in the config.

pinterest-for-woocommerce.php
budzanowski: I have accidentally committed this. Will remove. The update procedures are targeting particular versions so one needs to bump these values in order to trigger the updates.

src/API/FeedState.php
budzanowski: Configs will not be an empty array. It always has at least one config.
We also don't know how to fetch the fist element usin `??` because the `$config` array uses country slugs as keys.
Also this code will be altered when we will introduce full multifeed functionality.

src/LocalFeedConfigs.php
budzanowski: I think that we will not use filter here. Here is why.

For now we only use what is set in `get_base_country` or use default `US`.
When multifeed capability will be added the country list will be handled from the UI.
The UI part is an assumption that it will be up to the merchant, and should be easy to configure by them.

src/API/FeedState.php
budzanowski: sprintf is a good suggestion.
If the feed was not generated yet, this part of UI is not used, so we don't need to worry about no results.

src/Feeds.php
budzanowski: True.

src/LocalFeedConfigs.php
budzanowski: I am marking as resolved but please reopen in case you think that the filter is needed.

src/FeedRegistration.php
budzanowski: Def no.

src/LocalFeedConfigs.php
budzanowski: Nice catch. It was correct one commit before. I was cleaning up and by accident removed the tml part. Great spotting.

src/LocalFeedConfigs.php
budzanowski: That is a nice simplification. The singleton patter was added in one of the last commits and made this part redundant. Cool, removing.

src/LocalFeedConfigs.php
budzanowski: 👍

src/ProductsXmlFeed.php
budzanowski: Yes, it will allow us to control what shipping information is sent in Feed for the particular location and it will allow the creation of logic for currency swap and other types of internationalization.

src/ProductsXmlFeed.php
budzanowski: True.

src/PluginUpdate.php
budzanowski: I have created a separate issue for that: https://github.com/woocommerce/pinterest-for-woocommerce/issues/414

src/Utilities/FeedLogger.php
budzanowski: Agree. Changed.

src/LocalFeedConfigs.php
budzanowski: 👍

budzanowski:
@alexiglesias31 & @JPry 

Thank you very much for the review! ❤️
I have addressed all of the things that required code changes and commented in places where I thought that the changes are not necessary.

I have marked all of the discussions as resolved, but this was mostly for myself to be able to track what has been addressed and what has not. Please comment and unresolve anything that you think needs more explanation.

src/ProductSync.php
```diff
@@ -236,387 +127,34 @@ public static function is_product_sync_enabled() {
...
-	 *
-	 * @return boolean true if rescheduled, false otherwise.
 	 */
-	private static function trigger_async_feed_generation( $force = false ) {
+	private static function deregister() {
 
-		if ( $force || false === as_next_scheduled_action( self::ACTION_FEED_GENERATION, array(), PINTEREST_FOR_WOOCOMMERCE_PREFIX ) ) {
-			return 0 !== as_enqueue_async_action( self::ACTION_FEED_GENERATION, array(), PINTEREST_FOR_WOOCOMMERCE_PREFIX );
-		}
+		self::$feed_generator->deregister();
```
alexiglesias31: @budzanowski there is another issue here. If the user disconnect the account with `Product Sync` option disabled it causes a fatal error. That happens because if `Product Sync` is disabled, the feed components are never initialized [here](https://github.com/woocommerce/pinterest-for-woocommerce/pull/368/files#diff-9eb8ac4928cebc9cc9e52da272589aafa0747ad026d7f101c0f9d6b9951af91fR55)

alexiglesias31 (changes requested):

src/ProductSync.php
budzanowski: Changed everything into static so initialization is not necessary.

src/API/FeedState.php
JPry: Ah, I missed that it was keyed with a country slug rather than a numerical index. This makes sense 👍🏻

JPry (approved):
Looks good to me!

alexiglesias31 (approved):

alexiglesias31:
@budzanowski, @jconroy.
We finished the QA for this branch. LGTM!

findingsimple:
Great, thank you @alexiglesias31

PR #371, opened 2022-02-18
Escape XML special chars in SKU for the XML MPN section.

### LINKED ISSUE #365: When special characters are used on the SKU field of products, the feed is created with errors.
### Steps to reproduce:

1. Edit a product and add a special character on the SKU field as "&"
2. Install "Pinterest for WooCommerce" and do the ondobarding 
3. Cehck the feed created

### Expected behavior:
Feed should be created with out problems

### Actual behavior:
Feed is created with errors.

budzanowski:

Escape special chars for XML during MPN/SKU fetching for the feed file.

Closes #365  .

1. Create a simple product.
2. Assign a SKU to the product that has one of the XML special chars like `&`
3. Check for the feed file generation.
4. The MPN xml node for the product should have the special char escaped.

### Changelog entry

> Update - Escape XML special chars in SKU for the XML MPN section.

### DIFF
--- a/src/ProductsXmlFeed.php
+++ b/src/ProductsXmlFeed.php
@@ -395,7 +395,7 @@ class ProductsXmlFeed {
 	 * @return string
 	 */
 	private static function get_property_g_mpn( $product, $property ) {
-		return '<' . $property . '>' . $product->get_sku() . '</' . $property . '>';
+		return '<' . $property . '>' . esc_xml( $product->get_sku() ) . '</' . $property . '>';
 	}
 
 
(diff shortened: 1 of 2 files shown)


mikkamp (approved):
This is working well and the characters are being escaped. So everything is working as expected.

Although I've wondered about this before since we should be escaping as late as possible. For shipping name, or category name (used in product_type) we rely on the fact that the characters are saved in the DB in it's escaped form. Should we also be cautious for those? And if so what about creating a helper function for outputting a property to the feed and escaping it.

PR #373, opened 2022-02-21
Fix fatal error on deactivating or uninstalling the plugin

### LINKED ISSUE #360: [p1] Uninstalling/deactivating the plugin fails when I had WooCommerce plugin installed but not activated
### Describe the bug:
Uninstalling/deactivating the plugin fails when I had WooCommerce plugin installed but not activated
### Steps to reproduce:

1. Create a new jurassic.ninja instance
2. Install WooCommerce plugin but DONT activate it
3. Install WooCommerce for pinterest plugin (1.0.3 used for the bug) and activate it
3. Uninstalling/deactivating the WooCommerce for pinterest plugin 

### Expected behavior:
No errors during the process.

### Actual behavior:
Saw this stacktrace error
"
Fatal error: Uncaught Error: Call to undefined function Automattic\WooCommerce\Pinterest\as_unschedule_all_actions() in /srv/users/usera278ef7d/apps/usera278ef7d/public/wp-content/plugins/pinterest-for-woocommerce/src/ProductSync.php:661 Stack trace: #0 /srv/users/usera278ef7d/apps/usera278ef7d/public/wp-content/plugins/pinterest-for-woocommerce/pinterest-for-woocommerce.php(119): Automattic\WooCommerce\Pinterest\ProductSync::cancel_jobs() #1 /srv/users/usera278ef7d/apps/usera278ef7d/public/wp-includes/class-wp-hook.php(307): {closure}(false) #2 /srv/users/usera278ef7d/apps/usera278ef7d/public/wp-includes/class-wp-hook.php(331): WP_Hook->apply_filters('', Array) #3 /srv/users/usera278ef7d/apps/usera278ef7d/public/wp-includes/plugin.php(474): WP_Hook->do_action(Array) #4 /srv/users/usera278ef7d/apps/usera278ef7d/public/wp-admin/includes/plugin.php(812): do_action('deactivate_pint...', false) #5 /srv/users/usera278ef7d/apps/usera278ef7d/public/wp-admin/plugins.php(209): deactivate_plugins('pinterest-for-w...', false, false) in /srv/users/usera278ef7d/apps/usera278ef7d/public/wp-content/plugins/pinterest-for-woocommerce/src/ProductSync.php on line 661"

### Additional details:

Screen Shot 2022-02-08 at 1 44 54 PM

alexiglesias31:

Closes #360  .

Fix the fatal error if the plugin is uninstalled or deactivated and WooCommerce is installed but no active.
---
- Verify if action scheduler actions exists during deactivation of the plugin.
- Unschedule tasks on WooCommerce deactivation.

1. Install WooCommerce plugin but don't activate it.
2. Install Pinterest for WooCommerce plugin and activate it.
3. Uninstalling/deactivating the Pinterest for WooCommerce plugin should not throw fatal error.

### Changelog entry

> Fix - Fatal error on plugin deactivation if WooCommerce is not active.

### DIFF
--- a/class-pinterest-for-woocommerce.php
+++ b/class-pinterest-for-woocommerce.php
@@ -247,6 +247,9 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 
 			// Disconnect advertiser if advertiser or tag change.
 			add_action( 'update_option_pinterest_for_woocommerce', array( $this, 'maybe_disconnect_advertiser' ), 10, 2 );
+
+			// Unschedule tasks if woocommerce is deactivated.
+			add_action( 'deactivate_woocommerce/woocommerce.php', array( Pinterest\ProductSync::class, 'cancel_jobs' ) );
 		}
 
 
--- a/class-pinterest-for-woocommerce.php
+++ b/class-pinterest-for-woocommerce.php
@@ -249,7 +249,9 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 			add_action( 'update_option_pinterest_for_woocommerce', array( $this, 'maybe_disconnect_advertiser' ), 10, 2 );
 
 			// Unschedule tasks if woocommerce is deactivated.
-			add_action( 'deactivate_woocommerce/woocommerce.php', array( Pinterest\ProductSync::class, 'cancel_jobs' ) );
+			if ( defined( 'WC_PLUGIN_FILE' ) ) {
+				register_deactivation_hook( WC_PLUGIN_FILE, array( Pinterest\ProductSync::class, 'cancel_jobs' ) );
+			}
 		}
 
 
--- a/src/ProductSync.php
+++ b/src/ProductSync.php
@@ -690,6 +690,5 @@ class ProductSync {
 
 		as_unschedule_all_actions( self::ACTION_HANDLE_SYNC, array(), PINTEREST_FOR_WOOCOMMERCE_PREFIX );
 		as_unschedule_all_actions( self::ACTION_FEED_GENERATION, array(), PINTEREST_FOR_WOOCOMMERCE_PREFIX );
-		
 	}
 }
--- a/src/ProductSync.php
+++ b/src/ProductSync.php
@@ -684,7 +684,9 @@ class ProductSync {
 	 * @return void
 	 */
 	public static function cancel_jobs() {
-		as_unschedule_all_actions( self::ACTION_HANDLE_SYNC, array(), PINTEREST_FOR_WOOCOMMERCE_PREFIX );
-		as_unschedule_all_actions( self::ACTION_FEED_GENERATION, array(), PINTEREST_FOR_WOOCOMMERCE_PREFIX );
+		if ( function_exists( 'as_unschedule_all_actions' ) ) {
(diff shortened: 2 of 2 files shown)


ksere (approved):

src/ProductSync.php
```diff
@@ -684,7 +684,9 @@ public static function reschedule_if_errored() {
 	 * @return void
 	 */
 	public static function cancel_jobs() {
-		as_unschedule_all_actions( self::ACTION_HANDLE_SYNC, array(), PINTEREST_FOR_WOOCOMMERCE_PREFIX );
-		as_unschedule_all_actions( self::ACTION_FEED_GENERATION, array(), PINTEREST_FOR_WOOCOMMERCE_PREFIX );
+		if ( function_exists( 'as_unschedule_all_actions' ) ) {
+			as_unschedule_all_actions( self::ACTION_HANDLE_SYNC, array(), PINTEREST_FOR_WOOCOMMERCE_PREFIX );
+			as_unschedule_all_actions( self::ACTION_FEED_GENERATION, array(), PINTEREST_FOR_WOOCOMMERCE_PREFIX );
+		}
```
JPry: Minor: It is generally a better practice to return early rather than enclose actions within a conditional block, even though this is just a small block.

```suggestion
		if ( ! function_exists( 'as_unschedule_all_actions' ) ) {
			return;
		}

		as_unschedule_all_actions( self::ACTION_HANDLE_SYNC, array(), PINTEREST_FOR_WOOCOMMERCE_PREFIX );
		as_unschedule_all_actions( self::ACTION_FEED_GENERATION, array(), PINTEREST_FOR_WOOCOMMERCE_PREFIX );
		
```

class-pinterest-for-woocommerce.php
```diff
@@ -247,6 +247,9 @@ public function init_plugin() {
 
 			// Disconnect advertiser if advertiser or tag change.
 			add_action( 'update_option_pinterest_for_woocommerce', array( $this, 'maybe_disconnect_advertiser' ), 10, 2 );
+
+			// Unschedule tasks if woocommerce is deactivated.
+			add_action( 'deactivate_woocommerce/woocommerce.php', array( Pinterest\ProductSync::class, 'cancel_jobs' ) );
```
JPry: There is an edge case that I think we should consider here: sometimes the WooCommerce plugin can be uploaded to a site with a main directory named something other than `woocommerce`. In that scenario, this action will be missed. 

I think we should utilize the [`WC_PLUGIN_FILE`](https://github.com/woocommerce/woocommerce/blob/aa8b7a4a6b7d21ccddaac360766c4daa396b9a98/plugins/woocommerce/woocommerce.php#L19-L21) constant to help here. This method we're in is run on the `plugins_loaded` hook, so that means that WooCommerce will have been loaded by this point (if it is active).

```suggestion
			if ( defined( 'WC_PLUGIN_FILE' ) ) {
				register_deactivation_hook( WC_PLUGIN_FILE, array( Pinterest\ProductSync::class, 'cancel_jobs' ) );
			}
```

I think that this approach will overcome this edge case, as the `WC_PLUGIN_FILE` constant will be defined based on the actual path, regardless of any directory naming differences.

JPry (changes requested):

findingsimple:
@JPry can you please give this another review

JPry (approved):
Looks good to me! (once merge conflicts are resolved, of course)

PR #383, opened 2022-03-08
Take full size images

### LINKED ISSUE #374: Low resolution pictures
### Describe the bug:
some merchants are complaining on the resolution of the images that gets the plugin.

complete thread: https://wordpress.org/support/topic/low-resolution-pictures/

### Expected behavior:
feed is created with the correct resolution of the image

### Additional details:
could be related to #248

alexiglesias31:

Closes #374.

Take the full size image for xml instead of `'woocommerce_single'` size.
- If the image is bigger than [suggested limit](https://github.com/woocommerce/pinterest-for-woocommerce/issues/374#issuecomment-1062288825) `'woocommerce_single'` will be used in its place.
- The default full size can be filtered with `'pinterest_for_woocommerce_feed_image_size'`.

2. Check on the feed that the images are the full size ones.

### Changelog entry

> Tweak - Take full size images for the feed.

### DIFF
--- a/src/ProductsXmlFeed.php
+++ b/src/ProductsXmlFeed.php
@@ -60,13 +60,6 @@ class ProductsXmlFeed {
 	 */
 	const DESCRIPTION_SIZE_CHARS_LIMIT = 10000;
 
-	/**
-	 * Max image resolution (5000px*5000px).
-	 *
-	 * @var int
-	 */
-	const IMAGE_MAX_RESOLUTION = 5000 * 5000;
-
 
 	/**
 	 * Returns the XML header to be printed.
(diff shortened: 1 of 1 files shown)


ksere (approved):

JPry:
@alexiglesias31 @ksere Since the image size has come up a few times, I would propose that we add a filter to the image size we use here to give users greater control over what image size is used for their site.

budzanowski:
I don't think that going with 5000 x 5000 is a good idea. It is too arbitrary. If we want to set a limit then we should go with what is allowed by the API. That being said we are looking at a very massive resolutions, overkill for anything reasonable. Just going with 'full' and a filter if anyone wants something different is enough in my opinion.

ksere (approved):

budzanowski:
@alexiglesias31 is this good to go?

budzanowski (approved):

budzanowski:
As discussed in Slack this is good to go. Changing to `Ready for Review` and merging.

PR #390, opened 2022-03-13
Add plugin update framework.

budzanowski:

Add plugin update framework.
The code in this PR allows adding plugin update procedures in a centralized location. It provides structure and helper methods that allow clean implementation of migrations, updates, DB patches, and so on.

This PR was created in a response to https://github.com/woocommerce/pinterest-for-woocommerce/pull/358 where I have observed that we have the update code intertwined with application code.   

No testing, this is just a framework.
Unit tests are added to verify the basic behavior of the update component.

### Additional details:

How to use the framework?
Create your update procedure as a method in the PluginUpdate class.
Add it to the `perform_plugin_updates()` method.
Inside your update method test using `version_needs_update`  if the update should be performed.

Check https://github.com/woocommerce/pinterest-for-woocommerce/pull/391 to see a real-life implementation.

### Changelog entry

> Add -  Plugin update framework.


budzanowski:
After the review in #391 I will just merge this. On its own this does not affect any business logic.

PR #391, opened 2022-03-14
Use website verification instead of domains.

### LINKED ISSUE #354: Few api endpoint /v3/catalog/partner/connect/ requests are failing

budzanowski:

Closes #354 .

This PR is an implementation of #358 ( Use website verification instead of domains ) using an Update Framework that I have created in #390.

When I have noticed that #358 requires an update procedure for the plugin I have decided that it will be better if we will decouple the plugin execution code with the update procedure code. This is why #390 was created. This PR reworks #358 by moving the update code to a specialized update procedure in the update class.

1. Starting from the `develop` branch connect to Pinterest.
2. Checkout this branch.
3. Reload the page.
4. Observe that the feed is broken and the site is not verified. Feed claims that it is disabled but that is not true.
image

6. Change manually in code `PINTEREST_FOR_WOOCOMMERCE_VERSION` to `1.0.9`
7. Reload the page.
8. Feed should work again.

During the release, the version in `PINTEREST_FOR_WOOCOMMERCE_VERSION` will be changed by the release script which means that the update code will be triggered during the first plugin load.

### Additional details:

After #390 will be merged this will be automatically targeted at `develop` branch by Facebook.

### Changelog entry

> Tweak - Use website verification instead of domain verification.

### DIFF
--- a/assets/source/setup-guide/app/data/settings/selectors.js
+++ b/assets/source/setup-guide/app/data/settings/selectors.js
@@ -59,11 +59,15 @@ export const isDomainVerified = ( state ) => {
 		return false;
 	}
 
-	const { hostname } = new URL(
+	const { hostname, pathname } = new URL(
 		wcSettings.pinterest_for_woocommerce.homeUrlToVerify
 	);
+
+	// Build url for single site and multisite.
+	const urlToVerify = pathname !== '/' ? hostname + pathname : hostname;
+
 	return state?.settings?.account_data?.verified_user_websites.includes(
-		hostname
+		urlToVerify
 	);
 };
 
--- a/assets/source/setup-guide/app/data/settings/selectors.js
+++ b/assets/source/setup-guide/app/data/settings/selectors.js
@@ -55,14 +55,16 @@ export const isDomainVerified = ( state ) => {
 		return;
 	}
 
-	if ( undefined === state?.settings?.account_data?.verified_domains ) {
+	if ( undefined === state?.settings?.account_data?.verified_user_websites ) {
 		return false;
 	}
 
 	const { hostname } = new URL(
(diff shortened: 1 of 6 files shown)


alexiglesias31 (approved):
LGTM.
Just a minor comment for future work in order to improve a bit the maintainability: we could group the required callbacks for each version in a kind of array, but probably this would be wiser if the required migrations grow too much. For now we are ok as this imho.

PR #395, opened 2022-03-17
Show proper error if merchant is connected to a different platform

### LINKED ISSUE #364: [p2] User cannot be connected to multiple active accounts
### Describe the bug:
A user cannot both register and shopify and Woocommerce. The API will fail. Since connecting the partner to feed and related functions happen after user registration process, there is an error. However that error is not displayed correctly to the user
### Steps to reproduce:

1. Create a new pinterest business account or have an account with no stores connected
2. first goto shopify and create a store and install pinterest app from app store
3. connect the business account to pinterest app in shopify
4. create a shop in woocommerce and install the pinterest plugin. 
5. connect the business account to woo commerce

### Expected behavior:
A clear error message is shown to user either during registration or in the setup screen when the error happens during feed process

### Actual behavior:
No error message shown

### Additional details:
Screen Shot 2022-02-10 at 9 56 40 AM

alexiglesias31:

Closes #364.

This PR adds a proper error when the merchant is connected to a different e-commerce platform (like Shopify).

### Screenshots:

image

1. Create a new Pinterest business account or have an account with no stores connected.
2. Go to Shopify, create a store, install Pinterest app and connect your business account.
3. Create a store in WooCommerce, install the plugin and connect your business account.
4. A clear error should be prompted as shows the attached picture.

### Changelog entry

> Add - Clear error when the merchant is connected to another e-commerce platform.

### DIFF
--- a/src/Merchants.php
+++ b/src/Merchants.php
@@ -127,13 +127,14 @@ class Merchants {
 	 */
 	public static function update_or_create_merchant() {
 
-		$local_feed = ProductFeedStatus::get_local_feed();
+		$configs = LocalFeedConfigs::get_instance()->get_configurations();
+		$config  = reset( $configs );
 
 		$merchant_name = apply_filters( 'pinterest_for_woocommerce_default_merchant_name', esc_html__( 'Auto-created by Pinterest for WooCommerce', 'pinterest-for-woocommerce' ) );
 
 		$args = array(
 			'merchant_domains' => get_home_url(),
-			'feed_location'    => $local_feed['feed_url'],
+			'feed_location'    => $config['feed_url'],
 			'feed_format'      => 'XML',
 			'country'          => Pinterest_For_Woocommerce()::get_base_country() ?? 'US',
 			'locale'           => str_replace( '_', '-', determine_locale() ),
--- a/src/Merchants.php
+++ b/src/Merchants.php
@@ -148,9 +148,6 @@ class Merchants {
 			throw new Exception( __( 'Response error when trying create a merchant or update the existing one.', 'pinterest-for-woocommerce' ), 400 );
 		}
 
-		// Clean merchants error code.
-		Pinterest_For_Woocommerce()::save_data( 'merchant_connected_diff_platform', false );
-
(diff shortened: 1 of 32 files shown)


ksere (approved):
Ideally we shouldn't save this error status to the database but I don't see any other viable approach without adding more load (outgoing remote requests) to the health check endpoint.

src/ProductSync.php
```diff
@@ -278,6 +278,11 @@ public static function handle_feed_registration() {
 			throw new Exception( esc_html__( 'Could not register feed.', 'pinterest-for-woocommerce' ) );
 
 		} catch ( Throwable $th ) {
+			if ( 4163 === $th->get_pinterest_code() ) {
```
adarshakb: We are handling this error, what if other errors happen in partner connect call?

alexiglesias31: Indeed, we are capturing this specific error.
If there is a well-defined list with the possible errors for this kind of scenario, we could add an error codes map.
However, that is something that should be tackled as an enhancement imho.

alexiglesias31:
@jconroy this PR is ready for CR

src/Merchants.php
```diff
@@ -148,6 +148,9 @@ public static function update_or_create_merchant() {
 			throw new Exception( __( 'Response error when trying create a merchant or update the existing one.', 'pinterest-for-woocommerce' ), 400 );
 		}
 
+		// Clean merchants error code.
+		Pinterest_For_Woocommerce()::save_data( 'merchant_connected_diff_platform', false );
```
budzanowski: I would remove it from here and add it to the handle_feed_registration. This way it is always cleared and only set if we get to this particular 4163 code.

First of all, we would have handled `merchant_connected_diff_platform` around the same place, thus easier to maintain. Second, it would always get cleared meaning that if the user setup would be triggering another type of error ( for whatever reason ) we would have only one error to focus on.

alexiglesias31: @budzanowski, good point. Done.

budzanowski (approved):

budzanowski:
:shipit:

PR #434, opened 2022-04-04
Fix track event emitter matching.

tomalec:

- Fix track event emitter matching.

For example, distinguish emitters of `gla_setup_ads` from `gla_setup_ads_faq`.
Copied from woocommerce/google-listings-and-ads@59238f8

So far there was no colliding case in this repo, but one may appear in the future. Consider it devDependency update, given that we don't have a separate repo for this tool.

- [Regenerate tracking docs to bump code links.](https://github.com/woocommerce/pinterest-for-woocommerce/commit/76496507f5e0f239d44fec3bdd1f1b465ccf82a6)

1. Document an event which name is a subset of an existing event, for example:
   ```js
   /**
    * Substr
    * @event wcadmin_pfw_business_account
    */
   ```
3. Generate the docs `npm run doc:tracking`
4. See that emitters of `wcadmin_pfw_business_account_connect_button_click` are not reported as emitters of `wcadmin_pfw_business_account`.

### Changelog entry

> Tweak - Updated the Track event generation script to prevent future errors.

### DIFF
--- a/TRACKING.md
+++ b/TRACKING.md
@@ -17,7 +17,7 @@ Do not edit it manually!
 ### [`wcadmin_pfw_account_connect_button_click`](assets/source/setup-guide/app/components/Account/Connection.js#L37)
 Clicking on "Connect" Pinterest account button.
 #### Emitters
-- [`AccountConnection`](assets/source/setup-guide/app/components/Account/Connection.js#L80)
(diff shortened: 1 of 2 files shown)


mikkamp (approved):
Since the originating PR has been merged, I see no reason why we can't merge this one as well.

I tested the example and confirmed that `wcadmin_pfw_business_account_connect_button_click` is no longer reported as an emitter (before the PR it was).

PR #435, opened 2022-04-05
Onboarding inbox notifications.

budzanowski:

Add inbox notifications for various events related to the onboarding process:
- complete onboarding ( after 3, 7, 14, 30 days )
- enable sync
- sync error

### Screenshots:
Complete onboarding after 3 days:
image

Complete onboarding after 7, 14, 30 days:
image

Enable Sync:
image

Sync Error:
image

Since the inbox code is run once a day, via immediate action scheduler action that gets set up by cron heartbeat it is useful to have a way to trigger the cron actions. This can be done using a plugin like **WP Crontrol** or using **wpcli**:
```bash
wp cron event run pinterest_for_woocommerce_daily_heartbeat_cron
```

1. Complete onboarding after 3 days inbox message gets sent when the merchant has installed the plugin, but has not finished the onboarding and has 5 or more completed purchases:
    1. Install the plugin
    2. Do not complete the setup
    3. Have 5 completed orders.
    4. Trigger the daily heartbeat
    5. Wait 3 days ( kidding ), check wp_option  `pinterest-for-woocommerce-marketing-notifications-init-timestamp` and remove 4 days' worth of seconds from it.
    6. Trigger daily heartbeat again. 
    7. Go to WooCommerce > Home, and check the inbox. You may need to refresh the location once or twice before the message gets rendered.
    8. Click `Complete Setup` to check that the link works.

2. Complete onboarding after 7, 14, and 30 days. This is the same scenario as with 1. Just the timestamp option needs to be pulled back in time more.

3. Enable sync note is issued when the account is connected, and the product sync is disabled in settings.
     1. Connect account
     2. Go to Marketing > Pinterest > Settings and disable `Enable Product Sync`
     3. A new timestamp was recorded after connection. This time `pinterest-for-woocommerce-account-connection-timestamp`. Reduce 4 days' worth of time.
     4. Trigger daily heartbeat.
     5. Visit WooCommerce > Home to check that the Enable sync message is there.
     6. Check that the message link works.

4. Sync Error note is sent when the connection to the account is completed, sync is enabled but there are errors in the ingestion process. **If testing of this will not be feasible please review the code - I have tested this.**
    1. Setup the plugin and connect.
    2. Make the ingestion fail ( ¯\_(ツ)_/¯ ) for example leave the account connected for a day but make the feed not accessible ( disable local site ), so the next feed fetch will fail.
    3. When you see the error in the feed status trigger the daily hearbeat.
    4. Visit WooCommerce > Home to check that the Error sync message is there.

### Additional details:

- In order to minimize the resources footprint a daily heartbeat process was added. This allows testing the notes trigger condition just once a day. It may mean that the note's trigger time is offset +1 day but that does not change the usability of this solution. Making it more exact with timing would make the code much more complicated without a real benefit.

- When the plugin gets installed we don't have any information about when that has happened. We also do not have any information about when a successful connection has been made. Both of these values are required to know when a message should be sent to the inbox. For this purpose, this code adds two timestamp markers that help us to assume missing values.
The first marker is set when the inbox messages for the plugin code is triggered for the first time: https://github.com/woocommerce/pinterest-for-woocommerce/blob/09aa302659fc2c6662d1c860fa8be174c3a758ba/src/Notes/MarketingNotifications.php#L69 First call sets the marker. The notifications use this is a general start time ( fake plugin install time ). Because we can't know when the plugin was installed or not. This is only for the purpose of the Notifications so this is why we have it here. <br>
The second time is added to general utilities. The role of this timer is to catch the connection time. It is tied to the last step of the connection process. If the user changes tag settings this will get reset which makes sense because the connection parameters have been changed. ofc this will also get reset on a full reconnection. 
https://github.com/woocommerce/pinterest-for-woocommerce/blob/09aa302659fc2c6662d1c860fa8be174c3a758ba/src/Utilities/Utilities.php#L45

### Changelog entry

> Add - Inbox notifications for the onboarding process and catalog sync.


class-pinterest-for-woocommerce.php
```diff
@@ -153,6 +153,9 @@ public function maybe_init_plugin() {
 
 			$this->define_constants();
 
+			// Load utilities manually. This is not a class so autoloader misses it.
+			require_once 'src/Utilities/Utilities.php';
```
JPry: We should use the full path rather than relying on relative resolution.

```suggestion
			require_once __DIR__ . '/src/Utilities/Utilities.php';
```

class-pinterest-for-woocommerce.php
```diff
@@ -226,6 +229,10 @@ public function init_plugin() {
 
 			$this->includes();
 
+			// Start the heartbeat.
+			$this->heartbeat = new Pinterest\Heartbeat( WC()->queue() );
```
JPry: I know prior art here is to use relative class references, but I think we should start using imports. At a later date we can look into getting older code to match.

```suggestion
			$this->heartbeat = new Heartbeat( WC()->queue() );
```

class-pinterest-for-woocommerce.php
```diff
@@ -252,6 +259,9 @@ public function init_plugin() {
 			// Disconnect advertiser if advertiser or tag change.
 			add_action( 'update_option_pinterest_for_woocommerce', array( $this, 'maybe_disconnect_advertiser' ), 10, 2 );
 
+			// Init marketing notifications.
+			add_action( Pinterest\Heartbeat::DAILY, array( $this, 'init_marketing_notifications' ) );
```
JPry: Same as above.

```suggestion
			add_action( Heartbeat::DAILY, array( $this, 'init_marketing_notifications' ) );
```

class-pinterest-for-woocommerce.php
```diff
@@ -283,6 +293,16 @@ public function admin_init() {
 			$variation_attributes->register();
 		}
 
+		/**
+		 * Init marketing notifications.
+		 *
+		 * @since x.x.x
+		 */
+		public function init_marketing_notifications() {
+			$notifications = new Pinterest\Notes\MarketingNotifications();
```
JPry:
```suggestion
			$notifications = new MarketingNotifications();
```

src/Utilities/Utilities.php
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Set the account connection timestamp.
+ *
+ * @since x.x.x
+ * @throws Error Wrong value provided for the timestamp.
+ * @param int $timestamp Timestamp to set as the account connection moment.
+ * @return int
+ */
+function set_account_connection_timestamp( $timestamp = null ): int {
```
JPry: Just curious, why isn't this a class instead of individual functions? We're already using `static` everywhere in this plugin, so this wouldn't be any different than the other classes.

src/Utilities/Utilities.php
```diff
@@ -0,0 +1,53 @@
+<?php
+/**
+ * Pinterest for WooCommerce Feed Logger
+ *
+ * @package     Pinterest_For_WooCommerce/Classes/
+ * @since       1.0.10
```
JPry: This should be the placeholder version.

```suggestion
 * @since       x.x.x
```

src/Utilities/Utilities.php
```diff
@@ -0,0 +1,53 @@
...
+ * @return int Initialization timestamp.
+ */
+function get_account_connection_timestamp(): int {
+	$timestamp = get_option( PINTEREST_ACCOUNT_CONNECTION_TIMESTAMP );
+	if ( false !== $timestamp ) {
+		return (int) $timestamp;
+	}
+
+	$timestamp = set_account_connection_timestamp();
+	return $timestamp;
```
JPry: The `$timestamp` variable is unnecessary here.

```suggestion
	return set_account_connection_timestamp();
```

src/Utilities/Utilities.php
```diff
@@ -0,0 +1,53 @@
...
+		throw new Error( 'Account connection timestamp needs to be an integer value.' );
+	}
+
+	add_option( PINTEREST_ACCOUNT_CONNECTION_TIMESTAMP, $timestamp );
+	return $timestamp;
+}
+
+/**
+ * Gets the account connection timestamp.
+ * Initialize if not.
```
JPry:
```suggestion
 * Gets the account connection timestamp.
 *
 * Initialize if not.
```

src/Notes/Collection/CatalogSyncErrors.php
```diff
@@ -0,0 +1,119 @@
...
+				return false;
+			}
+			switch ( $workflow->workflow_status ) {
+				case 'COMPLETED':
+				case 'COMPLETED_EARLY':
+				case 'PROCESSING':
+				case 'UNDER_REVIEW':
+				case 'QUEUED_FOR_PROCESSING':
+					return false;
+					break;
```
JPry: `break` is redundant here.

```suggestion
					return false;
```

src/Notes/Collection/CatalogSyncErrors.php
```diff
@@ -0,0 +1,119 @@
...
+
+				case 'FAILED':
+				default:
+					return true;
+					break;
```
JPry:
```suggestion
					return true;
```

src/Notes/Collection/CatalogSyncErrors.php
```diff
@@ -0,0 +1,119 @@
...
+			}
+		} catch ( Throwable $th ) {
+			/*
+			 *	Whatever failed we don't care about it in this process.
+			 */
+			return false;
+		}
+
+		// We should never get here.
+		return true;
```
JPry: It's correct we can never get here, so I don't think we need anything after the `catch` block.

```suggestion
		}
```

src/Notes/Collection/CatalogSyncErrors.php
```diff
@@ -0,0 +1,119 @@
...
+	 */
+	protected function get_note_content(): string {
+		return __( 'Your product sync to Pinterest was unsuccessful. To complete your connection, Review and resolve issues in the extension.', 'pinterest-for-woocommerce' );
+	}
+
+	/**
+	 * Add button to Pinterest For WooCommerce landing page
+	 *
+	 * @since x.x.x
+	 * @param Note $note Note to which we add an action.
```
JPry, line 104: For compatibility with IDEs, this needs a `use` statement at the top of the file.

src/Notes/Collection/AbstractNote.php
```diff
@@ -0,0 +1,155 @@
...
+ */
+abstract class AbstractNote {
+
+	/**
+	 * Check if the note has been previously added.
+	 *
+	 * @throws NotesUnavailableException Throws exception when notes are unavailable.
+	 */
+	public static function note_exists() {
+		$data_store = Notes::load_data_store();
```
JPry, line 44: To help IDEs, we should have the following comment added here:

```suggestion
		/** @var DataStore $data_store */
		$data_store = Notes::load_data_store();
```

And the appropriate `use` statement at the top of the file:

```php
use Automattic\WooCommerce\Admin\Notes\DataStore;
```

src/Notes/Collection/AbstractNote.php
```diff
@@ -0,0 +1,155 @@
...
+ * class in the context of Pinterest For WooCommerce plugin.
+ *
+ * @since x.x.x
+ */
+abstract class AbstractNote {
+
+	/**
+	 * Check if the note has been previously added.
+	 *
+	 * @throws NotesUnavailableException Throws exception when notes are unavailable.
```
JPry, line 35: IDEs don't recognize this without the import at the top of the file:

```php
use Automattic\WooCommerce\Admin\Notes\NotesUnavailableException;
```

src/Notes/Collection/AbstractNote.php
```diff
@@ -0,0 +1,155 @@
...
+
+	/**
+	 * Add action to note if necessary.
+	 *
+	 * @since x.x.x
+	 * @param Note $note Note to which we want to add an action.
+	 */
+	protected function add_action( $note ): void {
+		return;
+	}
```
JPry: Since we don't need to return anything at all, I think it would be fine to put the empty body in one line:

```suggestion
	protected function add_action( $note ): void {}
```

src/Notes/Collection/AbstractNote.php
```diff
@@ -0,0 +1,155 @@
...
+	 */
+	private function get_slug(): string {
+		return PINTEREST_FOR_WOOCOMMERCE_PREFIX;
+	}
+
+	/**
+	 * Get note title.
+	 *
+	 * @since x.x.x
+	 * @return strings Note title.
```
JPry: Typo

```suggestion
	 * @return string Note title.
```

JPry (changes requested):
@budzanowski I got partway through reviewing this today and wanted to get the feedback I already created over to you. I can pick up and finish the review tomorrow.

src/Utilities/Utilities.php
budzanowski: The intention is to have utils as a bag for all sorts of things. Cramming them all into a class is OK, but I thought that having them separate makes a bit more sense. This is a personal preference based on how I tend to think about how methods and data in class should relate to each other. I am happy to change this if you think that this is not good.

src/Notes/Collection/AbstractNote.php
budzanowski: Correct.

src/Notes/Collection/AbstractNote.php
budzanowski: Cool.

src/Notes/Collection/AbstractNote.php
budzanowski: 👍

src/Notes/Collection/AbstractNote.php
budzanowski: 👍

budzanowski: @JPry It forced me to add a short description to the docblock:
```PHP
	        /**
		 * Notes Data Store.
		 *
		 * @var DataStore $data_store
		 */
```

Is there a less verbose format for type-hinting in the IDE?

src/Notes/Collection/CatalogSyncErrors.php
budzanowski: 👍

budzanowski:
@JPry thank you for the initial review. All things addressed.

You are always correcting my IDE type hinting shortcomings. That is great! ❤️
Is this a feature of your IDE setup that makes this quickly visible? Is something that you particularly spot easily on your own. Or not easily but just pay close attention to ( which I should be doing apparently ).

mikkamp:
Don't want to take over the PR review but I had a question about the Heartbeat class. It's a nice solution but isn't it a bit overkill? In other places in the plugin we just use a call to schedule a recurring action with [as_schedule_recurring_action](https://actionscheduler.org/api/#function-reference--as_schedule_recurring_action). We do the same in [GLA](https://github.com/woocommerce/google-listings-and-ads/blob/1.12.4/src/Notes/NoteInitializer.php#L103) (albeit through an interface).

Also if Utilities is always going to be loading on plugin initialization then we can use composer to [autoload it](https://getcomposer.org/doc/04-schema.md#files). But wouldn't converting it to a class give us the advantage that it's only loaded when needed?

budzanowski:
@mikkamp that is a fair question. The Heartbeat class is a mental model. It allows different parts of the plugin to subscribe to the same recurrent action without the necessity to create separate daily actions. Hence the name Heartbeat. The class easily can be extended to add an hourly heartbeat which I envision to be useful in the next iterations of the plugin. It centralizes and allows synchronization of various cyclic processes in the plugin. In the future, I want to move registration logic that is called often to a Heartbeat event.
 
So the class is just a model of behavior. The implementation using Crone is just for performance reasons. The gain is small initially but with more subscribers, the performance is getting better.

Facebook For WooCommerce is using the same approach. https://github.com/woocommerce/facebook-for-woocommerce/blob/master/includes/Utilities/Heartbeat.php

What do you think?

> Also if Utilities is always going to be loading on plugin initialization then we can use composer to [autoload it](https://getcomposer.org/doc/04-schema.md#files). But wouldn't converting it to a class gives us the advantage that it's only loaded when needed?

I will turn it into a class bc I can see from the review that it is a confusing implementation 👍🏻

mikkamp:
> The Heartbeat class is a mental model. It allows different parts of the plugin to subscribe to the same recurrent action without the necessity to create separate daily actions. Hence the name Heartbeat. The class easily can be extended to add an hourly heartbeat which I envision to be useful in the next iterations of the plugin. It centralizes and allows synchronization of various cyclic processes in the plugin. In the future, I want to move registration logic that is called often to a Heartbeat event.

Thanks for the clarification, that makes a lot of sense. Although I suppose the part I have the most issues with is using WP Cron for triggering the Heartbeat. Action Scheduler does a better job at registering a queue runner both from cron as well as from admin requests (and has worked through lots of issues to ensure it's run), so in my opinion it would be nicer to take advantage of that.

The class mentions you use cron instead of AS because:
> WP Cron events are stored in an auto-loaded option so the performance impact is much lower than checking and scheduling an Action Scheduler action.

But couldn't we handle that by only running that code on plugin activation/update, and clearing on plugin deactivation? Otherwise we are using a recurring cron action to trigger a single scheduled action (on a daily basis), sounds simpler if we could just setup a recurring scheduled action or even use [as_schedule_cron_action](https://actionscheduler.org/api/#function-reference--as_schedule_cron_action) so we can influence time of day when it's run.

src/Notes/Collection/AbstractNote.php
JPry: @budzanowski In other codebases, we specifically exclude that PHPCS sniff that throws an error for that comment format. E.g. [here](https://github.com/woocommerce/google-listings-and-ads/blob/develop/phpcs.xml.dist#L13) in Google Listings & Ads:

```xml
		<exclude name="Generic.Commenting.DocComment.MissingShort"/>
```

That does eliminate the warning for a short description in other comment variations, but I think the benefit for these inline comments outweighs possibly missing shot descriptions elsewhere.

JPry: @budzanowski It looks like that `<exclude>` tag would go in this block:

https://github.com/woocommerce/pinterest-for-woocommerce/blob/780345abf527d4c36bbd31bcdece343b3d375347/phpcs.xml#L20-L22

budzanowski:
> But couldn't we handle that by only running that code on plugin activation/update, and clearing on plugin deactivation?

We can but IMO this makes it more complicated. We need to provide handling, as you mention, for activation, update, and deactivation. Otherwise, we would need to use a similar approach as now with AS using `as_has_scheduled_action` which compared to checking cron is a much more expensive operation. 

I don't mind changing it though if this is something you feel strongly about - I do not so please do not hesitate to request a change.

JPry:
@budzanowski 

> Is this a feature of your IDE setup that makes this quickly visible? 

Yes, it's a feature of PhpStorm. Issues are identified visually, so it's very easy to skim for these while also reviewing the code as a whole. Here's an example showing first what the IDE looks like, and then what information I see by hovering my cursor over the line:

Screen Shot 2022-04-07 at 3 26 23 PM

Screen Shot 2022-04-07 at 3 26 36 PM

JPry:
Regarding the discussion about Action Scheduler vs. WP Cron, I agree with what @mikkamp said:

> Action Scheduler does a better job at registering a queue runner both from cron as well as from admin requests (and has worked through lots of issues to ensure it's run), so in my opinion it would be nicer to take advantage of that.

But I'm not familiar with what @budzanowski mentioned:

> Otherwise, we would need to use a similar approach as now with AS using `as_has_scheduled_action` which compared to checking cron is a much more expensive operation.

Could you clarify with some more details, or else point me to some more discussion around why Action Scheduler is a more expensive operation? I feel that's the only fair way to weigh the merits of each approach.

class-pinterest-for-woocommerce.php
```diff
@@ -226,6 +228,10 @@ public function init_plugin() {
 
 			$this->includes();
 
+			// Start the heartbeat.
+			$this->heartbeat = new Heartbeat( WC()->queue() );
```
JPry, line 240: Looks like this property isn't declared at the top of the class. Could you please add that declaration?

budzanowski:
@JPry this

> Could you clarify with some more details, or else point me to some more discussion around why Action Scheduler is a more expensive operation? I feel that's the only fair way to weigh the merits of each approach.

This is an implementation detail. If you want to check if CRON action exists you use `wp_next_scheduled` function. This function checks if your CRON action exists by basically checking what is inside the autoloaded option fetched using `get_option( 'cron' );` So something quite cheap from the perspective of code execution.

If you want to check if an Action Scheduler action is already set up you need to use `as_has_scheduled_action` which in order to figure out if the action exists will perform an actual DB query through `wpdb` . 

If you have multiple AS actions that you maintain throughout the plugin you need multiple DB calls each `init` or `admin_init` if you check each time if the action is set and set it if not. Or you need activate/update/deactivate code to handle your action. Which is OK to have and if you are careful it does not cause a lot of issues ( except the occasional double actions ). But IMO this is actually more complex compared to the Heartbeat Cron->AS as there is a single entry point that helps to reduce the complexity.

src/Notes/Collection/CatalogSyncErrors.php
```diff
@@ -0,0 +1,116 @@
...
+ */
+class CatalogSyncErrors extends AbstractNote {
+
+	const NOTE_NAME = 'pinterest-catalog-sync-error';
+
+	/**
+	 * Should the note be added to the inbox.
+	 *
+	 * @since x.x.x
+	 * @return boolean
```
JPry: Minor: `bool` instead of `boolean`

```suggestion
	 * @return bool
```

src/Notes/Collection/CatalogSyncErrors.php
```diff
@@ -0,0 +1,116 @@
...
+					return false;
+
+				case 'FAILED':
+				default:
+					return true;
+			}
+		} catch ( Throwable $th ) {
+			/*
+			 *	Whatever failed we don't care about it in this process.
+			 */
```
JPry: Minor: This comment doesn't need multiple lines

```suggestion
			// Whatever failed we don't care about it in this process.
```

src/Notes/Collection/CompleteOnboardingAfterThreeDays.php
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Class CompleteOnboardingAfterThreeDays.
+ *
+ * Class responsible for admin Inbox notification after three days from setup.
+ *
+ * @since x.x.x
+ */
+class CompleteOnboardingAfterThreeDays extends AbstractCompleteOnboarding {
+
+	const DELAY     = 0;
```
JPry: Should this be `3` since this is the `CompleteOnboardingAfterThreeDays` class?

src/Notes/Collection/CompleteOnboardingReminderAfterFourteenDays.php
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Class CompleteOnboardingAfterFourteenDays.
+ *
+ * Class responsible for admin Inbox notification after fourteen days from setup.
+ *
+ * @since x.x.x
+ */
+class CompleteOnboardingReminderAfterFourteenDays extends CompleteOnboardingReminderAfterSevenDays {
+
+	const DELAY = 14;
```
JPry:
```suggestion
	const DELAY     = 14;
```

src/Notes/Collection/CompleteOnboardingReminderAfterSevenDays.php
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Class CompleteOnboardingAfterSevenDays.
+ *
+ * Class responsible for admin Inbox notification after seven days from setup.
+ *
+ * @since x.x.x
+ */
+class CompleteOnboardingReminderAfterSevenDays extends AbstractCompleteOnboarding {
+
+	const DELAY = 7;
```
JPry:
```suggestion
	const DELAY     = 7;
```

src/Notes/Collection/CompleteOnboardingReminderAfterThirtyDays.php
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Class CompleteOnboardingAfterThirtyDays.
+ *
+ * Class responsible for admin Inbox notification after thirty days from setup.
+ *
+ * @since x.x.x
+ */
+class CompleteOnboardingReminderAfterThirtyDays extends CompleteOnboardingReminderAfterSevenDays {
+
+	const DELAY = 30;
```
JPry:
```suggestion
	const DELAY     = 30;
```

src/Notes/Collection/EnableCatalogSync.php
```diff
@@ -0,0 +1,88 @@
...
+ */
+class EnableCatalogSync extends AbstractNote {
+
+	const NOTE_NAME = 'pinterest-enable-catalog-sync';
+
+	/**
+	 * Should the note be added to the inbox.
+	 *
+	 * @since x.x.x
+	 * @return boolean
```
JPry:
```suggestion
	 * @return bool
```

src/Notes/Collection/EnableCatalogSync.php
```diff
@@ -0,0 +1,88 @@
...
+	 */
+	protected function get_note_content(): string {
+		return __( 'Your Catalog sync with Pinterest has been disabled. Select “Enable Product Sync” to sync your products and reach shoppers on Pinterest.', 'pinterest-for-woocommerce' );
+	}
+
+	/**
+	 * Add button to Pinterest For WooCommerce landing page
+	 *
+	 * @since x.x.x
+	 * @param Note $note Note to which we add an action.
```
JPry, line 79: This needs an import at the top of the file for the IDE to recognize it.

```php
use Automattic\WooCommerce\Admin\Notes\Note;
```

src/Notes/MarketingNotifications.php
```diff
@@ -0,0 +1,82 @@
...
+
+	// All options common prefix.
+	const OPTIONS_PREFIX = PINTEREST_FOR_WOOCOMMERCE_PREFIX . '-marketing-notifications';
+
+	// Timestamp option marking the moment we start to count time.
+	const INIT_TIMESTAMP = self::OPTIONS_PREFIX . '-init-timestamp';
+
+	// List of marketing notifications that we want to send.
+	const NOTES = array(
+		'EnableCatalogSync',
```
JPry: For these items, it would be preferable to use the `::class` constant to reference each one. That would give us two benefits:

1. IDEs will have an easier time determining where each class is actually used.
2. More importantly, in the `init_notifications()` method below, we don't need to manually prepend the proper namespace to each item, but can immediately just call `new $note()`

src/Notes/MarketingNotifications.php
```diff
@@ -0,0 +1,82 @@
...
+	 * Trigger inbox messages.
+	 *
+	 * @since x.x.x
+	 * @return void
+	 */
+	public function init_notifications(): void {
+
+		foreach ( self::NOTES as $note ) {
+			$note         = 'Automattic\WooCommerce\Pinterest\Notes\Collection\\' . $note;
+			$notification = new $note();
```
JPry: This change should be coupled with the explanation in the comment above regarding the use of `::class`. We can also type-hint against our `AbstractNote` class.

```suggestion
			/** @var AbstractNote $notification */
			$notification = new $note();
```

src/Utilities/Utilities.php
```diff
@@ -0,0 +1,60 @@
...
+	 * @throws Error Wrong value provided for the timestamp.
+	 * @param int $timestamp Timestamp to set as the account connection moment.
+	 * @return int
+	 */
+	public static function set_account_connection_timestamp( $timestamp = null ): int {
+		if ( is_null( $timestamp ) ) {
+			$timestamp = time();
+		}
+
+		if ( ! is_int( $timestamp ) ) {
```
JPry: Do you think there's ever a time where we might want to specify the timestamp instead of it just being the current moment? Right now when this is called, the timestamp is not specified. This makes me think that maybe we don't need the `$timestamp` parameter at all and we can just call `time()` directly. That would also have the additional benefit of not needing these checks of the passed parameter.

If you do feel that keeping the parameter is necessary, then I think it would make sense to declare it as a nullable type, e.g.:

```php
	public static function set_account_connection_timestamp( ?int $timestamp = null ): int {
```

Then we only need to verify if it's `null`, and don't need to otherwise type-check the parameter.

src/Utilities/Utilities.php
```diff
@@ -0,0 +1,60 @@
...
+	/**
+	 * Set the account connection timestamp.
+	 *
+	 * @since x.x.x
+	 * @throws Error Wrong value provided for the timestamp.
+	 * @param int $timestamp Timestamp to set as the account connection moment.
+	 * @return int
+	 */
+	public static function set_account_connection_timestamp( $timestamp = null ): int {
+		if ( is_null( $timestamp ) ) {
```
JPry: If we end up keeping this (based on the comment below), then this should compare directly to `null` instead of using the `is_null()` function.

```suggestion
		if ( null === $timestamp ) {
```

src/Utilities/Utilities.php
```diff
@@ -0,0 +1,60 @@
...
+	 * @since x.x.x
+	 * @return int Initialization timestamp.
+	 */
+	public static function get_account_connection_timestamp(): int {
+		$timestamp = get_option( self::PINTEREST_ACCOUNT_CONNECTION_TIMESTAMP );
+		if ( false !== $timestamp ) {
+			return (int) $timestamp;
+		}
+
+		return self::set_account_connection_timestamp();
```
JPry: I'm not so sure how I feel about automatically setting the connection timestamp from this method. It seems a bit counter-intuitive to call **get* timestamp, and then have it set as a result.

JPry (changes requested):

JPry:
@budzanowski 

> If you have multiple AS actions that you maintain throughout the plugin you need multiple DB calls each `init` or `admin_init` if you check each time if the action is set and set it if not. Or you need activate/update/deactivate code to handle your action. Which is OK to have and if you are careful it does not cause a lot of issues ( except the occasional double actions ). But IMO this is actually more complex compared to the Heartbeat Cron->AS as there is a single entry point that helps to reduce the complexity.

I see what you're saying, but I'm not sure the performance is as much of an issue as you think. For one thing, even though it's true that the cron array is autoloaded anyway, it's also a single option stored in the `options` table. We have no control over how many (or how few) other plugins are making use of the native cron, but I think it is advisable to avoid adding to the size of that option.

On the other hand, Action Scheduler defaults to using its own database tables, and querying for a scheduled action is hitting an indexed column in the table. This means that the overhead of querying for an existing action is minimal, even with multiple queries going on during a request.

src/Notes/Collection/CompleteOnboardingAfterThreeDays.php
budzanowski: Oh no! I did change it for testing purposes and forgot about this.

src/Notes/MarketingNotifications.php
budzanowski: Very cool!

mikkamp:
> I don't mind changing it though if this is something you feel strongly about - I do not so please do not hesitate to request a change.

I don't have strong feelings about it at all. My thought pattern is mainly that there has been so much work done in Action Scheduler for improving the lacks of WP Cron. So why not take advantage of it at every opportunity we can.

src/Utilities/Utilities.php
```diff
@@ -0,0 +1,60 @@
...
+	public static function set_account_connection_timestamp( $timestamp = null ): int {
+		if ( is_null( $timestamp ) ) {
+			$timestamp = time();
+		}
+
+		if ( ! is_int( $timestamp ) ) {
+			throw new Error( 'Account connection timestamp needs to be an integer value.' );
+		}
+
+		add_option( self::PINTEREST_ACCOUNT_CONNECTION_TIMESTAMP, $timestamp );
```
mikkamp: I think the intention was probably to use `update_option` based on this comment:
> If the user changes tag settings this will get reset which makes sense because the connection parameters have been changed. ofc this will also get reset on a full reconnection.

src/Notes/MarketingNotifications.php
```diff
@@ -0,0 +1,89 @@
...
+ * Marketing notifications for merchants (admin).
+ *
+ * @package Pinterest_For_WooCommerce/Classes/
+ * @since   x.x.x
+ */
+
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\Pinterest\Notes;
+use Automattic\WooCommerce\Pinterest\Notes\Collection\EnableCatalogSync;
```
JPry: Coupled with fixing the PHPCS error here, we should also add one more import.

```suggestion
namespace Automattic\WooCommerce\Pinterest\Notes;

use Automattic\WooCommerce\Pinterest\Notes\Collection\AbstractNote;
use Automattic\WooCommerce\Pinterest\Notes\Collection\EnableCatalogSync;
```

src/Utilities/Utilities.php
budzanowski: Removing the param. We can always add it later if it will turn out to be necessary.

src/Utilities/Utilities.php
budzanowski: Feature removed.

src/Utilities/Utilities.php
budzanowski: Changed. Account connection timestamp is not deliberately set by MarketingNotifications class.

budzanowski:
@JPry @mikkamp 
The Heartbeat is AS only now. Thanks for pushing me on this.

JPry (approved):

src/Notes/Collection/AbstractCompleteOnboarding.php
```diff
@@ -0,0 +1,97 @@
...
+		if ( false !== $timestamp ) {
+			// Timestamp already init, return.
+			return (int) $timestamp;
+		}
+
+		// Timestamp not initialized, create a new one.
+		$timestamp = time();
+		add_option( self::INIT_TIMESTAMP, $timestamp );
+		return $timestamp;
+	}
```
mikkamp: I know this is very opinionated, but my thought was that this function would be clearer if we initialize within the if statement and have only one return. Which in my opinion makes the extra comment lines unnecessary since we can describe what it does in the function comment.
```suggestion
	protected function get_init_timestamp(): int {
		$timestamp = get_option( self::INIT_TIMESTAMP );
		if ( false === $timestamp ) {
			$timestamp = time();
			add_option( self::INIT_TIMESTAMP, $timestamp );
		}

		return (int) $timestamp;
	}
```

src/Notes/Collection/AbstractCompleteOnboarding.php
```diff
@@ -0,0 +1,97 @@
...
+		$note->add_action(
+			'goto-pinterest-landing',
+			__( 'Complete setup', 'pinterest-for-woocommerce' ),
+			wc_admin_url( '&path=/pinterest/landing' )
+		);
+	}
+
+	/**
+	 * Check if the init timestamp exist.
+	 * Initialize if not.
```
mikkamp: I think this should mention it retrieves the timestamp as it's main function. The secondary function is to intialize if it hasn't been done before.

src/Utilities/Utilities.php
```diff
@@ -0,0 +1,46 @@
...
+		$timestamp = time();
+
+		add_option( self::PINTEREST_ACCOUNT_CONNECTION_TIMESTAMP, $timestamp );
+
+		return $timestamp;
+	}
+
+	/**
+	 * Gets the account connection timestamp.
+	 * Initialize if not.
```
mikkamp: I think this line is no longer the case, it's already mentioned in the return but if we want to mention it again, I think it should specify it returns 0 as it doesn't really initialize anything.

src/Utilities/Utilities.php
```diff
@@ -0,0 +1,46 @@
...
+	 *
+	 * @since x.x.x
+	 * @return int Initialization timestamp. Zero if not set.
```
mikkamp: Isn't this the "account connection timestamp"?

src/Notes/Collection/AbstractCompleteOnboarding.php
```diff
@@ -0,0 +1,97 @@
...
+		if ( Pinterest_For_Woocommerce()::is_setup_complete() ) {
+			return false;
+		}
+
+		if ( self::note_exists() ) {
+			return false;
+		}
+
+		// Are we there yet?
+		if ( time() < ( DAY_IN_SECONDS * static::DELAY + $this->get_init_timestamp() ) ) {
```
mikkamp: Since we are using this "constant" in an abstract class we are forcing every class that extends this to declare it. Otherwise we could get `Undefined constant ExtendedClass::DELAY`. Same for the NOTE_NAME constant.

Any solutions to check they are implemented in the class would seem like overkill, but do we have a simpler way like a comment in the abstract class to identify what is required?

mikkamp (approved):
I mainly looked over the last few commits. In general the code looks good, I mainly left a few opinionated comments, but none of them are blocking.

JPry:
@mikkamp Thanks for the review! Based on your feedback regarding the `DELAY` and `NOTE_NAME` constants, I decided to do a bit of reworking the notes. I also removed the commit where I moved the method to the Abstract note class. This should be ready for another look.

src/Notes/MarketingNotifications.php
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	/**
+	 * Get the notification init timestamp.
+	 *
+	 * @since x.x.x
+	 * @return int Initialization timestamp.
+	 */
+	protected function get_init_timestamp(): int {
+		return get_option( self::INIT_TIMESTAMP, 0 );
```
mikkamp: Wouldn't it be safer to cast it to an int here. The docs for [`get_option`](https://developer.wordpress.org/reference/functions/get_option/) mentions (most of the time) any scalar value is returned as a string. The default value is only returned if the key is not found, but if it's found with an "empty" value it still returns a string.

src/Notes/MarketingNotifications.php
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	/**
+	 * Set the notification init timestamp to the current time.
+	 *
+	 * @since x.x.x
+	 * @return void
+	 */
+	public function set_init_timestamp(): void {
+		add_option( self::INIT_TIMESTAMP, time() );
```
mikkamp: It looks like we only want to ever set this once, but what if it happens to be set with an empty value, it will never be updated unless we use update_option. Since we explicitly only set this if it's 0 wouldn't it just be safer to use update_option?

src/Utilities/Utilities.php
```diff
@@ -0,0 +1,40 @@
...
+
+	const PINTEREST_ACCOUNT_CONNECTION_TIMESTAMP = PINTEREST_FOR_WOOCOMMERCE_PREFIX . '-account-connection-timestamp';
+
+	/**
+	 * Set the account connection timestamp.
+	 *
+	 * @since x.x.x
+	 */
+	public static function set_account_connection_timestamp() {
+		add_option( self::PINTEREST_ACCOUNT_CONNECTION_TIMESTAMP, time() );
```
mikkamp: Here also, wouldn't it be better to use update_option? I don't see the value every getting reset if we disconnect, and since this function will only be triggered if it detects a 0, it seems it would always keep the timestamp of the first time we connected. Is that correct?

mikkamp (commented):
Those changes look good. It just made me curious why NOTE_NAME uses a constant and DELAY got moved to a function that can be overwritten. Either implementation works for me, as long as the abstract class has a default.

I just left some additional comments about the peculiarities of the `get_option` and `add_option` functions.

src/Notes/MarketingNotifications.php
JPry: Good catch!

JPry:
@mikkamp Thanks for the feedback. I added a few more tweaks based on your suggestions, and also ensured the options will be deleted when the plugin is uninstalled (as long as the "erase all data" option is set).

JPry:
@mikkamp 

> It just made me curious why NOTE_NAME uses a constant and DELAY got moved to a function that can be overwritten.

I did it this way because the `NOTE_NAME` constant is used in `AbstractNote::note_exists()`. I thought it would be easier to keep a constant here.

PR #428, opened 2022-04-01
Include all tracking properties with JS tracking

JPry:
### Changes proposed in this Pull Request:

Closes #400.

This updates the tracking code triggered by the "Add to cart" button to include the same data that is provided from the regular product page.

### Detailed test instructions:

1. Install the plugin and do the onboarding with a merchant.
2. Make sure that tracking is properly configured.
3. From the shop page, click the "Add to cart" button for a product. In the network tab of your browser dev tools, you should see the data in the tracking event named `AddToCart` includes all of the correct items: `product_id`, `product_name`, `value`, `order_quantity`, and `currency`.
4. Navigate to an individual product page, and add the product to the cart
5. Check the event and see that all of the same parameters are included in the `AddToCart` event

### Screenshots

Screen Shot 2022-04-05 at 6 07 08 PM

### Changelog entry

> Fix - Ensure Add to cart tag data is consistent

PR #437, opened 2022-04-05
Send currency in product page visit event

### LINKED ISSUE #421: Currency should be inluded as an event property when price/value are sent
### Describe the bug:

When price/value properties are sent to Pinterest in events the currency property should also be used.

Currently, the product page view event does not send the currency e.g.

https://github.com/woocommerce/pinterest-for-woocommerce/blob/develop/src/Tracking.php#L363-L374

All events should be double check when fixing this issue.

Example of the AddToCart event which DOES sends currency

https://github.com/woocommerce/pinterest-for-woocommerce/blob/develop/src/Tracking.php#L224-L233

Example of Checkout event which DOES sends currency

https://github.com/woocommerce/pinterest-for-woocommerce/blob/develop/src/Tracking.php#L277-L286

### Expected behavior:
Currency is included

### Actual behavior:
Currency only included in AddToCart and Checkout events.

alexiglesias31:

Closes #421.

Double check all events to send the currency in all events.
- Added currency on product page visit event.
- Verified: currency sent on checkout event.
- Verified: currency sent on add to cart event.

- Missing: currency on add to cart event using AJAX.
NOTE: The PR #428 from @JPry adds the currency in the AJAX event, so we don't need to add it here.

1. Install the plugin and do the onboarding with a merchant.
2. Make sure that tracking is properly configured.
3. Visit a product page and check the `pagevisit` event, currency should be sent along with the other product parameters.
4. Verify that currency is also sent on addToCart (not AJAX) event and checkout event.

### Additional details:

> Fix - Currency is now being sent on the product page visit event.

### DIFF
--- a/src/Tracking.php
+++ b/src/Tracking.php
@@ -368,6 +368,7 @@ class Tracking {
 				'product_id'    => $product->get_id(),
 				'product_name'  => $product->get_name(),
 				'product_price' => $product->get_price(),
+				'currency'      => get_woocommerce_currency(),
 			);
 		}
 


alexiglesias31:
@jconroy.
As part of the PR we double checked the currency field across all the events. The details are in the PR description.

findingsimple:
Thanks @alexiglesias31 !

mikkamp (approved):
The code change looks good. I've tested all events (except AJAX add to cart) and I can consistently see the currency added.

It's added now for the pagevisit event where it previously wasn't:
```
event: pagevisit
ed: {"product_id":72,"product_name":"Test Product","product_price":"10","currency":"EUR"}
```

PR #445, opened 2022-04-11
Add additional images limit on feed

### LINKED ISSUE #382: Value for additional_image_link to big - Pinterest only supports 10 additional images
### Describe the bug:
details: https://wordpress.org/support/topic/value-for-additional_image_link-to-big

### Steps to reproduce:

1. Add more than 10 aditional images to a product
2. install and setup Pinterest plugin
3. XML is ingested to Pinterest and shows the error.

alexiglesias31:

Closes #382.

Pinterest has a limit of 10 additional images per product in the feed. This PR adds that limit on the feed generation.

1. Add more than 10 additional images to a product.
2. Install and setup the Pinterest plugin.
3. The generated feed should contain only 10 additional images.
4. There should be no error on the Pinterest ingestion.

### Changelog entry

> Fix - Limit the number of additional images to 10

### DIFF
--- a/src/ProductsXmlFeed.php
+++ b/src/ProductsXmlFeed.php
@@ -60,6 +60,13 @@ class ProductsXmlFeed {
 	 */
 	const DESCRIPTION_SIZE_CHARS_LIMIT = 10000;
 
+	/**
+	 * Limit of additional images allowed by Pinterest.
+	 *
+	 * @var int
+	 */
+	const ADDITIONAL_IMAGES_LIMIT = 10;
+
 
 	/**
 	 * Returns the XML header to be printed.
@@ -444,7 +451,7 @@ class ProductsXmlFeed {
 			return;
 		}
 
-		return '<' . $property . '><![CDATA[' . implode( ',', $images ) . ']]></' . $property . '>';
+		return '<' . $property . '><![CDATA[' . implode( ',', array_slice( $images, 0, self::ADDITIONAL_IMAGES_LIMIT ) ) . ']]></' . $property . '>';
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


ksere (approved):
LGTM

alexiglesias31:
@jconroy @JPry this PR is ready for review.

src/ProductsXmlFeed.php
```diff
@@ -444,7 +451,7 @@ private static function get_property_g_additional_image_link( $product, $propert
 			return;
 		}
 
-		return '<' . $property . '><![CDATA[' . implode( ',', $images ) . ']]></' . $property . '>';
+		return '<' . $property . '><![CDATA[' . implode( ',', array_slice( $images, 0, self::ADDITIONAL_IMAGES_LIMIT ) ) . ']]></' . $property . '>';
```
JPry: Minor: the `array_slice()` call should be applied to `$images` before the output string is returned. Alternatively, this would be more readable using `sprintf()` as a wrapper to generate this element.

JPry (commented):
I think we can tweak where the `array_slice()` call happens for better readability, but this otherwise looks good.

src/ProductsXmlFeed.php
alexiglesias31: Thanks @JPry. Already pushed the changes for applying `array_slice()` and `implode()` before the output string.
About the `sprintf()`, after the changes the readability is much better, also the rest of returned stings follow the same pattern.
Wdyt? Should we go for all of them?

JPry: > About the `sprintf()`, after the changes the readability is much better, also the rest of returned stings follow the same pattern.
> Wdyt? Should we go for all of them?

I think we can keep the consistency in this PR, and a follow-up PR later can convert to `sprintf()` format.

alexiglesias31:
@JPry , will merge this one and link it later in the follow-up PR.

PR #447, opened 2022-04-14
Add action scheduler to min requirements

### LINKED ISSUE #443: Possible fatal error if action scheduler is missing
### Describe the bug:
The following bug appeared in a customer site:
Fatal error with the following plugin versions:
- Pinterest for WooCommerce (1.0.10)
- WooCommerce (6.3.1)

This shouldn't happen if WooCommerce is properly installed. It has been reported in only one site, but probably it would be nice to add an extra verification on `ProductSync` class to check if action scheduler is defined, just as we are doing in the deactivation hook:

https://github.com/woocommerce/pinterest-for-woocommerce/blob/8838bab30411e884a799c7c1535fdaa559e7fb4e/uninstall.php#L22-L26

Also, probably we need to add a notice to the user informing that action scheduler is not found, so the client is aware that feed generation is not being triggered.

alexiglesias31:

Closes #443.

The plugin requires a version of Action Scheduler package >= 3.3.0. If for some reason, a plugin loads an older version of the package, the plugin will throw fatal error.

The changes on this PR include:
- Add the action scheduler v3.3.0 as a min requirement.

### Screenshots:

image

1. Install the Pinterest for WooCommerce plugin along with WooCommerce.
2. Install a Woo extension that uses an old version of the action scheduler package, such as WooCommerce Smart Coupons v4.8.0.
3. Manually edit the file woocommerce -> `class-woocommerce` and remove the action scheduler loader (forcing the action scheduler on Smart Coupons plugin).
4. The plugin should throw an error since action scheduler min requirement is not fulfilled (check screenshot).

### Changelog entry

> Add - Add action scheduler as a minimum plugin requirement.

### DIFF
--- a/class-pinterest-for-woocommerce.php
+++ b/class-pinterest-for-woocommerce.php
@@ -49,9 +49,10 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 		 * Set the minimum required versions for the plugin.
 		 */
 		const PLUGIN_REQUIREMENTS = array(
-			'php_version' => '7.3',
-			'wp_version'  => '5.6',
-			'wc_version'  => '5.3',
+			'php_version'      => '7.3',
+			'wp_version'       => '5.6',
+			'wc_version'       => '5.3',
+			'action_scheduler' => '3.3.0',
 		);
 
 		/**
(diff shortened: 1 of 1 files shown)


ksere (approved):

alexiglesias31:
@jconroy @JPry, the PR is ready for review.

JPry (approved):
Looks good to me!

PR #451, opened 2022-04-15
Check if get_pinterest_code mehod exists

### LINKED ISSUE #449: Fatal error due to call to get_pinterest_code() method that doesn't exist
### Describe the bug:

As reported [in this forum post](https://wordpress.org/support/topic/uncaught-error-call-to-undefined-method-exceptionget_pinterest_code/), this section of code calls the `get_pinterest_code()` method on an object which might not have that method available:

https://github.com/woocommerce/pinterest-for-woocommerce/blob/9c5c1b9bc514871993ea1c7d775d3efa98a9efb4/src/FeedRegistration.php#L97-L105

If we are going to call `get_pinterest_code()` on the `$th` object, we need to ensure that the object actually has that method available.

alexiglesias31:

Closes #449.

We need to make sure that `get_pinterest_code()` method exists in the `$th` object.

### Changelog entry

> Fix - Fix fatal error if `get_pinterest_code()` doesn't exists on Throwable object

### DIFF
--- a/src/FeedRegistration.php
+++ b/src/FeedRegistration.php
@@ -95,7 +95,7 @@ class FeedRegistration {
 			throw new Exception( esc_html__( 'Could not register feed.', 'pinterest-for-woocommerce' ) );
 
 		} catch ( Throwable $th ) {
-			if ( 4163 === $th->get_pinterest_code() ) {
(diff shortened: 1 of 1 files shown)


JPry (approved):
Looks good!

ksere (approved):

PR #436, opened 2022-04-05
Sync enhanced match support with Pinterest

alexiglesias31:
### Changes proposed in this Pull Request:

Closes #429
Closes #430 

This PR enables the enhanced match setting and syncs this through the Pinterest API.
It covers the following points:

- Set `aem_enabled` when creating a tag.
- For all users that are updating to 1.0.11 we will update the tag with the same value than the setting.
- When the settings are updated trigger a tag update to sync the option with Pinterest.

### Screenshots:

image

#### Creating a new tag
1. Create a new Pinterest account or get an account which has 0 tags.
2. Connect the account using the plugin.
3. Visit the store to generate any tag event.
4. On Pinterest page -> Ads -> Conversions, Automatic enhanced match should be enabled.

#### Update routine
1. On Pinterest page -> Ads -> Conversions disable/enable the Automatic enhanced match (to make it different than the plugin settings).
2. Temporarily bump the value PINTEREST_FOR_WOOCOMMERCE_VERSION in the main plugin file to 1.0.11.
3. Refresh any admin page.
4. On Pinterest page -> Ads -> Conversions, Automatic enhanced match should be synched with the plugin.

#### Enable/Disable in settings
2. On the Pinterest plugin > Settings page enable/disable the enhanced match and save the settings.
3. On Pinterest page -> Ads -> Conversions, Automatic enhanced match should be synched with the plugin.

PR #452, opened 2022-04-19
Revert "Sync enhanced match support with Pinterest"

JPry:
Reverts woocommerce/pinterest-for-woocommerce#436

Pending further discussion around this feature, we want to revert it for now.

This PR will need to be reverted later when/if we are ready to include the enhanced match support.

### Changelog entry

>


findingsimple:
LGTM

PR #458, opened 2022-05-06
Improve error logs when debug logging is disabled

### LINKED ISSUE #187: Include essential request details in error logs
### Describe the bug:

This issue is slightly related to https://github.com/woocommerce/pinterest-for-woocommerce/issues/186.

When I go to the WooCommerce logs page (`/wp-admin/admin.php?page=wc-status&tab=logs`), I see many "ERROR Error Processing Request: Invalid parameters." with no error details:

image

Without error details (e.g. which request, which parameters are invalid) and stack trace, the logs here are not really helpful to troubleshoot issues. The logs are one of the main things that our customer support (Happiness Engineer) use to help the users.

### Steps to reproduce:

Pre-condition: follow the steps to reproduce in https://github.com/woocommerce/pinterest-for-woocommerce/issues/186. 

1. Go to the WooCommerce logs page (`/wp-admin/admin.php?page=wc-status&tab=logs`)
2. Select Pinterest log file and click on the View button.

#### Expected behavior

- In verbose/debug mode, more details and possibly a stack trace should be displayed. Ideally, for `Invalid parameters`, it should be possible to find out what parameters are invalid 🕵️.
- In normal/production mode, each log should include enough info to be useful, e.g. request details and response code.

#### Actual behavior

Logs contain a raw error message with no context - e.g. `Error Processing Request: Invalid parameters`.

alexiglesias31:

Closes #187.

Improvements on the log:
- Added the stack trace in verbose/debug mode.
- Added essentials request parameters in normal/production mode (request method, endpoint, error code)

### Screenshots:

**Log in normal/production mode:**
image

**Fragment of log in verbose/debug mode.**
image

Get an API error (this error throws when the domain is connected to another Pinterest account)
1. Install the plugin and do the onboarding with a Pinterest account.
2. Logout of the Pinterest account and log into a different one.
3. Do the plugin's onboarding again.
4. An error will appear in logs (explaining that user is already connected to a different Pinterest account)
5. The log for the error should be more clear now (as described above).

### Changelog entry

> Tweak - Improvements on the error logs.

### DIFF
--- a/src/API/Base.php
+++ b/src/API/Base.php
@@ -122,8 +122,18 @@ class Base {
 			} else {
 				$pinterest_code = method_exists( $e, 'get_pinterest_code' ) ? '( pinterest code: ' . $e->get_pinterest_code() . ')' : '';
 
-				/* Translators: 1: Request method 2: Request endpoint 3: Response status code 4: Response message 5: Pinterest code */
-				Logger::log( sprintf( esc_html__( "%1\$s Request: %2\$s\nStatus Code: %3\$s\nAPI response: %4\$s%5\$s", 'pinterest-for-woocommerce' ), $method, $request['url'], $e->getCode(), $e->getMessage(), $pinterest_code ), 'error' );
+				Logger::log(
+					sprintf(
+						/* Translators: 1: Request method 2: Request endpoint 3: Response status code 4: Response message */
+						esc_html__( "%1\$s Request: %2\$s\nStatus Code: %3\$s\nAPI response: %4\$s", 'pinterest-for-woocommerce' ),
+						$method,
+						$request['url'],
+						$e->getCode(),
+						$e->getMessage(),
+						$pinterest_code
+					),
+					'error'
+				);
 			}
 
 			throw $e;
--- a/src/API/Base.php
+++ b/src/API/Base.php
@@ -114,13 +114,34 @@ class Base {
 			}
 
 			return $response;
+		} catch ( ApiException $e ) {
+
+			if ( ! empty( Pinterest_For_WooCommerce()::get_setting( 'enable_debug_logging' ) ) ) {
+				/* Translators: 1: Error message 2: Stack trace */
+				Logger::log( sprintf( "%1\$s\n%2\$s", $e->getMessage(), $e->getTraceAsString() ), 'error' );
+			} else {
+
+				Logger::log(
+					sprintf(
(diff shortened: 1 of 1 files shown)


ksere (approved):

src/API/Base.php
```diff
@@ -116,7 +116,15 @@ public static function make_request( $endpoint, $method = 'POST', $payload = arr
...
 
-			Logger::log( $e->getMessage(), 'error' );
+			if ( ! empty( Pinterest_For_WooCommerce()::get_setting( 'enable_debug_logging' ) ) ) {
+				/* Translators: 1: Error message 2: Stack trace */
+				Logger::log( sprintf( "%1\$s\n%2\$s", $e->getMessage(), $e->getTraceAsString() ), 'error' );
+			} else {
+				$pinterest_code = method_exists( $e, 'get_pinterest_code' ) ? '( pinterest code: ' . $e->get_pinterest_code() . ')' : '';
+
+				/* Translators: 1: Request method 2: Request endpoint 3: Response status code 4: Response message 5: Pinterest code */
+				Logger::log( sprintf( esc_html__( "%1\$s Request: %2\$s\nStatus Code: %3\$s\nAPI response: %4\$s%5\$s", 'pinterest-for-woocommerce' ), $method, $request['url'], $e->getCode(), $e->getMessage(), $pinterest_code ), 'error' );
```
JPry: This line should be split across multiple lines for better readability.

Also, if we're going to look at the `$pinterest_code` variable as something relevant, why not have a separate `catch` statement for the `PinterestApiException` object where we know that method will exist?

alexiglesias31: @JPry about the multi catch statement, would be a nice touch from a code perspective.
However, having 2 separate calls to `Logger`, will create 2 `ERROR` lines (with the time header), not sure if this will cause some kind of confusion, wdyt?

JPry: @alexiglesias31 The two separate error logs would be within separate `catch` statements, so there wouldn't be duplicate log statements. Therefore I don't think there will be much confusion 🙂

JPry (approved):
This looks good. I think separate `catch` statements based on the type of exception would be an improvement, but it isn't a necessity.

alexiglesias31:
Done with the multi `catch` statements :+1:
Just pointing up here, if the Pinterest code is not a must we can skip this in order of avoid duplicate code.

cc @jconroy

JPry (approved):

PR #459, opened 2022-05-09
Bump WC & WP compatibility

tomalec:

- Use `PageController::is_admin_page` if available,
instead of deprecated `Loader::is_admin_page`,
to support WC 6.5 without a warning.

Quickly smoke test:
1. Plugin installation & activation
2. Account setup
3. Changing settings
4. Disconnecting account
5. `npm i && npm run test:js` (`npm run test:php` doesn't work on my local docker setup)

### Changelog entry

> Tweak - WooCommerce 6.5 and WordPress 6.0 compatibility.

### DIFF
--- a/pinterest-for-woocommerce.php
+++ b/pinterest-for-woocommerce.php
@@ -22,11 +22,11 @@
  * Domain Path:       /i18n/languages
  *
  * Requires at least: 5.6
- * Tested up to: 5.9
+ * Tested up to: 6.2
  * Requires PHP: 7.3
  *
  * WC requires at least: 5.3
- * WC tested up to: 6.3
+ * WC tested up to: 6.5
  */
 
 /**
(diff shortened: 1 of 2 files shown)


tomalec:
@ksere would you like to do any additional testing?

pinterest-for-woocommerce.php
```diff
@@ -22,11 +22,11 @@
  * Domain Path:       /i18n/languages
  *
  * Requires at least: 5.6
- * Tested up to: 5.9
+ * Tested up to: 6.2
```
JPry, line 25: @tomalec These bumps can be done during the release process as part of Woorelease rather than by manually changing the versions in a PR.

tomalec: Should I revert it, or could we leave it as is?

tomalec:
If there are no more tests to be performed, I mark the PR as ready for review.

findingsimple:
Yep lets just go for it - not getting any replies to this one which I'll flag

pinterest-for-woocommerce.php
findingsimple: @JPry but we'd still have to include changelog entry via PR right?

JPry: @jconroy 

> but we'd still have to include changelog entry via PR right?

Using the Woorelease commands to do the update will also add to the changelog.

@tomalec 

> Should I revert it, or could we leave it as is?

We can leave it since it's already there, but for the future it is easier to bump version headers with Woorelease.

findingsimple:
@JPry besides the comment updates is it ready to merge? I want to ship it

tomalec:
Merged given the approval on slack

PR #465, opened 2022-05-20
Verify if parent product is valid

### LINKED ISSUE #462: orphaned variation breaks XML feed
### Describe the bug:

In https://github.com/woocommerce/pinterest-for-woocommerce/blob/develop/src/ProductsXmlFeed.php#L172:L173, `wc_get_product()` returns `false` when a product/product_variation's parent does not exist. This logs a warning in the Pinterest -> Catalog screen:

```
XML feed:	 Could not get feed info.   •   Last activity: 11 mins ago - Automattic\WooCommerce\Pinterest\Product\Attributes\AttributeManager::get_all_values(): Argument #1 ($product) must be of type WC_Product, bool given, called in [/var/www/example.com/wp-content/plugins/pinterest-for-woocommerce/src/ProductsXmlFeed.php](http://example.com/wp-content/plugins/pinterest-for-woocommerce/src/ProductsXmlFeed.php) on line 166
```

### Steps to reproduce:

1.  Edit a test product in the database and give it a post_parent value that is greater than 0 and is not the ID of a product.
2.  Go to the Pinterest -> Catalog screen
3. 

Tested with the latest version of WC and this plugin.

alexiglesias31:
Fix orphaned variation products breaking the feed.

Closes #462.

### Screenshots:

Current behavior:
Screenshot_20220520_122650

1. Install the plugin and do the onboarding process.
2. Create a variable product with some variations.
3. Edit one of the variations directly on the database and change the parameter post_parent to a different value (>0 and != than parent product ID).
4. The feed should be generated without errors.

### Changelog entry

> Fix - Error on feed generation due to orphaned variations.

### DIFF
--- a/src/ProductsXmlFeed.php
+++ b/src/ProductsXmlFeed.php
@@ -168,7 +168,7 @@ class ProductsXmlFeed {
 		$xml               = '';
 
 		// Merge with parent's attributes if it's a variation product.
-		if ( $product instanceof WC_Product_Variation ) {
+		if ( $product instanceof WC_Product_Variation && ! empty( $product->get_parent_id() ) ) {
 			$parent_product    = wc_get_product( $product->get_parent_id() );
 			$parent_attributes = $attribute_manager->get_all_values( $parent_product );
 			$attributes        = array_merge( $parent_attributes, $attributes );
--- a/src/ProductsXmlFeed.php
+++ b/src/ProductsXmlFeed.php
@@ -168,10 +168,13 @@ class ProductsXmlFeed {
 		$xml               = '';
 
 		// Merge with parent's attributes if it's a variation product.
-		if ( $product instanceof WC_Product_Variation && ! empty( $product->get_parent_id() ) ) {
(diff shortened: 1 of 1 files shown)


sc-ecolasurdo (approved):

src/ProductsXmlFeed.php
```diff
@@ -168,7 +168,7 @@ private static function get_attributes_xml( $product, $indent ) {
 		$xml               = '';
 
 		// Merge with parent's attributes if it's a variation product.
-		if ( $product instanceof WC_Product_Variation ) {
+		if ( $product instanceof WC_Product_Variation && ! empty( $product->get_parent_id() ) ) {
```
JPry: Instead of checking `! empty( $product->get_parent_id() )`, I think it would be better to check the value of `$parent_product` from the next line. A parent ID can be non-empty (therefore passing the if statement), yet still point to an invalid product (therefore triggering `wc_get_product()` to return a non-object).

JPry (changes requested):

JPry (approved):
Looks good!

Issue #372, opened 2022-02-21
Product attributes Condition and Google Category fields can only be added to Simple/variable products


ChristianAF:
### Describe the bug:
Condition and Google Category fields can only be added to Simple/variable products. For other products like Subscriptions after editing and saving, Condition and Google Category fields are still not saved.

### Steps to reproduce:

1. Install Pinterest plugin (1.0.2 or higher)
2. Edit a Subscription product
3. go to "Pinterest" Tab to edit the Attributes Condition and Google Category
4. Save
5. Go to "Pinterest" Tab again

### Expected behavior:
the Attributes Condition and Google Category are saved

### Actual behavior:
the Attributes Condition and Google Category are not saved

### Additional details:
For Simple/Variable products is working ok

findingsimple:
@mikkamp was there a technical/support reason behind this? I'm guessing we copied the same approach as GLA.

mikkamp:
> was there a technical/support reason behind this? I'm guessing we copied the same approach as GLA

Yes the same logic was copied as GLA where we support only specific product types. Although Subscriptions in particular shows/hides data fields for Simple/Variable products (depending on subscription type). So we added an additional [compatibility class](https://github.com/woocommerce/google-listings-and-ads/blob/1.11.1/src/Integration/WooCommerceSubscriptions.php) if the Subscriptions plugin was activated (that class didn't get added here).

Based on the description in this PR it sounds like Subscription products should be synced in the feed? Is that the intention, and if so is it correctly populating the feed with the subscription data?

Looks like the only product type we are excluding by default is "grouped". If we are going to include all other product types and expect their attribute data to be set for pinterest then we will need some changes to get it working for all other product types.

findingsimple:
Marking as on-hold while waiting for confirmation about subscriptions support from @niketpins

findingsimple:
Subscriptions aren't supported (but the "products" are) - however the price field doesn't really allow for all the different factors that go into a subscription product pricing model e.g.

![free-trial-sign-up-fee@2x](https://user-images.githubusercontent.com/355014/158925981-7338ddbb-8188-4ffe-9939-a65808f485f6.jpg)

Safest option at the moment is to exclude subscriptions products (like GLA) and we can circle back to this at a later point

alexiglesias31:
@jconroy

> Safest option at the moment is to exclude subscriptions products (like GLA) and we can circle back to this at a later point

[...]

PR #468, opened 2022-05-24
Remove subscription products support

alexiglesias31:

Remove the support for subscription products as discussed [here](https://github.com/woocommerce/pinterest-for-woocommerce/issues/372)
- Exclude subscription products from the feed.
- Hide Pinterest attributes tab on subscription products.

### Screenshots:

Screenshot_20220524_153811

2. Add one/many subscription products.
3. Subscription products should be excluded from the feed.
4. Pinterest attribute tab should be hidden on subscription products.

### DIFF
--- a/src/Admin/Product/Attributes/AttributesTab.php
+++ b/src/Admin/Product/Attributes/AttributesTab.php
@@ -188,6 +188,8 @@ class AttributesTab {
 	/**
 	 * Return an array of WooCommerce product types that the Pinterest tab cannot be displayed for.
 	 *
+	 * @since x.x.x
+	 *
 	 * @return array of WooCommerce product types (e.g. 'subscription', 'variable-subscription', etc.)
 	 */
 	protected function get_hidden_product_types(): array {
--- a/src/Admin/Product/Attributes/AttributesTab.php
+++ b/src/Admin/Product/Attributes/AttributesTab.php
@@ -94,7 +94,14 @@ class AttributesTab {
 			$this->get_applicable_product_types()
 		);
 
-		$classes = array_merge( array( 'pinterest' ), $shown_types );
+		$hidden_types = array_map(
+			function ( string $product_type ) {
+				return "hide_if_${product_type}";
+			},
+			$this->get_hidden_product_types()
+		);
+
+		$classes = array_merge( array( 'pinterest' ), $shown_types, $hidden_types );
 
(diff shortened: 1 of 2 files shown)


sc-ecolasurdo (approved):

findingsimple:
Currently in draft, I'll still request a review from Ventures, but sing out if I'm missing something

alexiglesias31:
@jconroy we were doing some extra QA on this. It's ready for review now.

jorgemd24 (approved):
Thanks, @alexiglesias31, LGTM 👍 

I have tested locally with two subscription products: simple and with variations and I can confirm that the Pinterest tab is not displayed on the product page and the feed is excluding those products too.

alexiglesias31:
Thanks @jorgemd24

findingsimple:
I'm torn, I feel like it would make more sense to have an "include" list rather than exclude.

mikkamp:
> I'm torn, I feel like it would make more sense to have an "include" list rather than exclude.

In GLA we use an include list for product types that are synced.

However for hiding the product tabs we need an exclude list as well. Because subscription products show a lot of the same data as simple/variable for the product tabs. Since those product types are included by default it's also shown on the subscription product types. So we need to specifically add the `hide_if` to prevent the tab from showing up.

jorgemd24:
I see that we have an "include" already in the code however it is fetching all the product types and then excluding.

https://github.com/woocommerce/pinterest-for-woocommerce/blob/c4c6fa0583a3f83803586586f68fd86533ca0498/src/FeedGenerator.php#L222-L245

We could have something similar to the GLA plugin, allowing specific product types :

https://github.com/woocommerce/google-listings-and-ads/blob/fbb8f211d4972f0b2c2101af282c8a5dc9a368e2/src/Product/ProductSyncer.php#L250-L257

https://github.com/woocommerce/google-listings-and-ads/blob/6b02effe129d935da66f9e2fab79be92d9d740d4/src/DB/ProductFeedQueryHelper.php#L132-L133

Not sure what is more convenient, if by default we allow all product types except some of them or we only allow specific product types.

findingsimple:
> We could have something similar to the GLA plugin, allowing specific product types :

I think that works better so that we aren't syncing (custom) product types we aren't familiar with and then add a filter if someone wants to customize. 

My read at the moment is all product types will sync even ones that might(?) not work such as events/tickets/bookings

PR #469, opened 2022-05-25
Redirect user to onboarding page when the plugin is activated

### LINKED ISSUE #420: As soon as our plugin is activated on a new install, redirect user to the plugin main page
As soon as our plugin is activated on a new install, redirect user to the plugin main page

This will help users take action with connecting with pinterest as soon as they install

alexiglesias31:

Closes #420.

Redirect the user to the onboarding page when the plugin is activated on a fresh install.

### Screenshots:

Screenshot_20220524_231207

1. Install Woocommerce plugin and activate.
2. Install the Pinterest for Woocommerce plugin and activate.
3. The user should be redirected to the Get started page automatically. 

### Changelog entry

> Add - The user will be redirected to the Get started page on first activation.

### DIFF
--- a/src/Admin/ActivationRedirect.php
+++ b/src/Admin/ActivationRedirect.php
@@ -26,7 +26,7 @@ class ActivationRedirect {
 	 *
 	 * @since  x.x.x
 	 */
-	protected const REDIRECT_SETTING = 'redirect_to_onboarding';
+	protected const DID_REDIRECT_SETTING = 'did_redirect_to_onboarding';
 
 	/**
 	 * Initiate class.
@@ -46,34 +46,32 @@ class ActivationRedirect {
 	 * Checks if merchant should be redirected to the onboarding page.
 	 *
 	 * @since  x.x.x
-	 * @return bool True if the redirection should have happened
 	 */
-	public function maybe_redirect_to_onboarding(): bool {
+	protected function maybe_redirect_to_onboarding() {
 		if ( wp_doing_ajax() ) {
-			return false;
+			return;
 		}
 
 		// If we have redirected before do not attempt to redirect again.
-		if ( ! $this->redirect_setting() ) {
-			return false;
+		if ( ! $this->did_redirect_setting() ) {
+			return;
 		}
 
 		// Do not redirect if setup is already complete.
 		if ( Pinterest_For_Woocommerce()::is_setup_complete() ) {
-			$this->update_redirect_setting( false );
-			return false;
+			$this->update_did_redirect_setting( false );
+			return;
 		}
 
 		// Do not redirect if we already are in the Get Started page.
-		if ( $this->is_onboarding_page() && $this->redirect_setting() ) {
-			$this->update_redirect_setting( false );
-			return false;
+		if ( $this->is_onboarding_page() && $this->did_redirect_setting() ) {
+			$this->update_did_redirect_setting( false );
+			return;
 		}
 
 		// Redirect if setup is not complete.
 		$this->redirect_to_onboarding_page();
 
-		return true;
 	}
 
 	/**
@@ -89,7 +87,7 @@ class ActivationRedirect {
 			return;
 		}
 
-		$this->update_redirect_setting( true );
+		$this->update_did_redirect_setting( true );
 
 		wp_safe_redirect( admin_url( add_query_arg( $this->onboarding_page_parameters(), 'admin.php' ) ) );
 		exit();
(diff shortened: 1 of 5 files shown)


sc-ecolasurdo (approved):

src/Admin/ActivationRedirect.php
```diff
@@ -0,0 +1,135 @@
+<?php
+/**
+ * Helper class for handling the redirection to the onboarding page.
+ *
+ * @package Automattic\WooCommerce\Pinterest\Admin
+ * @since   X.X.X
```
JPry: All of these placeholders should be lowercase. I won't flag them all, but they should be updated in all of these files.

```suggestion
 * @since   x.x.x
```

findingsimple:
Forgot to leave a note that it is currently in draft but I've still request a review from Ventures because it is already got an approval

src/Admin/ActivationRedirect.php
```diff
@@ -0,0 +1,135 @@
...
+
+	/**
+	 * Get the redirect setting.
+	 *
+	 * @since  X.X.X
+	 * @return bool
+	 */
+	protected function redirect_setting(): bool {
+		$redirect_setting = Pinterest_For_Woocommerce()->get_setting( self::REDIRECT_SETTING );
+		return isset( $redirect_setting ) ? (bool) $redirect_setting : false;
```
JPry: Since the default value returned from `get_setting()` is `false`, we can simplify this to a single line.

```suggestion
		return (bool) Pinterest_For_Woocommerce()->get_setting( self::REDIRECT_SETTING );
```

src/Admin/ActivationRedirect.php
```diff
@@ -0,0 +1,135 @@
...
+		);
+	}
+
+	/**
+	 * Checks if merchant should be redirected to the onboarding page.
+	 *
+	 * @since  X.X.X
+	 * @return bool True if the redirection should have happened
+	 */
+	public function maybe_redirect_to_onboarding(): bool {
```
JPry: Since the only call to this method is enclosed in an anonymous function above, this can be `protected` or `private`.

src/Admin/ActivationRedirect.php
```diff
@@ -0,0 +1,135 @@
...
+				$this->maybe_redirect_to_onboarding();
+			}
+		);
+	}
+
+	/**
+	 * Checks if merchant should be redirected to the onboarding page.
+	 *
+	 * @since  X.X.X
+	 * @return bool True if the redirection should have happened
```
JPry: I don't think we should have any kind of return value. This will either do the redirect and subsequently exit without actually returning anything, or it won't, and no return is needed.

src/Admin/ActivationRedirect.php
```diff
@@ -0,0 +1,135 @@
...
+	 * @since  X.X.X
+	 * @return void
+	 */
+	protected function redirect_to_onboarding_page(): void {
+		// If we are already on the onboarding page, do nothing.
+		if ( $this->is_onboarding_page() ) {
+			return;
+		}
+
+		$this->update_redirect_setting( true );
```
JPry: The current name for the method `update_redirect_setting()` is a bit confusing. I'm noting it here because this is the main place where tracking the setting because a bit hard. Is the setting that we need to do the redirect? Or is it saying that we've already done the redirect?

Reading through this, it looks like the intention is the latter. With that in mind, I think we should update some of the method names to be more descriptive based on what we're actually checking or doing.

JPry (changes requested):

alexiglesias31:
Thanks for the review @JPry. Just updated the changes :+1:

src/PluginHelper.php
```diff
@@ -54,4 +54,31 @@ protected function is_debug_mode(): bool {
...
+	 *
+	 * @return bool Wether the current page is the Get Started page.
+	 */
+	protected function is_onboarding_page(): bool {
+
+		if ( count( $this->onboarding_page_parameters() ) === count( array_intersect_assoc( $_GET, $this->onboarding_page_parameters() ) ) ) { // phpcs:disable WordPress.Security.NonceVerification.Recommended
+			return true;
+		}
+
+		return false;
```
JPry: I think we can simplify this method body and make it slightly more readable at the same time:

```suggestion
		$page_parameters = $this->onboarding_page_parameters();

		return count( $page_parameters ) === count( array_intersect_assoc( $_GET, $page_parameters) );
```

JPry (commented):
Thanks for the update @alexiglesias31. One more tweak and then I think this is good to go.

alexiglesias31:
Thanks for the extra suggestion @JPry. Tweak added :+1:

JPry (approved):
Looks good!

PR #474, opened 2022-06-07
Add current strings to pot file

### LINKED ISSUE #470: Change strings and Add the POT files into the build for i18n
1. Remove html encoding in the strings from POT

```
line 473 msgid "Search for a category&hellip;" (replace &hellip; with ...)
line 12 msgid "Cheatin&#8217; huh?" (replace &8217; with ') 

```

2. Please checkin the POT files into the build so that it can be monitored for new strings to help with translation

JPry:

Closes #470.

### Changelog entry

> Add - .pot file for translators


findingsimple:
@jorgemd24 when you have a moment can you double check this one please

JPry:
@jconroy @jorgemd24 I just noticed that this didn't pick up any of the JS translatable strings. I'm looking into getting those added before having this reviewed.

budzanowski:
Sorry for the drive-by comment 😉

The .pot file should be removed from the repository entirely. It should never be committed.
It should only be a part of a release package, hence generated during the build process.

The preferable way to generate the .pot file is using WP-CLI `i18n` command. Preferably in the `./languages` location. 

The reasoning behind this is the same as with any generated files. On top of that this file can change extensively even with small changes to the codebase, polluting the diffs with no essential changes.

JPry:
Thanks @budzanowski, hope your leave is going well 🙂 

> The .pot file should be removed from the repository entirely. It should never be committed.
> It should only be a part of a release package, hence generated during the build process.

You are correct, but this is at the request of the Pinterest team. They would like to include the `.pot` file in the repo as part of their process, which is why we're adding it now.

> On top of that this file can change extensively even with small changes to the codebase, polluting the diffs with no essential changes.

This would be minimized by only updating it (and committing those updates) during the release process rather than with every change that would otherwise impact that file.

JPry:
@jorgemd24 This is ready for review now.

package.json
```diff
@@ -102,13 +102,14 @@
...
-    "postbuild": "composer install",
-    "build:zip": "npm run build",
+    "prebuild": "composer install",
+    "build": "wp-scripts build && gulp build && npm run i18n",
+    "build:zip": "npm run build && composer install --no-dev -n && gulp zip",
+    "postbuild:zip": "composer install",
     "dev": "wp-scripts build && gulp build",
     "doc:tracking": "jsdoc ./assets/source/ -c .jsdocrc.json -t ./woo-tracking-jsdoc",
     "format:js": "wp-scripts format ./assets/source",
+    "i18n": "WP_CLI_PHP_ARGS='-d memory_limit=2048M' ./vendor/bin/wp i18n make-pot ./ i18n/languages/$npm_package_name.pot --slug=$npm_package_name --domain=$npm_package_name --include=assets/build,includes,src,views,*.php",
```
jorgemd24: ❓

Is it not a bit too permissive to have *.php? as it will include any PHP file independent of its directory, I see that by default [wp i18n make-pot](https://developer.wordpress.org/cli/commands/i18n/make-pot/) will exclude `node_modules, .git, .svn, .CVS, .hg, vendor, *.min.js.`. but some folders like bin/tests etc could be excluded too. Compared with the GLA plugin we exclude some folders like bin, data, tests, etc

`--exclude=bin,data,js/src,node_modules,tests,vendor`

What do you think?

jorgemd24 (commented):
Thanks, @JPry, for adding this! I did not see any testing steps 😆 but this is what I tested so far:

- I checked that PHP and JS translations are in the POT file.
- I checked that `npm run i18n` is including the right folders. I left a comment about it.
- I run `npm run build` and `npm run build:zip` and I see that the POT is regenerated correctly. I am unsure if adding the POT to the zip file is needed? 
- I have updated some strings and the POT is updated after running `npm run build`.
- I see in the following issue https://github.com/woocommerce/pinterest-for-woocommerce/issues/470 that there is an issue with the HTML encoding, is this part of this PR? Currently, the POT still showing HTML encoding characters.

budzanowski:
@JPry 

> Thanks @budzanowski, hope your leave is going well 🙂

Thanks, it is going well. 

> You are correct, but this is at the request of the Pinterest team. They would like to include the .pot file in the repo as part of their process, which is why we're adding it now.

🤷🏻‍♂️ OK

package.json
JPry: @jorgemd24 I think you're right. I played around with some of the options to try to get the right rules, and pushed 6ababe2.

My goal here was to exclude the built files, but still capture things correctly from the `assets/` directory.

package.json
```diff
@@ -102,13 +102,14 @@
...
-    "postbuild": "composer install",
-    "build:zip": "npm run build",
+    "prebuild": "composer install",
+    "build": "wp-scripts build && gulp build && npm run i18n",
+    "build:zip": "npm run build && composer install --no-dev -n && gulp zip",
+    "postbuild:zip": "composer install",
     "dev": "wp-scripts build && gulp build",
     "doc:tracking": "jsdoc ./assets/source/ -c .jsdocrc.json -t ./woo-tracking-jsdoc",
     "format:js": "wp-scripts format ./assets/source",
+    "i18n": "WP_CLI_PHP_ARGS='-d memory_limit=2048M' ./vendor/bin/wp i18n make-pot ./ i18n/languages/$npm_package_name.pot --slug=$npm_package_name --domain=$npm_package_name --exclude='bin,*.asset.php,assets/source/_maps,assets/setup-*,assets/catalog-sync,assets/build,tests,woo-tracking-jsdoc,index.php,gulpfile.js,jest.config.js'",
```
jorgemd24, line 112: @JPry,  I have a doubt here: is your intention here to exclude `assets/setup-*` and `assets/catalog-sync`? If so shouldn't this be `assets/src/setup-*` and `assets/src/catalog-sync`? 

Is there a reason to exclude it? Or could we just do 
`--exclude='bin,*.asset.php,assets/source/_maps,assets/build,tests,woo-tracking-jsdoc,index.php,gulpfile.js,jest.config.js'` ?

JPry: @jorgemd24 

> is your intention here to exclude `assets/setup-*` and `assets/catalog-sync`? If so shouldn't this be `assets/src/setup-*` and `assets/src/catalog-sync`? 

My intention is to target the development files for the `i18n` strings rather than the built files. While some of the build files are in `assets/build`, some are in those individual directories as well, which can be determined by the `.gitignore` file:

https://github.com/woocommerce/pinterest-for-woocommerce/blob/3adfe41bfcde27a0a3cc3fd4ed57c615e4265304/.gitignore#L11-L16

So the set of exclude rules here is meant to exclude all of the build directories in `assets/`.

jorgemd24: @JPry Thanks for the clarification!

jorgemd24 (approved):
LGTM 👍 

- I tested that PHP and JS translations are in the POT file.
 - I have updated some strings and the POT is updated after running npm run build.

PR #475, opened 2022-06-08
Use `-jsdoc` plugins from woocommerce/grow

tomalec:
_This PR requires https://github.com/woocommerce/grow/pull/13_

Use `-jsdoc` plugins from woocommerce/grow,
instead of maintaining a local copy.

1. Remove a few lines from https://github.com/woocommerce/pinterest-for-woocommerce/blob/develop/TRACKING.md?plain=1#L17-L22 to see that the result is generated
2. ```sh
   npm i
   npm run doc:tracking
   ```
3. Check that `TRACKING.md` is generated correctly


### Additional details:

<!--
Optional.
Enter a summary of all changes in this Pull Request, which will be added to the changelog if accepted.
Each line should start with `(Fix|Add|Tweak|Update) - `, for example:
> Fix - I took care of something that wasn't working.
> Add - I added something new that's pretty cool.
> Tweak - I made a small change.
> Update - I made big changes to something that wasn't broken.

Leave the "Changelog entry" header in place completely empty, without any summary if no changelog entry is needed.
If you remove the "Changelog entry" header, the title of Pull Request will be used as the changelog entry.
-->
### Changelog entry


TRACKING.md
```diff
@@ -8,9 +8,9 @@ When a store opts in to WooCommerce usage tracking and uses _Pinterest for WooCo
 
 As in WooCommerce core, only non-sensitive data about how a store is set up and managed is tracked. We **do not track or store personal data** from your clients.
 
-<woo-tracking-jsdoc>
+<woocommerce-grow-tracking-jsdoc>
```
jorgemd24, line 11: ❔

Not related to Pinterest itself, more related to `woocommerce-grow-jsdoc` 😄  wouldn't it be better to use a more generic name instead of `woocommerce-grow-tracking-jsdoc`? Something like `<tracking-jsdoc>`? I am just thinking that maybe other divisions or external people want to use this package outside of Grow and they would like to have something more generic. It is not an issue of course but more a curiosity.

jorgemd24 (approved):
Thanks @tomalec, I followed the steps mentioned in the PR and I can see that the TRACKING.md is generated correctly. 

LGTM 👍

TRACKING.md
tomalec: We briefly discussed it at https://github.com/woocommerce/grow/pull/6#pullrequestreview-982416113 and I created a somewhat related issue in https://github.com/woocommerce/grow/issues/7

I think once, we decide it's ready for publishing to the outer world, we could change that. To me one of the purposes of `/grow` is to incubate tools before proposing to external people. 

Before that, I think it would be nice to have that explicit distinction, between which things are already woo-generic, and which are our incubated things.

However, it would require tedious work to change that later on, and having it generic and short from day one could give more benefits.

------

Speaking of `<tracking-jsdoc>` the good practice for custom element names is to prefix them with the vendor. Even though technically this is not a custom element per se, just a custom named `HTMLElement` `(`HTMLUnknownElement`). But IMHO it's good to stick to that practice, as we may someday create a custom element for that that could do some work in the browser.

Personally, I'd rather shorten this to `<woo-tracking-jsdoc>`/`<wc-tracking-jsdoc>`

jorgemd24: Thanks, @tomalec for your detailed explanation! 😄

PR #488, opened 2022-06-28
Save settings if new value is different

### LINKED ISSUE #477: Domain verification page/save settings error during the onboarding
Discription:
Users report a problem with site verification and saving settings during onboarding. It's impossible to go through and finish.

Steps:
Claim your website" the website is prefilled and when I press start verification it brings up the error message "There was an error saving the settings"
logs:
“2022-06-08T15:10:15+00:00 ERROR Sorry! We couldn't find that merchant. Please ensure you have access and a valid merchant id.”

Massage from Piterest:
I am seeing a pattern from 05/15. The errors have spiked from that date. Was there any change on that day? One main problem I see is _Partner connect failed with exception No relevant meta tag was found._

I speicifically couldnt find the above websites in the log, but there is a clear pattern from that day

image

you can see that 99.1% of the errors are:
_Domain_Verifification_Error_Realtime_Nonretryable failed for website [bathroombutler.com/us](http://bathroombutler.com/us) and user 686376936867416861 with result no_tag_
and the domain does not have the meta tag
_<meta name="p:domain_verify" content="..."/>_

I checked these above domains and they too dont have the meta tag
https://angelblessings.ca/
raymond-weil.com/en/

Something to check on plugin's end ?
user is bathroombutler on pinterest.

alexiglesias31:

Closes #477.

The `account_data` setting is updated as part of the domain verification.
The call to `save_settings` after domain verification was throwing error, since old and new values are the same. A verification of old and new values will prevent this error.

2. There should be no error after domain verification.

*Changelog entry once this is ready:* `Fix - Error after domain verification`. The changelog entry below is left empty on purpose.

### Changelog entry

>

### DIFF
--- a/class-pinterest-for-woocommerce.php
+++ b/class-pinterest-for-woocommerce.php
@@ -504,6 +504,11 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 		 * @return boolean
 		 */
 		public static function save_settings( $settings, $option = PINTEREST_FOR_WOOCOMMERCE_OPTION_NAME ) {
+
+			if ( self::get_settings( true, $option ) === $settings ) {
+				return true;
+			}
+
 			self::$dirty_settings[ $option ] = true;
 			return update_option( $option, $settings );
 		}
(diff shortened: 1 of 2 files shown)


sc-ecolasurdo (approved):

JPry (approved):
This approach is fine, but I feel like the `Options` class should really be handling this kind of situation in a better way. Since we're wrapping the `update_option()` function, we should improve on the handling, otherwise there's not a lot of benefit to wrapping it.

If you think that would be better as a separate issue/PR, I'm fine with that too. But I wanted to mention it while we're solving this issue here.

alexiglesias31:
@JPry 
Good point. I think it makes sense to include it in this PR. Will be adding the changes.

JPry (approved):
Looks good!

PR #495, opened 2022-07-12
Detect other tags installed

### LINKED ISSUE #417: Detect if there are other tags installed on website
## User story
Customers might already have other tag integrations installed before installing our plugin

We see at the creation of this ticket around atleast 8% of installed users have multiple tag plugins installed.

We need to detect these tags and warn the users to uninstall/remove pinterest tag so duplicate events arent sent

1. google tag manager
2. pixelyoursite
3. premmerce

Detecting the above 3 tags would help us in reducing a majority of cases

alexiglesias31:
Closes #417.

There are other tags that may potentially cause conflict with the plugin. We are showing warning messages, so the customer can decide if removing the other tags or disable the plugin's tracking.

The current PR adds detection for the following tags:
- Google Tag Manager.
- PixelYourSite - Pinterest Addon.

As mentioned by @jconroy [here](https://github.com/woocommerce/pinterest-for-woocommerce/issues/417#issuecomment-1173745843), in order to keep this PR small we can add support for more third party tags in other PRs.

### Screenshots:

pinterest-warning-catalog
pinterest-warning-tracking

2. Install and enable either Google Tag Manager or PixelYourSite - Pinterest Addon plugins.
3. The warning messages should be displayed on the Catalog tab and the Tracking section (Connection tab).
4. After disabling/uninstalling the third party extensions the warning should not be displayed.

### Changelog entry

> Add - Warning message on potential conflicting tags.


sc-ecolasurdo (approved):

JPry:
@alexiglesias31 @sc-ecolasurdo Is this PR ready for review?

findingsimple:
@JPry [there was a question here](https://github.com/woocommerce/pinterest-for-woocommerce/issues/417#issuecomment-1189335266)

alexiglesias31:
@jconroy our QA team needs a valid license of the plugin PixelYourSite - Pinterest Addon to fully validate the task.
We also need a copy of the plugin [Pinterest for Woocommerce by Softblue extension](https://woocommerce.com/products/softblues-pinterest-for-woocommerce/).

findingsimple:
@alexiglesias31 I've shared extensions and license keys via slack

alexiglesias31:
@jconroy @JPry this is ready for review.

findingsimple:
Thanks @alexiglesias31

src/Tracking.php
```diff
@@ -566,4 +566,61 @@ private static function filter_add_to_cart_attributes( array $args, WC_Product $
...
+
+		return wp_kses_post(
+			sprintf(
+				/* Translators: %1$s Conflicting plugins */
+				__( 'The following installed plugin(s) can potentially cause problems with tracking: %1$s. %2$sRemove conflicting plugins%3$s or %4$smanage tracking settings%5$s.', 'pinterest-for-woocommerce' ),
+				implode( ', ', $third_party_tags ),
+				sprintf( '<a href="%s" target="_blank">', esc_url( admin_url( 'plugins.php' ) ) ),
+				'</a>',
+				sprintf( '<a href="%s" target="_blank">', esc_url( wc_admin_url( '&path=/pinterest/settings' ) ) ),
+				'</a>',
```
JPry: Since the closing tag is the same, we can use the numbered placeholders to eliminate an extra argument.

```suggestion
				__( 'The following installed plugin(s) can potentially cause problems with tracking: %1$s. %2$sRemove conflicting plugins%4$s or %3$smanage tracking settings%4$s.', 'pinterest-for-woocommerce' ),
				implode( ', ', $third_party_tags ),
				sprintf( '<a href="%s" target="_blank">', esc_url( admin_url( 'plugins.php' ) ) ),
				sprintf( '<a href="%s" target="_blank">', esc_url( wc_admin_url( '&path=/pinterest/settings' ) ) ),
				'</a>',
```

src/Tracking.php
```diff
@@ -566,4 +566,61 @@ private static function filter_add_to_cart_attributes( array $args, WC_Product $
...
+
+		$third_party_tags = self::get_third_party_installed_tags();
+
+		if ( empty( $third_party_tags ) ) {
+			return '';
+		}
+
+		return wp_kses_post(
+			sprintf(
+				/* Translators: %1$s Conflicting plugins */
```
JPry: All of the numbered placeholders should be included here, not just the first one. Also, you can leave off the `%$s` portion, so it would look something like this:

```suggestion
				/* Translators: 1: Conflicting plugins, 2: Plugins Admin page opening tag, 3: Pinterest settings opening tag, 4: closing anchor tag */
```

src/API/FeedState.php
```diff
@@ -141,6 +142,7 @@ public function get_feed_state() {
 
 			$result = $this->add_local_feed_state( $result );
 			$result = $this->add_feed_registration_state( $result );
+			$result = $this->add_third_party_tags_warning( $result );
```
JPry: Since we're adding more items to these results, and the structure of the code is treating these assignments as if we're using `apply_filters()`, I feel like we should just use a single call with `apply_filters()` here.

```php
return apply_filters( 'pinterest_for_woocommerce_feed_state', array() );
```

Then we can hook each of these methods (and any future methods) into the filter.

src/API/HealthCheck.php
```diff
@@ -44,17 +43,17 @@ public function health_check() {
...
 		try {
 
+			$response = array();
+
 			if ( ! Pinterest_For_Woocommerce()::get_data( 'merchant_id' ) ) {
-				return array(
-					'status' => 'pending_initial_configuration',
-				);
+				$response['status'] = 'pending_initial_configuration';
+				return $response;
```
JPry: Instead of using the `$response` variable, we can just directly return the array here.

```suggestion
				return array( 'status' => 'pending_initial_configuration' );
```

src/API/HealthCheck.php
```diff
@@ -44,17 +43,17 @@ public function health_check() {
...
 			}
 
 			$merchant_connected_diff_platform = Pinterest_For_Woocommerce()::get_data( 'merchant_connected_diff_platform' );
 			if ( $merchant_connected_diff_platform ) {
-				return array(
-					'status' => 'merchant_connected_diff_platform',
-				);
+				$response['status'] = 'merchant_connected_diff_platform';
+				return $response;
```
JPry: Similar suggestion as the above:

```suggestion
				return array( 'status' => 'merchant_connected_diff_platform' );
```

src/Tracking.php
```diff
@@ -566,4 +566,61 @@ private static function filter_add_to_cart_attributes( array $args, WC_Product $
...
+	 */
+	public static function get_third_party_tags_warning_message() {
+
+		$third_party_tags = self::get_third_party_installed_tags();
+
+		if ( empty( $third_party_tags ) ) {
+			return '';
+		}
+
+		return wp_kses_post(
```
JPry: Do we need to use `wp_kses_post()` for this whole message? It seems that we're only adding tags for links, so we could use `esc_html__()` instead of `__()` for the string below.

src/Tracking.php
```diff
@@ -566,4 +566,61 @@ private static function filter_add_to_cart_attributes( array $args, WC_Product $
...
+
+		if ( defined( 'GTM4WP_VERSION' ) ) {
+			$third_party_tags['gtm'] = 'Google Tag Manager';
+		}
+
+		if ( defined( 'PYS_PINTEREST_VERSION' ) ) {
+			$third_party_tags['pys'] = 'Pixel Your Site - Pinterest Addon';
+		}
+
+		if ( class_exists( '\Premmerce\WooCommercePinterest\PinterestPlugin' ) ) {
```
JPry: We should add a `use` import statement at the top of the file, then use `PinterestPlugin::class` here instead of the string version of the namespace.

```suggestion
		if ( class_exists( PinterestPlugin::class ) ) {
```

src/API/FeedState.php
```diff
@@ -324,6 +326,36 @@ private function add_feed_registration_state( $result ) {
...
+
+		if ( empty( $tags ) ) {
+			return $result;
+		}
+
+		$result['workflow'][] = array(
+			'label'        => esc_html__( 'Pinterest tag', 'pinterest-for-woocommerce' ),
+			'status'       => 'warning',
+			'status_label' => esc_html__( 'Potential conflicting plugins', 'pinterest-for-woocommerce' ),
+			'extra_info'   => Tracking::get_third_party_tags_warning_message(),
```
JPry: Instead of separate methods `get_third_party_installed_tags()` and `get_third_party_tags_warning_message()`, it might make more sense to have only one method, probably the warning message method. If there are any warning messages, then we can display them, but if they are empty then we know not to send any warnings.

JPry (changes requested):

assets/source/setup-guide/app/components/HealthCheck/index.js
```diff
@@ -248,16 +248,18 @@ const HealthCheck = () => {
...
-			status={ notice.status }
-			isDismissible={ notice.dismissible }
-			actions={ notice.actions || [] }
-			className="pinterest-for-woocommerce-healthcheck-notice"
-		>
-			<Text variant="titleSmall">{ notice.message }</Text>
-			<FormattedReasons reasons={ notice.reasons } />
-			{ notice.body }
-		</Notice>
+		<>
```
tomalec: :grey_question: Why is this Fragment needed here?

assets/source/setup-guide/app/components/ThirdPartyTagsNotice.js
```diff
@@ -0,0 +1,34 @@
...
+	}`;
+
+	return (
+		<Notice
+			status="warning"
+			isDismissible={ true }
+			className="pinterest-for-woocommerce-healthcheck-notice"
+		>
+			<span
+				dangerouslySetInnerHTML={ {
```
tomalec, line 26: I'd rather use `createInterpolateElement` from `@wordpress/element` To put messages in `<strong>` and `a` instead of using `dangerouslySetInnerHTML`

src/Tracking.php
tomalec: I think it's better to interpolate in one place. 
Currently, we interpolate messages, URLs, `a`nchors here in PHP, then interpolate message within `<strong>` and `<span>` in JS. 

I'd rather interpolate everything in JS. 
Then we get:
- Entire interpolation logic in one place, to make it easier to change and maintain in future.
- Security of not stamping any HTML from external sources.
- Flexibility of styling, as then we could control any additional classes, attributes, or HTMLElements/ReactComponents used to render the links.

alexiglesias31: Sounds good, my comment here would be that we also need the interpolated warning message to add it to the feed state. We probably can move the `<strong>` and `<span>` interpolation to PHP too in order to avoid duplicated code.

cc @tomalec @JPry

alexiglesias31: The warning message in the feed state should not contain the `<strong>` fragment tho, since that fragment is rendered by the 'status_label' [here](https://github.com/woocommerce/pinterest-for-woocommerce/blob/308b801e0111be8e6123562024c9a6e391026748/src/API/FeedState.php#L359)

alexiglesias31: @tomalec @JPry just a kind remainder about my question here :point_up: :relaxed:

JPry: @alexiglesias31 If we need to make use of the output in both JS and PHP, then I agree that it makes sense to do the interpolation in PHP.

src/Tracking.php
```diff
@@ -566,4 +567,58 @@ private static function filter_add_to_cart_attributes( array $args, WC_Product $
...
+
+		$third_party_tags = self::get_third_party_installed_tags();
+
+		if ( empty( $third_party_tags ) ) {
+			return '';
+		}
+
+		return sprintf(
+				/* Translators: 1: Conflicting plugins, 2: Plugins Admin page opening tag, 3: Pinterest settings opening tag, 4: Closing anchor tag */
+			esc_html__( 'The following installed plugin(s) can potentially cause problems with tracking: %1$s. %2$sRemove conflicting plugins%4$s or %4$smanage tracking settings%4$s.', 'pinterest-for-woocommerce' ),
```
JPry:
```suggestion
			esc_html__( 'The following installed plugin(s) can potentially cause problems with tracking: %1$s. %2$sRemove conflicting plugins%4$s or %3$smanage tracking settings%4$s.', 'pinterest-for-woocommerce' ),
```

JPry (approved):
Looks like one minor typo in the interpolated string, but otherwise the changes look good on the PHP side.

src/Tracking.php
tomalec: > The warning message in the feed state should not contain the `<strong>` fragment tho, since that fragment is rendered by the 'status_label'

I'm still a bit confused. Does it mean we need this `<strong>` to be interpolated that `dangerously…` way or not https://github.com/woocommerce/pinterest-for-woocommerce/pull/495/files#r947690286

alexiglesias31: @tomalec we can use `createInterpolateElement` as you suggested to interpolate the `<strong>Potential conflicting plugins detected.</strong>` tag and keep the rest of interpolations on PHP, since we are re-using the warning message on the feed state info too, and another prefix [is used there](https://github.com/woocommerce/pinterest-for-woocommerce/pull/495/files#diff-d7374912c385ac51b92ce25d055029bdb751eeb9f7e1cddd17015e182ee015f9R359).

findingsimple:
@tomalec can you please give this another review?

src/Tracking.php
alexiglesias31: @tomalec another alternative is to use a combination of `createInterpolateElement` and `RawHTML`

```
{ createInterpolateElement(
	__(
		'<strong>Potential conflicting plugins detected.</strong> <warningMessage/>'
	),
	{
		strong: <strong />,
		warningMessage: (
			<RawHTML>
				{
					wcSettings.pinterest_for_woocommerce
						.conflictingTagsWarning
				}
			</RawHTML>
		),
	}
) }
```

tomalec (approved):
As stated in https://github.com/woocommerce/pinterest-for-woocommerce/pull/495#discussion_r947703745 and https://github.com/woocommerce/pinterest-for-woocommerce/pull/495/files#r947690286 
I do have concerns regarding rendering HTML in PHP and the use of APIs like <code><strong>dangerously</strong>SetInnerHTML</code>.

IMHO, mixing responsibilities leads to harder and more expensive maintenance, and is bug-prone. I'd leave HTML rendering to react, and let PHP operate on bare data, regardless of client-side rendering language.

However, fixing that seems to require changing the flow around `$extra_info` which was not introduced here, and doing that would probably be beyond the scope of this PR. 

So this PR LGTM to be merged. 

However, I suggest noting that and cleaning it up in the future

src/Tracking.php
tomalec: I think we could as well re-use the interpolation code on the client side. 
But that would require a bit of  change around `extra_info` rendering.
Plus, a warning message could be considered something different that "extra info" and have its own property to be stored and handled a bit differently (even though eventually rendered in the same spot)

PR #496, opened 2022-07-14
Improve include and exclude methods

### LINKED ISSUE #372: Product attributes Condition and Google Category fields can only be added to Simple/variable products

### LINKED ISSUE #543: Exclude Composite products from feed.
### Describe the bug:
Merchant selling composite product:

>  I sell Composite products using the WooCommerce Composite Products Plugin. I would like to exclude them from the Pinterest feed because the single components don't make sense on Pinterest. I guess I just need to exclude the entire category "configurator" from my xml file (and also our gift vouchers which is another product category). How can I do that? Do I just edit the xml file through FTP or will it be overwritten by the plugin on a regular basis? I'd very much appreciate your support with this, thank you!

### Steps to reproduce:

1. install the plugin WooCommerce Composite Products
2. create a composite product with single components
3. Regenerate feed

### Expected behavior:
Composite product to be excluded from feed

### Actual behavior:
Composite products are not exculed

alexiglesias31:
Closes #372
Closes #543 

This PR follows up the PR #468.
As [discussed here](https://github.com/woocommerce/pinterest-for-woocommerce/pull/468#issuecomment-1150902058), the plugin will sync only the supported products, which can be filtered for customization by the customer.

2. The feed should contain by default only the following product types: 'simple', 'external' and 'variation'.
3. The included product types can be filtered using: `pinterest_for_woocommerce_included_product_types`.
4. Default product types for which Google Category and Condition is displayed are 'simple' and 'variable', it won't be displayed for the rest of product types by default.
5. The product types for which Pinterest Google Condition tab is displayed can be edited using the following filters: `wc_pinterest_attributes_tab_applicable_product_types` and `wc_pinterest_attributes_tab_hidden_product_types`.

### Changelog entry

>

### DIFF
--- a/src/Admin/Product/Attributes/AttributesTab.php
+++ b/src/Admin/Product/Attributes/AttributesTab.php
@@ -91,14 +91,14 @@ class AttributesTab {
 			function ( string $product_type ) {
 				return "show_if_${product_type}";
 			},
-			$this->get_applicable_product_types()
+			array_keys( $this->get_applicable_product_types() )
 		);
 
 		$hidden_types = array_map(
 			function ( string $product_type ) {
 				return "hide_if_${product_type}";
 			},
-			$this->get_hidden_product_types()
+			array_keys( $this->get_hidden_product_types() )
 		);
 
 		$classes = array_merge( array( 'pinterest' ), $shown_types, $hidden_types );
@@ -182,7 +182,13 @@ class AttributesTab {
 	 * @return array of WooCommerce product types (e.g. 'simple', 'variable', etc.)
 	 */
 	public static function get_applicable_product_types(): array {
-		return apply_filters( 'wc_pinterest_attributes_tab_applicable_product_types', array( 'simple', 'variable' ) );
+		return apply_filters(
+			'wc_pinterest_attributes_tab_applicable_product_types',
+			array(
+				'simple'   => __( 'Simple product', 'pinterest-for-woocommerce' ),
+				'variable' => __( 'Variable product', 'pinterest-for-woocommerce' ),
+			)
+		);
 	}
 
 	/**
@@ -193,7 +199,13 @@ class AttributesTab {
 	 * @return array of WooCommerce product types (e.g. 'subscription', 'variable-subscription', etc.)
 	 */
 	protected function get_hidden_product_types(): array {
-		return apply_filters( 'wc_pinterest_attributes_tab_hidden_product_types', array_diff( array_keys( wc_get_product_types() ), $this->get_applicable_product_types() ) );
+		return apply_filters(
+			'wc_pinterest_attributes_tab_hidden_product_types',
+			array_diff_key(
+				wc_get_product_types(),
+				$this->get_applicable_product_types()
+			)
+		);
 	}
 
 	/**
--- a/src/Admin/Product/Attributes/AttributesTab.php
+++ b/src/Admin/Product/Attributes/AttributesTab.php
@@ -105,7 +105,7 @@ class AttributesTab {
 
 		$tabs['pinterest_attributes'] = array(
 			'label'  => 'Pinterest',
-			'class'  => join( ' ', $classes ),
+			'class'  => $classes,
 			'target' => 'pinterest_attributes',
 		);
 
@@ -193,7 +193,7 @@ class AttributesTab {
 	 * @return array of WooCommerce product types (e.g. 'subscription', 'variable-subscription', etc.)
 	 */
 	protected function get_hidden_product_types(): array {
-		return apply_filters( 'wc_pinterest_attributes_tab_hidden_product_types', array( 'subscription', 'variable-subscription' ) );
+		return apply_filters( 'wc_pinterest_attributes_tab_hidden_product_types', array_diff( array_keys( wc_get_product_types() ), $this->get_applicable_product_types() ) );
 	}
 
 	/**
--- a/src/Admin/Product/Attributes/AttributesTab.php
+++ b/src/Admin/Product/Attributes/AttributesTab.php
@@ -181,7 +181,7 @@ class AttributesTab {
 	 *
 	 * @return array of WooCommerce product types (e.g. 'simple', 'variable', etc.)
 	 */
-	protected function get_applicable_product_types(): array {
+	public static function get_applicable_product_types(): array {
 		return apply_filters( 'wc_pinterest_attributes_tab_applicable_product_types', array( 'simple', 'variable' ) );
 	}
 
--- a/src/FeedGenerator.php
+++ b/src/FeedGenerator.php
@@ -222,13 +222,8 @@ class FeedGenerator extends AbstractChainedJob {
 			// Get included product types.
 			$included_product_types = self::get_supported_product_types();
 
-			// Get excluded product types.
-			$excluded_product_types = self::get_unsupported_product_types();
-
-			$types = array_diff( $included_product_types, $excluded_product_types );
-
 			$products_query_args = array(
-				'type'       => $types,
+				'type'       => $included_product_types,
 				'include'    => $items,
 				'visibility' => 'catalog',
 				'orderby'    => 'none',
@@ -553,25 +548,6 @@ class FeedGenerator extends AbstractChainedJob {
 				'simple',
 				'external',
 				'variation',
-			)
-		);
-	}
-
-	/**
-	 * Return the list of hidden product types.
-	 *
-	 * @since x.x.x
-	 *
-	 * @return array
-	 */
-	private function get_unsupported_product_types(): array {
-		return (array) apply_filters(
-			'pinterest_for_woocommerce_excluded_product_types',
-			array(
-				'grouped',
-				'variable',
-				'subscription',
-				'variable-subscription',
 			)
 		);
 	}
--- a/src/FeedGenerator.php
+++ b/src/FeedGenerator.php
@@ -549,7 +549,6 @@ class FeedGenerator extends AbstractChainedJob {
 			'pinterest_for_woocommerce_included_product_types',
 			array(
 				'simple',
-				'external',
 				'variation',
 			)
 		);
--- a/src/FeedGenerator.php
+++ b/src/FeedGenerator.php
@@ -219,28 +219,11 @@ class FeedGenerator extends AbstractChainedJob {
 	 */
 	protected function process_items( array $items, array $args ) {
 		try {
-			/**
-			 * Prepare allowed product types.
-			 * For variations:
-			 * We just want variations ( variable children ) but not parents ( variable ).
-			 */
-			$included_product_types = array_merge( array_keys( wc_get_product_types() ), array( 'variation' ) );
+			// Get included product types.
+			$included_product_types = self::get_supported_product_types();
 
-			/**
-			 * Filter excluded product types.
-			 *
-			 * @since 1.0.10
-			 * @param array Array of excluded product types.
-			 */
-			$excluded_product_types = apply_filters(
-				'pinterest_for_woocommerce_excluded_product_types',
-				array(
-					'grouped',
-					'variable',
-					'subscription',
-					'variable-subscription',
-				)
-			);
+			// Get excluded product types.
+			$excluded_product_types = self::get_unsupported_product_types();
 
 			$types = array_diff( $included_product_types, $excluded_product_types );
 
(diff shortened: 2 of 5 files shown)


sc-ecolasurdo (approved):

JPry:
@alexiglesias31 @sc-ecolasurdo Is this ready for review? Since it's still marked as "Draft", I wasn't sure.

findingsimple:
@JPry [there was a question here](https://github.com/woocommerce/pinterest-for-woocommerce/issues/372)

alexiglesias31:
@jconroy this PR is ready for review. Although, I'd like to place an **important** comment here.
As discussed in the thread inside the issue #372 the idea was to have a more "boolean" way of show/hide Google Category and Condition and also product types on the feed.
As mentioned on the PR description :point_up:, 'simple' and 'variable' products will display Google Condition by default and 'simple', 'external' and 'variation' products will be the ones in the feed by default. The values can be filtered too.

src/Admin/Product/Attributes/AttributesTab.php
```diff
@@ -181,7 +181,7 @@ protected function get_form( WC_Product $product ): AttributesForm {
 	 *
 	 * @return array of WooCommerce product types (e.g. 'simple', 'variable', etc.)
 	 */
-	protected function get_applicable_product_types(): array {
+	public static function get_applicable_product_types(): array {
```
JPry, line 184: Given where this method is used, it might make more sense to have this provide a key-based array, rather than a value-based array. When called in `get_hidden_product_types()`, we can then use `array_diff_key()` to compare. And when this is called from `VariationsAttrubutes::register()`, we can use `array_key_exists()` as a more efficient comparison.

That will require some small tweaks to the `add_tab()` method, but I think it would be beneficial to switch to that approach.

src/Admin/Product/Attributes/AttributesTab.php
```diff
@@ -193,7 +193,7 @@ protected function get_applicable_product_types(): array {
 	 * @return array of WooCommerce product types (e.g. 'subscription', 'variable-subscription', etc.)
 	 */
 	protected function get_hidden_product_types(): array {
-		return apply_filters( 'wc_pinterest_attributes_tab_hidden_product_types', array( 'subscription', 'variable-subscription' ) );
+		return apply_filters( 'wc_pinterest_attributes_tab_hidden_product_types', array_diff( array_keys( wc_get_product_types() ), $this->get_applicable_product_types() ) );
```
JPry: Along with the previous comment, we should split this across multiple lines to make it a bit more readable.

src/FeedGenerator.php
```diff
@@ -561,4 +533,42 @@ protected function get_batch_size(): int {
...
+	 * Return the list of supported product types.
+	 *
+	 * @since x.x.x
+	 *
+	 * @return array
+	 */
+	private function get_supported_product_types(): array {
+		return (array) apply_filters(
+			'pinterest_for_woocommerce_included_product_types',
+			array(
```
JPry, line 550: Given that this is a similar list to the `AttributesTab` class, it might make sense to abstract this into a `trait` that can be included in both classes.

JPry (changes requested):

src/FeedGenerator.php
alexiglesias31: Just a couple of comments before going with this approach.

1. @JPry, we are not doing any diff operation with this array. Are you suggesting to use a `trait` and set the filter name and the default array as class parameters, or to use a key-based array on this method?

2. @jconroy with these changes we are including in the feed only the supported products listed here (simple, external, variation). The customer can add extra product types to the feed filtering that value with a php snippet. This also will break existing installs with custom product types. Just wanna add this extra confirmation, since some customers may be not too happy with using php to add new product types to the feed. I'm not sure if the original though was to exclude only subscriptions from the feed and use the filter array for Google Condition or use this supported array approach for the feed too. What do you think?

JPry: @alexiglesias31 

> we are not doing any diff operation with this array. Are you suggesting to use a `trait` and set the filter name and the default array as class parameters, or to use a key-based array on this method?

Sorry that I wasn't more clear in my comment. It looked to me like it was a similar set of product types between this class and the `AttributesTab` class. The only difference is that this class includes the `external` product type.

My thinking was that this might be a place for combining the two arrays into one location. But with that said, I don't feel very strongly about it, so if it doesn't make sense then we can skip it. I wasn't sure where else these kinds of things might be needed in the future.

I'll let @jconroy or @budzanowski comment on point 2, but if we have documentation in place around the use of the filter, then that should help users have proper expectations around what changing this value will impact. We could also ensure that the place where we pull in the product types can handle custom types gracefully. 

We should also try to handle the deprecated `pinterest_for_woocommerce_excluded_product_types` filter with a proper warning, and if at all possible, an option to use the old behavior of that filter.

budzanowski: > This also will break existing installs with custom product types.

We should have rectified the use of that filter a long time ago when the user base was very small. I think that it is still small enough that we should try to do that. Maybe the fallout will not be big and as @JPry mentions with proper documentation we should minimise the potential hassle after the release. We would need to warn the support people about this beforehand. My guess is that this should be rather painless ( #famouslastwords ) - maybe we will even fix some of the scenarios like https://github.com/woocommerce/pinterest-for-woocommerce/issues/543

> We should also try to handle the deprecated pinterest_for_woocommerce_excluded_product_types filter with a proper warning, and if at all possible, an option to use the old behaviour of that filter.

I agree with that. I think that we can do this with small change to the code organisation: 

```PHP
			/**
			 * Filter excluded product types.
			 *
			 * @since 1.0.10
			 * @param array Array of excluded product types.
			 */
			$excluded_product_types = apply_filters(
				'pinterest_for_woocommerce_excluded_product_types',
				array(
					'grouped',
					'variable',
					'subscription',
					'variable-subscription',
				)
			);

		   $supported_product_types = self::get_supported_product_types();

           $included_product_types = array_diff( $supported_product_types, $excluded_product_types );
```

The construction looks a bit over-engineered but it preserves the old behaviour of filtering out products using `pinterest_for_woocommerce_excluded_product_types` and with the new filter new product types could be added. 

Ultimately the plugin that extends the list of available product types should declare using the `pinterest_for_woocommerce_included_product_types` filter that it supports Pinterest integration and the product makes sense in the Feed file.

@jconroy, @JPry, @alexiglesias31 thoughts?

alexiglesias31: @budzanowski I'm totally agree with your comments.
I think allowing only certain defined product types in Google's feed and terms is the way to go. My main concern about this (and the cause of my warning comment :slightly_smiling_face: ) is users reaction. But to be honest, I think this will save us some headaches in the future, as any kind of custom product can break the feed or be displayed incorrectly.

budzanowski: @alexiglesias31 great. I will let others to chime in on what we are proposing here. @jconroy maybe there are other people we should ask for opinion?

For the meantime please adjust the code preserve the `pinterest_for_woocommerce_excluded_product_types` functionality. We will need to prepare documentation update bc this filter is not covered anywhere.

findingsimple: > Ultimately the plugin that extends the list of available product types should declare using the pinterest_for_woocommerce_included_product_types filter that it supports Pinterest integration and the product makes sense in the Feed file.

Agree

>  I think this will save us some headaches in the future, as any kind of custom product can break the feed or be displayed incorrectly.

Agree.

> We would need to warn the support people about this beforehand. My guess is that this should be rather painless ( #famouslastwords ) - maybe we will even fix some of the scenarios like https://github.com/woocommerce/pinterest-for-woocommerce/issues/543

Lets chat with Sunday on his recommendation for best way to do this. 

Before we ship I think it would be good to give HEs a heads up and [add a section to docs like this one about what products types are supported by default](https://woocommerce.com/document/google-listings-and-ads-faqs/#section-3).

Agree it would be nice to add a warning and give option to use old handling if they want.

I think the fact that we've been included product types that we generally know aren't supported (subscriptions/composite products) I agree it should be better overall for customers.

alexiglesias31: @JPry

> Sorry that I wasn't more clear in my comment. It looked to me like it was a similar set of product types between this class and the AttributesTab class. The only difference is that this class includes the external product type.
My thinking was that this might be a place for combining the two arrays into one location. But with that said, I don't feel very strongly about it, so if it doesn't make sense then we can skip it. I wasn't sure where else these kinds of things might be needed in the future.

Giving a second though to this :thinking:, perhaps external product should be included/excluded in both places. Any thoughts @jconroy? I didn't find external products on GLA extension.


Edit:
However, in order to pack it on a trait...
We need `variation` for the feed and `variable` for the Pinterest tab.

alexiglesias31: @budzanowski 

> The construction looks a bit over-engineered but it preserves the old behaviour of filtering out products using pinterest_for_woocommerce_excluded_product_types and with the new filter new product types could be added.

The filter is back :ok_hand:. Not sure if we should exclude composite products by default too :thinking:.
cc @jconroy

findingsimple: Google merchant center [doesn't allow affiliate products](https://support.google.com/merchants/answer/6363310?hl=en) so [external aren't supported](https://github.com/woocommerce/google-listings-and-ads/blob/e17daaf1526493eeac1febce6b8d60f57b9ffe0f/src/Product/ProductSyncer.php#L256).

I'd follow the same approach for Pinterest

budzanowski:
@sukafia We would like to proceed with this. Sorry for the long.
There are some concerns regarding this that you may help to dismiss or you have more issues and objections with this.

budzanowski:
@darcie thoughts?

darcie:
Hey @budzanowski thanks for tagging me in. Since Woo Happiness hasn't historically supported Pinterest, our HEs won't come into this much to have any input. I think making things clearer for our merchants now is a good idea, even if there's a bit of pain in the transition. I am for this change.

A section in the documentation on what types of products are supported by the extension would be helpful to prevent/answer future questions regarding why a product will not sync. 

I ran through some tickets with potentially invalid product types and found very few examples. Some of these were denied by Pinterest for selling services rather than physical products when not all their products are services or they sell some custom products instead. 
5382794-zen
5457702-zen 
5510334-zen
5576542-zen
5444157-zen 

Excluding unsupported product types should help with merchant approval.

budzanowski:
@darcie Thank you!
I think that we will use this PR + Documentation. It should be OK.

src/FeedGenerator.php
```diff
@@ -561,4 +536,60 @@ protected function get_batch_size(): int {
...
 	}
+
+	/**
+	 * Return the list of supported product types.
+	 *
+	 * @since x.x.x
+	 *
+	 * @return array
+	 */
+	private function get_supported_product_types(): array {
```
message-dimke: My five cents would be to rename the method from `supported` into `included` to correspond to the filter name, also for `excluded` the method is called `excluded` and not `unsupported`, just for the sake of consistency. Other than that, everything seems to be working, I tested it with different types of products: subscriptions, external, composite, simple, variable, etc. Good to go.

Rename `get_supported_product_types` into `get_included_product_types`

message-dimke:
```suggestion
	private function get_included_product_types(): array {
```

src/FeedGenerator.php
```diff
@@ -219,47 +219,22 @@ protected function get_items_for_batch( int $batch_number, array $args ): array
...
-				'pinterest_for_woocommerce_excluded_product_types',
-				array(
-					'grouped',
-					'variable',
-					'subscription',
-					'variable-subscription',
-				)
+			// Get included product types.
+			$included_product_types = array_diff(
+				self::get_supported_product_types(),
```
message-dimke:
```suggestion
				self::get_included_product_types(),
```

message-dimke (changes requested):

message-dimke (approved):
Looks Good! Thank you, @alexiglesias31 !

message-dimke (approved):
Looks good, thanks!

message-dimke (approved):
Approve

message-dimke (approved):

JPry (approved):

PR #498, opened 2022-07-21
Use custom actions of github-actions package to annotate errors of eslint and stylelint

eason9487:

Use the custom actions of `github-actions` package to annotate linting results of `eslint` and `stylelint` in GitHub Actions via their formatters.

### Screenshots:

image

1. Go to this [commit that seeded lint errors](https://github.com/woocommerce/pinterest-for-woocommerce/pull/498/commits/c9a7912b673b1172ea09f6d502cd52de790b3046) to check if the annotations are attached at the right files and lines.
2. Go to the [Lint JavaScript job](https://github.com/woocommerce/pinterest-for-woocommerce/runs/7442337838?check_suite_focus=true#step:5:12) to check the logs. The workflow commands should be in the Annotation commands group and the default `eslint` CLI results should still be printed.
3. Go to the [Lint CSS job](https://github.com/woocommerce/pinterest-for-woocommerce/runs/7442337775?check_suite_focus=true#step:5:17) to check the logs. The workflow commands should be in the Annotation commands group and the default `stylelint` CLI results should still be printed.

### DIFF
--- a/.github/workflows/js-css-linting.yml
+++ b/.github/workflows/js-css-linting.yml
@@ -35,22 +35,11 @@ jobs:
         with:
           node-version-file: ".nvmrc"
 
-      - name: Save code linting report JSON
-        run: npm run lint:js -- --quiet --output-file eslint_report.json --format json
-        # Continue to the next step even if this fails
-        continue-on-error: true
+      - name: Prepare annotation formatter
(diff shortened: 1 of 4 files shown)


rawdreeg (approved):
LGTM! 🚀 

Thanks @eason9487

PR #499, opened 2022-07-21
Show the correct extra info if product sync is not working

### LINKED ISSUE #415: Breaks if a user using 1.0.9 rollbacks to 1.0.8 and then re-updates to 1.0.9
### Describe the bug:
If a user that already updated to the v1.0.9 and is properly connected rollbacks to v1.0.8 and then updates again to v1.0.9 the website verification breaks.
The class `PluginUpdate` (which was added since v1.0.9) adds the parameter `pinterest-for-woocommerce-update-version` to the DB. If the user rollbacks to a previous version (e.g., v1.0.8) and then update again to v1.0.9, since the parameter `pinterest-for-woocommerce-update-version` is already in the DB the migration routines are not called, which are needed to migrate to website migration.
Also, if the user rollbacks to v1.0.8 he will need to re-verify the domain, since domain verification was used till v.1.0.8.
It's not a usual usage of a plugin, but we should handle this problem imho.

### Steps to reproduce:

1. Install the plugin v1.0.9 and do the onboarding.
2. Rollback to v1.0.8.
3. Update again to v1.0.9.
4. The plugin throws the following error `Product Sync is disabled`.

### Screenshot:
image

### Expected behavior:
A user is capable of rollback and update again to v1.0.9 without re-verify the website manually.

### Actual behavior:
A user using v1.0.9 that rollbacks to v1.0.8 and again updates to v1.0.9 needs to re-verify the website manually.

### Additional details:
- Probably a hint to the user to manually re-verify the site will be enough.

alexiglesias31:

Closes #415 .

A rollback from v1.0.9 to v1.0.8 and back to v1.0.9 breaks the site verification (since website verification is used since v1.0.9), so the user needs to re-verify the site directly on the connection tab.

The error `Product sync is disabled - Visit the settings page to enable it.` is displayed if:
- The setting Product sync is disabled.
- The domain is not verified.
- The tracking tag is not set.

However, the same error is displayed for any of the previous conditions :point_up: 
Since preventing all the possible errors with after jumping back in the versions can become difficult, it may be preferable to adjust the error message according to which condition is broken.

- Product sync is disabled -> `Product sync is disabled - Visit the settings page to enable it.`
- The domain is not verified -> `Product sync is disabled - The domain is not verified, visit the connection page to enable it.`
- The tracking tag is not set  -> `Product sync is disabled - The tracking tag is nor configured, visit the connection page to configure it.`

### Screenshots:

Previous error:
image

Adjusted error:
image

2. Rollback to the v1.0.8 and re-verify the site.
3. Update to a version >= 1.0.9.
4. The displayed error should ask to re-verify the site.

### Changelog entry

> Fix - Display correct error if product sync is broken.

### DIFF
--- a/src/ProductSync.php
+++ b/src/ProductSync.php
@@ -144,6 +144,64 @@ class ProductSync {
 	}
 
 	/**
+	 * Get the proper extra_info for the feed status.
+	 *
+	 * @return string
+	 */
+	public static function get_feed_status_extra_info() {
+
+		if ( self::is_product_sync_enabled() ) {
+			return '';
+		}
+
+		if ( ! Pinterest_For_Woocommerce()::is_domain_verified() ) {
+			return sprintf(
+				/* Translators: %1$s The URL of the connection page */
+				__( 'The domain is not verified, visit the <a href="%1$s">connection</a> page to verify it.', 'pinterest-for-woocommerce' ),
+				esc_url(
+					add_query_arg(
+						array(
+							'page' => 'wc-admin',
+							'path' => '/pinterest/connection',
+						),
+						admin_url( 'admin.php' )
+					)
+				)
+			);
+		}
+
+		if ( ! Pinterest_For_Woocommerce()::is_tracking_configured() ) {
+			return sprintf(
+				/* Translators: %1$s The URL of the connection page */
+				__( 'The tracking conversions is disabled, visit the <a href="%1$s">connection</a> page to enable it.', 'pinterest-for-woocommerce' ),
+				esc_url(
+					add_query_arg(
+						array(
+							'page' => 'wc-admin',
+							'path' => '/pinterest/connection',
+						),
+						admin_url( 'admin.php' )
+					)
+				)
+			);
+		}
+
+		return sprintf(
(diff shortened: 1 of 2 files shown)


sc-ecolasurdo (approved):
LGTM!

alexiglesias31:
@JPry just a kindly remainder here :smile:

src/ProductSync.php
```diff
@@ -143,6 +143,64 @@ public static function is_product_sync_enabled() {
...
+	 */
+	public static function get_feed_status_extra_info() {
+
+		if ( self::is_product_sync_enabled() ) {
+			return '';
+		}
+
+		if ( ! Pinterest_For_Woocommerce()::is_domain_verified() ) {
+			return sprintf(
+				/* Translators: %1$s The URL of the connection page */
```
JPry: Minor: for this and the below `translators` comments, you can just use `1:` rather than `%1$s`.

Also note that "translators" should be all lowercase.

JPry (approved):
Only a minor tweak needed, but this otherwise looks good.

findingsimple:
@alexiglesias31 Feedback above

Issue #385, opened 2022-03-08
Support Automatic enhanced match feature apart from email specific enhanced match


adarshakb:
## User story
Enhanced match in the plugin uses only email field https://help.pinterest.com/en/business/article/enhanced-match

We can upgrade this to use [Automatic enhanced match](https://help.pinterest.com/en/business/article/automatic-enhanced-match)

In setup: For automatic enhanced match, in setting we must provide users with higher level setting for disabling/enabling the feature. But also more granular setting like the ability to opt-in/out of name fields, other field types.

Legal: The language in the terms and conditions while connecting need to be updated.

Initial state: We can automatically opt-in the user when they connect for first time similar to enhanced match as long as Legal is clear on this. 

### Is your feature request related to a problem? 
Helps merchants better attribute Pinterest conversions

adarshakb:
Here is the legal copy :
Enable automatic enhanced match
Enabling automatic enhanced match may improve the performance of your campaigns (return on ad spend and cost per action) and can help increase the size of your Pinterest tag audiences. Automatic enhanced match uses hashed information that your customers have already provided to your business to help match more of your website visitors and conversions to people on Pinterest. 
Learn more (---> Link goes to https://www.pinterest.com/_/_/help/business/article/automatic-enhanced-match)

findingsimple:
@adarshakb 

> Legal: The language in the terms and conditions while connecting need to be updated.

Just confirming, that is an update on the Pinterest side right?

Within https://business.pinterest.com/business-terms-of-service/ or https://business.pinterest.com/en/pinterest-advertising-services-agreement ?

findingsimple:
@adarshakb another question - would we replace standard Enhanced Match with AEM? Or have multiple options?

adarshakb:
@jconroy - Since they are different features it would be 2 different options

adarshakb:
link needs to be added/updated onboarding flo
<img width="123" alt="Screen Shot 2022-04-13 at 7 20 40 PM" src="https://user-images.githubusercontent.com/1510328/163301415-a3aa188c-19be-41df-a502-2f1dc39c4d7d.png">
w

findingsimple:
I will update with links to designs and text in the next day. They are complete - just waiting for final confirmation of no feedback.

Would recommend developing using a feature branch with smaller PRs for review.

findingsimple:
[...]

PR #506, opened 2022-08-04
Sync settings from Pinterest to the plugin

### LINKED ISSUE #440: Changes on enhanced match value on Pinterest aren't synced with the plugin
### Describe the bug:
In edge cases where the user disables/enables the enhanced match (EM) setting in Pinterest UI, when they come to the plugin's settings page the change is not synched.

### Steps to reproduce:
2. Go to Pinterest UI -> Ads -> Conversions -> Change the tag's enhanced match setting.
3. Go to the plugin's settings page: The change is not synched with the plugin.

### Expected behavior:
We should find a way to sync the change on Pinterest side with the plugin.

### Actual behavior:
The change on Pinterest side is not synched with the plugin.

alexiglesias31:

Closes #440 .

- Base class in case we need to sync more settings on the future.
- Current synced setting: `aem_enabled` -> 'Enhanced match support'

### Screenshots:

image

2. Go to Pinterest UI -> Ads -> Conversions -> Change the tag's enhanced match setting.
3. Go to the plugin's settings page.
4. The settings are automatically synced from the advertiser account, the user can also manually sync the setting in case of errors.

### Changelog entry

> Add - Support to sync setting changes from Pinterest to the plugin.


sc-ecolasurdo (approved):
LGTM!

alexiglesias31:
@jconroy, just a confirmation on how sync should happen.
With this PR, the Enhanced match support setting is synced from Pinterest to the plugin. If a customer disables/enables the setting in their business account settings (Pinterest Hub) the change is synced to the plugin. But, if the customer changes the setting on the plugin, the value is replaced with the incoming settings during the sync, right?

sc-ecolasurdo (approved):

findingsimple:
> If a customer disables/enables the setting in their business account settings (Pinterest Hub) the change is synced to the plugin. But, if the customer changes the setting on the plugin, the value is replaced with the incoming settings during the sync, right?

I think it probably depends. For clarification.

Syncing on page load ([as described here](https://github.com/woocommerce/pinterest-for-woocommerce/issues/440#issuecomment-1151947075))

On page load the setting page checks the setting at Pinterest, syncs.
If the merchant changes the settings, push the change to Pinterest on save.
Next time the user comes back check the setting at Pinterest, sync.

When using the sync button, I think because on page load we grab the settings from Pinterest it makes sense that the sync button grabs the settings from Pinterest also. Then the merchant can update.

@shivanarrthine is that what you were thinking would happen?

alexiglesias31:
@jconroy 

> If the merchant changes the settings, push the change to Pinterest on save.

I think to recall that at some point there was a discussion about that, where @adarshakb mentioned that there is a kind of policy that avoids that changes on the plugin settings be pushed to Pinterest, but I couldn't find the thread right now. Perhaps he can clarify us on that.

alexiglesias31:
Besides that clarification, the PR is ready for review. And as described above, the behavior right now is the following:
- On Settings tab load, the plugin syncs the settings from Pinterest to the plugin.
- If the customer presses the Sync button the settings are synced again.
- If the customer changes the settings and saves it, the change is not pushed to Pinterest, and when the tab is reloaded, the setting is synced again from Pinterest to the plugin.

adarshakb:
>If the customer changes the settings and saves it, the change is not pushed to Pinterest, and when the tab is reloaded, the setting is synced again from Pinterest to the plugin.

Why do we allow the customer to change the settings? Can this be read-only then? Also please add a help text to explain the behaviour and where the customer can change the value

shivanarrthine:
> On page load the setting page checks the setting at Pinterest, syncs.
If the merchant changes the settings, push the change to Pinterest on save.
Next time the user comes back check the setting at Pinterest, sync.

@jconroy yes, that's my understanding too.

> If the customer changes the settings and saves it, the change is not pushed to Pinterest, and when the tab is reloaded, the setting is synced again from Pinterest to the plugin.

Don't think I fully understand why we do this - it defeats the purpose of surfacing these settings on the plugin if users can't edit them. But if that's the case, we should make it read-only and show a link to the Pinterest ads manager/settings where they can edit it.

findingsimple:
> mentioned that there is a kind of policy that avoids that changes on the plugin settings be pushed to Pinterest but I couldn't find the thread right now. Perhaps he can clarify us on that.

@adarshakb I don't recall this convo, can you confirm if we are allowed to edit the settings?

If not then I 100% agree it is should be disabled and some help text and link need to be added to direct the user to Pinterest 

If we "can" update the settings on save then I think we should do it (cc @alexiglesias31

alexiglesias31:
@jconroy @adarshakb, just shared the link of the discussion thread on the Slack channel :blush:

alexiglesias31:
@jconroy, this PR is ready for review.
There was a misunderstanding here related to the settings to sync. [Here](https://github.com/woocommerce/pinterest-for-woocommerce/pull/506/commits/463e719b24d6dfc3f97a3090a92ef4bef5797e0a#diff-7ca76bacf2369f64dae785d0492c674ff07e9e01313719fc38149b128407ecfeR116) we were syncing the 'Automatic enhanced match' to the 'Enhanced match support' setting on the plugin.
As @adarshakb mentioned on Slack, they are 2 different settings. EM can be handled by the plugin itself without any syncing, and AEM should be readonly on the plugin due to the legal disclaimers.
Conclusion, the issue #440 is not an issue itself, since Enhanced match support is not Automatic enhanced match. I still would let this PR either active or merged after CR, since it will be used when we add Automatic enhanced match support to the plugin #385.

alexiglesias31:
@jconroy I re-based this PR to be merged into [this branch](https://github.com/woocommerce/pinterest-for-woocommerce/tree/feature/support-automatic-enhanced-match). In order to split the feature in smaller PRs.

PR #512, opened 2022-08-16
Progressive back off for merchant connect

### LINKED ISSUE #446: Investigate No relevant meta tag was found exception
Partner connect failed with exception "No relevant meta tag was found when creating merchant"

Over the last week a majority of errors came from
Domain ['https://scarfbank.com/gallery'] verification failed for user scarfbank
Domain ['https://www.bathroombutler.com/us'] verification failed for user bathroombutler

Below domains also had errors but number of errors were lower
Domain ['https://bathroombutdev.wpengine.com/us'] verification failed for user bathroombutler
Domain ['https://jorgephyttasraposo.com'] verification failed for user jaffrayhunter3426

JPry:

Implement progressive backing off of calling the merchant connect API

Closes #446 .

### Additional details:

This adds a caching mechanism to requests made to the `/catalogs/partner/connect/` endpoint. The cache starts off at 10 minutes, and then will double with every call to the API. 

### Changelog entry

> Tweak - Add progressive back-off for querying the merchant connect API

### DIFF
--- a/src/API/Base.php
+++ b/src/API/Base.php
@@ -13,6 +13,7 @@ use Automattic\WooCommerce\Pinterest as Pinterest;
 use Automattic\WooCommerce\Pinterest\Logger as Logger;
 use Automattic\WooCommerce\Pinterest\PinterestApiException as ApiException;
 use \Exception;
+use Pinterest_For_Woocommerce;
 
 
 if ( ! defined( 'ABSPATH' ) ) {
@@ -505,11 +506,21 @@ class Base {
 	 * @return mixed
 	 */
 	public static function update_or_create_merchant( $args ) {
-		return self::make_request(
+		// Get the current delay value, defaulting to 10 minutes.
+		$delay = Pinterest_For_Woocommerce::get_data( 'create_merchant_delay' ) ?? 10 * MINUTE_IN_SECONDS;
+
+		$result = self::make_request(
 			'catalogs/partner/connect/',
 			'POST',
 			$args,
+			'',
+			$delay
 		);
+
+		// Double the delay for next time, to ensure the API is called progressively less often.
+		Pinterest_For_Woocommerce::save_data( 'create_merchant_delay', $delay * 2 );
+
+		return $result;
 	}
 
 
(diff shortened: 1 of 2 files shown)


JPry:
Just a note that I was trying to add some unit tests, but the current codebase structure makes this challenging. Most of the classes are `static` and called directly where needed, which makes the implementation brittle and difficult to swap out for tests.

src/API/Base.php
```diff
@@ -13,6 +13,7 @@
 use Automattic\WooCommerce\Pinterest\Logger as Logger;
 use Automattic\WooCommerce\Pinterest\PinterestApiException as ApiException;
 use \Exception;
+use Pinterest_For_Woocommerce;
```
mikkamp, line 16: Since you corrected it in the other file it seems like we can do the same here and remove the leading backslash from `\Exception`.

mikkamp (commented):
I haven't been able to fully test the changes.

Looking at the code I found it to be a bit of strange pattern to be caching POST requests, as that's generally something that would be used for GET requests. Understandably we aren't going to be repeating the request if it succeeded and also the complete payload is used in generating the hash so any change in payload would bust the cache. Just wanted to mention it in case there is an alternative way to handle a progressive back off, it does seem like it could work in this case though.

Another question I had is it seems the cache is only saved when a valid result is returned from the request. However the original issue mentions:
> Partner connect failed with exception "No relevant meta tag was found when creating merchant"

In the `make_request` function I don't see it populating the cached result within the catch blocks. So if an exception would occur, it would repeat the request right away since the cache would still be empty.

alexiglesias31:
First of all x2 to the comments from @mikkamp.

> Another question I had is it seems the cache is only saved when a valid result is returned from the request. However the original issue mentions

Yes, only success requests are cached, which will cause to repeat the call over and over again on errors.

I had a slightly different approach in mind to tackle this. Using a try catch on `update_or_create_merchant` [method](https://github.com/woocommerce/pinterest-for-woocommerce/pull/512/files#diff-2e9f186919892af02b66ec357eedaf7b792e239ed236e6cfbd2344fdd9387d59L507) and cache only error responses, since we need to get fresh data as much as possible.
We can also use a hash of the payload as part of the transient key or a similar mechanism to bust the cache if the payload changes.

budzanowski:
Closing in favour of #610

PR #533, opened 2022-09-06
Add background billing data check

### LINKED ISSUE #531: Ads Credit Flow: Background Check

In cases where a merchant has not setup billing or is in the process of setting up billing and closed the modal we should check on a regular basis to see if the billing is setup and redeem any available credit

Catalog

## Technical

* Check for billing setup once every 24 hours
* ~~If billing is setup, check eligibility and redeem credit~~
* ~~Catalog overview page will update to reflect whether credit is redeem as per figma designs and https://github.com/woocommerce/pinterest-for-woocommerce/issues/516~~

## Figma link

[https://www.figma.com/file/ojzkxOhHQnIZx4QjsFH5Zy/Native-Pinterest-Integration?node-id=3446%3A59130](https://www.figma.com/file/ojzkxOhHQnIZx4QjsFH5Zy/Native-Pinterest-Integration?node-id=3446%3A59130)

## Dependencies

* https://github.com/woocommerce/pinterest-for-woocommerce/issues/516

## Out of bounds/rabbit holes

n/a

## Event tracking

n/a

alexiglesias31:

Closes #531.

1. Install a build version of this PR.
2. The background check is hooked to the daily heartbeat.
3. You can manually change the billing data on the Pinterest account and manually run the heartbeat action so the `is_billing_setup` parameter inside `account_data` is updated.

### Changelog entry

> Add - Daily background billing data check.


alexiglesias31:
@budzanowski placed the base code, we can polish this when we have more info about the billing data returned by the API.

budzanowski:
For the scheduling please use https://github.com/woocommerce/pinterest-for-woocommerce/blob/develop/src/Heartbeat.php functionality.

You can see its usage here:

https://github.com/woocommerce/pinterest-for-woocommerce/blob/94012b15aad2cab96fff0d44e35e460aa7feaa4c/class-pinterest-for-woocommerce.php#L269

alexiglesias31:
@budzanowski I'm tagging this as ready for review.
However, I'm not 100% convinced about the approach, perhaps I'm missing the big pic tho.
As you suggested we have a recurrent method that checks if billing is set up on a daily basis and the result is stored in a plugin's option.
My main concern would be, if we are not adding any other extra check or verification on the daily event. Perhaps we can wrap up the scheduled action in the same class that you added `Billing`. It may worth to keep them separately if we'd need some extra checks in the future tho.

src/BillingSetupCheck.php
```diff
@@ -0,0 +1,50 @@
...
+ * Pinterest for WooCommerce Ads Billing Setup Check.
+ *
+ * @package     Pinterest_For_WooCommerce/Classes/
+ * @version     1.0.10
+ */
+
+namespace Automattic\WooCommerce\Pinterest;
+
+use Exception;
+use Throwable;
```
budzanowski: Not needed `use` statements.

src/BillingSetupCheck.php
```diff
@@ -0,0 +1,50 @@
...
+	 *
+	 * @since x.x.x
+	 *
+	 * @return mixed
+	 *
+	 * @throws Exception PHP Exception.
+	 */
+	public function handle_billing_setup_check() {
+
+		Pinterest_For_Woocommerce()::save_setting( 'is_billing_setup', Billing::has_billing_set_up() );
```
budzanowski: Best place for this setting would be in the `account_data`, because this is an account specific field. ~~We would need to also add it there in the `update_account_data` procedure~~ during advertiser connection ( I will create a PR for that )

budzanowski (changes requested):

budzanowski:
>My main concern would be, if we are not adding any other extra check or verification on the daily event. Perhaps we can wrap up the scheduled action in the same class that you added Billing. It may worth to keep them separately if we'd need some extra checks in the future tho.

@alexiglesias31 Could you please elaborate a bit about the checks you mention? I don't think I follow 😅

src/BillingSetupCheck.php
budzanowski: I have created #549 with the the functionality mentioned ^. Please review that PR.

In your PR you can just use `	Pinterest_For_Woocommerce()::add_billing_setup_info_to_account_data();` and the status will be added to the `account_data`

alexiglesias31:
@budzanowski, yes, sorry.
The issue's requirements are the following:

> Check for billing setup once every 24 hours
If billing is setup, check eligibility and redeem credit

My question is, do we want to merge all those checks ☝️ (billing setup, eligibility and redeem) into a single class (which would be the class added in this PR, we would need a better name in that case 🙂) or are we only checking every 24h for the billing setup (in this case, we can move the `handle_billing_setup_check` method to the `Billing` class that you already added)

budzanowski:
@alexiglesias31 no this PR will handle only cyclic check, so please only implement that per suggestions in the review.

The remaining items have been moved to separate Issues:
https://github.com/woocommerce/pinterest-for-woocommerce/issues/554
https://github.com/woocommerce/pinterest-for-woocommerce/issues/553

And they will be handled there. So merging this will close #531

budzanowski (approved):
Nice work!

PR #540, opened 2022-09-12
Fix abnormal disconnect state.

### LINKED ISSUE #363: Store stays connected to Pinterest account after disconnecting. Prevents connecting with a different Pinterest account
### Steps to reproduce:

1. Create a new store
2. Disconnect the store
3. connect it to another store

### Expected behavior:
Should allow users to connect to different pinterest account

### Actual behavior:
Prevents connecting with a different Pinterest account

### Additional details:

1Cl8CMFtdVB_DXBkQBHyzFzyTlvYObS5438wgF0CA

budzanowski:

Closes #363 .
I am not 100% sure that the PR fixes all of the possible variants of #363 but for sure it fixes some of them. I still believe that we can close #363 and wait for more ( probably different ) findings that will require separate fixes.

When the site connection fails, for example already cleaning website, the user has no way to disconnect from the site using the plugins UI. In my observation this was because of the missing merchant. The PR detects that state and flushes the system settings so the plugin can assume that it is disconnected. now new connection will be allowed.

1. Please check #363 for testing details.

### Additional details:

More cases can be added in that code area address different API issues.    

### Changelog entry

> Fix - Clear connection options during disconnection error to allow new connection.

### DIFF
--- a/class-pinterest-for-woocommerce.php
+++ b/class-pinterest-for-woocommerce.php
@@ -673,7 +673,8 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 
 				if ( PinterestApiException::MERCHANT_NOT_FOUND === $code ) {
 					Pinterest\Logger::log( esc_html__( 'Trying to disconnect while the merchant ( id ) was not found.', 'pinterest-for-woocommerce' ) );
-					/**
+
+					/*
 					 * This is an abnormal state of the application. Caused probably by issues during the connection process.
 					 * It looks like the best course of actions is to flush the options and assume that we are disconnected.
 					 * This way we restore UI connect functionality and allow merchant to retry.
--- a/class-pinterest-for-woocommerce.php
+++ b/class-pinterest-for-woocommerce.php
@@ -671,7 +671,7 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 			} catch ( PinterestApiException $e ) {
 				$code = $e->get_pinterest_code();
 
-				if ( PinterestApiException::MERCHANT_NOT_FOUND == $code ) {
+				if ( PinterestApiException::MERCHANT_NOT_FOUND === $code ) {
 					Pinterest\Logger::log( esc_html__( 'Trying to disconnect while the merchant ( id ) was not found.', 'pinterest-for-woocommerce' ) );
 					/**
 					 * This is an abnormal state of the application. Caused probably by issues during the connection process.
--- a/class-pinterest-for-woocommerce.php
+++ b/class-pinterest-for-woocommerce.php
@@ -9,6 +9,7 @@
 use Automattic\WooCommerce\Pinterest as Pinterest;
 use Automattic\WooCommerce\Pinterest\Heartbeat;
 use Automattic\WooCommerce\Pinterest\Notes\MarketingNotifications;
+use Automattic\WooCommerce\Pinterest\PinterestApiException;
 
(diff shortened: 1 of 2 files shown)


class-pinterest-for-woocommerce.php
```diff
@@ -667,6 +668,21 @@ public static function disconnect() {
 
 				// At this point we're disconnected.
 				return true;
+			} catch ( PinterestApiException $e ) {
+				$code = $e->get_pinterest_code();
+
+				if ( PinterestApiException::MERCHANT_NOT_FOUND == $code ) {
```
JPry: Do we want a strict comparison here, or are we expecting multiple types of values? `get_pinterest_code()` is supposed to return an `int`.

class-pinterest-for-woocommerce.php
```diff
@@ -667,6 +668,21 @@ public static function disconnect() {
...
+					Pinterest\Logger::log( esc_html__( 'Trying to disconnect while the merchant ( id ) was not found.', 'pinterest-for-woocommerce' ) );
+					/**
```
JPry:
```suggestion
					/*
```

class-pinterest-for-woocommerce.php
```diff
@@ -667,6 +668,21 @@ public static function disconnect() {
 
 				// At this point we're disconnected.
 				return true;
+			} catch ( PinterestApiException $e ) {
+				$code = $e->get_pinterest_code();
+
+				if ( PinterestApiException::MERCHANT_NOT_FOUND == $code ) {
+					Pinterest\Logger::log( esc_html__( 'Trying to disconnect while the merchant ( id ) was not found.', 'pinterest-for-woocommerce' ) );
```
JPry:
```suggestion
					Pinterest\Logger::log( esc_html__( 'Trying to disconnect while the merchant (id) was not found.', 'pinterest-for-woocommerce' ) );
```

src/PinterestApiException.php
```diff
@@ -51,4 +51,12 @@ public function get_pinterest_code() {
 		return $this->pinterest_code;
 	}
 
+	/**
+	 * Merchant not found during the API call. API response message:
+	 * "Sorry! We couldn't find that merchant. Please ensure you have access and a valid merchant id."
+	 *
+	 * @var int MERCHANT_NOT_FOUND Error code for merchant not found API error.
+	 */
+	public const MERCHANT_NOT_FOUND = 650;
```
JPry: Minor: this new constant should be at the top of the file before any methods.

JPry (changes requested):

budzanowski:
[pinterest-for-woocommerce.zip](https://github.com/woocommerce/pinterest-for-woocommerce/files/9568943/pinterest-for-woocommerce.zip)

budzanowski:
[pinterest-for-woocommerce.zip](https://github.com/woocommerce/pinterest-for-woocommerce/files/9569035/pinterest-for-woocommerce.zip)

budzanowski:
[pinterest-for-woocommerce.zip](https://github.com/woocommerce/pinterest-for-woocommerce/files/9569071/pinterest-for-woocommerce.zip)

alexiglesias31:
I'd say x2 to all @JPry 's comments, specially the one with the strict comparison.
Besides that, it LGTM.

class-pinterest-for-woocommerce.php
budzanowski: We want `int` and only int. Fixing.

budzanowski:
@JPry, @alexiglesias31 all done.

JPry (approved):
LGTM 👍🏻

PR #542, opened 2022-09-12
Add method to check if billing is setup for user. 

### LINKED ISSUE #541: Use billing profile endpoint for billing profile.
It was suggested the we should use `/partners/billing_profile` instead of `billing_data` endpoint.

### Story
As a developer I want to be able to check if the users advertiser account has a correct billing setup.

As a merchant I want to be informed if there is a problem with my billing setup.

### Details
The following endpoint should be used for obtaining the billing information:

`advertisers/{$advertiser_id}/partners/billing_profiles`

```GET
Sample response
{
    "status": "success",
    "code": 0,
    "data": {
        "is_billing_setup": true,
        "profiles": [
            {
                "advertiser_id": 12345,
                "billing_type": 3,
                "merchant_id": 213,
                "status": 1
            },        
]
    },
    "message": "ok",
    "endpoint_name": "get_partners_billing_profiles"
}```

The `is_billing_setup` has the status for the billing setup.

### PR DESCRIPTION
**Pull request by budzanowski**

Adds method that uses the `billing_profiles` endpoint to get the user billing status information.

Closes #541.

1. Connect to Pinterest
2. Use the method somewhere in the code and observe the return value

### Additional details:
This should replace the use of billing_data endpoint.

### Changelog entry

> Add - Method for obtaining billing status information.

### DIFF
--- a/src/API/Options.php
+++ b/src/API/Options.php
@@ -44,7 +44,6 @@ class Options extends VendorAPI {
 	 * @return array
 	 */
 	public function get_settings() {
-		$billing = Billing::has_billing_set_up();
 		return array(
 			PINTEREST_FOR_WOOCOMMERCE_OPTION_NAME => Pinterest_For_Woocommerce()::get_settings( true ),
 		);
--- a/src/API/Options.php
+++ b/src/API/Options.php
@@ -8,6 +8,7 @@
 
 namespace Automattic\WooCommerce\Pinterest\API;
 
+use Automattic\WooCommerce\Pinterest\Billing;
 use \WP_Error;
 use \WP_REST_Server;
 use \WP_REST_Request;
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/Billing.php [hunk 1]
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Billing endpoint helper methods.
+ *
+ * @package Pinterest_For_WooCommerce/Classes
+ */
+
+namespace Automattic\WooCommerce\Pinterest;
+
+use Automattic\WooCommerce\Pinterest\API\Base;
+use Automattic\WooCommerce\Pinterest\Logger;
```
JPry: This import is unnecessary, since the namespace for this class is the same as `Logger`.

src/API/Options.php
```diff
@@ -8,6 +8,7 @@
 
 namespace Automattic\WooCommerce\Pinterest\API;
 
+use Automattic\WooCommerce\Pinterest\Billing;
```
JPry: This import isn't used in this file.

JPry (approved):
Looks good, just a couple very minor issues to address before merge.

budzanowski:
Merging with broken CI tests, they have been fixed upstream.

PR #545, opened 2022-09-15
Remove from PHP test matrix WP 5.6 and 5.7, add 5.9

budzanowski:
## Info

Current PHP tests are failing. The reason for that is the latest version of WooCommerce.
We only test with the latests, to be precise we test with trunk. The latest version 6.9 does not support the WP versions before WP 5.8. The 5.8 is the first version that adds the `_wp_to_kebab_case` function required by blocks plugin in the latest WooCommerce.

It would be nice to find a way to test with older than trunk versions of WooCommerce.

1. All CI tests should pass.

### Changelog entry

> Dev - Remove from PHP test matrix WP 5.6 and 5.7, add 5.9

### DIFF
--- a/.github/workflows/php-unit-tests.yml
+++ b/.github/workflows/php-unit-tests.yml
@@ -31,7 +31,7 @@ jobs:
     strategy:
       matrix:
         php: [7.3, 7.4]
-        wp-version: [5.8, latest]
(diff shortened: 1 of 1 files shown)


rawdreeg (approved):
LGTM! 👍🏾

PR #549, opened 2022-09-17
Check billing status during account connection.

budzanowski:
Check connected advertiser billing status. Since the billing status is connected to the advertiser id we can only do that after advertiser gets connected. We also clear the status after advertiser disconnection.

1. Connect to pinters.
2. Check the `pinterest_for_woocommerce` option in wp_options.
3. There should be `is_billing_setup`field in that option:
 a) with value `0` when the advertiser has no billing setup
 b) with value `1` when the advertiser has billing setup ( I think that internal billing also works )

### Changelog entry

> Add - Billing status check during advertiser conncetion.

### DIFF
--- a/class-pinterest-for-woocommerce.php
+++ b/class-pinterest-for-woocommerce.php
@@ -7,6 +7,7 @@
  */
 
 use Automattic\WooCommerce\Pinterest as Pinterest;
+use Automattic\WooCommerce\Pinterest\Billing;
 use Automattic\WooCommerce\Pinterest\Heartbeat;
 use Automattic\WooCommerce\Pinterest\Notes\MarketingNotifications;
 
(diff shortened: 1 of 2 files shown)


class-pinterest-for-woocommerce.php
```diff
@@ -835,6 +836,13 @@ public static function update_account_data() {
 					)
 				);
 
+				/*
+				 * For now we assume that the billing is not setup.
+				 * We will be able to check that only when the advertiser will be connected.
+				 * The billing is tied to advertiser.
+				 */
+				$data['is_billing_setup'] = false;
```
alexiglesias31, line 844: This can be added as a default parameter on the `array_intersect_key` operation.

alexiglesias31 (changes requested):
Just a minor comment.

class-pinterest-for-woocommerce.php
budzanowski: I don't think that it can. Or I don't really understand `array_interesect_key`. 

`array_intersect_key() returns an array containing all the entries of array which have keys that are present in all the arguments.`

`$account_data['data']` does not have the `is_billing_setup` key. Even if we would add it to the list the resulting array would not have that, hence we need to add it manually.

For reference: https://www.php.net/manual/en/function.array-intersect-key.php

alexiglesias31: You are right, the parameter is not present in the account data info. You can ignore my previous comment, sorry.

alexiglesias31 (approved):
LGTM!

PR #561, opened 2022-09-26
Interactions datastore implementation.

budzanowski:
### Information.
To help with implementation ( this and future tasks ) I have build a "interactions" datastore and API endpoint that will be useful for acting on and responding to various user interactions that require storing or pulling data from the server. This way we are separating concerns from our other data-stores that we already have in the application. This will also become useful when we will decide to expand this functionality to other campaigns.


budzanowski:
This is a repurposed copy of datastore we have for other tasks plus a simple API endpoint. Since this blocks some of the other tasks I will go and merge this without a review.

Issue #516, opened 2022-08-17
Ads Credit Flow: Billing Check and Redemption of Ads Credit


findingsimple:
## User story

As a product manager
I want to ensure Merchants set up billing in Pinterest
So that they can redeem Ads Credits

As a merchant
I want to be prompted and directed to set up billing in Pinterest
So that onboarding is as seamless as possible and so that I can begin advertising

## Description

After Merchant has completed onboarding and is considered "connected".

* [Check to see if the Merchant has set up billing on their account](https://github.com/woocommerce/pinterest-for-woocommerce/issues/527)
* [Select relevant coupon based on store currency](https://github.com/woocommerce/pinterest-for-woocommerce/issues/555)
* [Check to see if the merchant is eligible for coupon and if they can redeem - if so do it automattically](https://github.com/woocommerce/pinterest-for-woocommerce/issues/530)

If there is any errors display an error message/toast and log

---

If there is no active credit/coupon - don't do anything

---

If Billing is not setup prompt the user to set up billing

**Note** - [We have another issue for creating the modals](https://github.com/woocommerce/pinterest-for-woocommerce/issues/550)

![Catalog-setup](https://user-images.githubusercontent.com/355014/187667432-f5e6b972-b992-4087-a22b-6f20fb08f099.png)

* Link the merchant to [https://ads.pinterest.com/advertiser/{advertiser_id}/billing/](https://ads.pinterest.com/advertiser/%7Bavertiser_id%7D/billing/) in a new window tab
* If the original window/tab "gets focus" again check if billing is setup and poll every X seconds (open to discussion) to check again
* Once billing is setup, check eligibility and redeem
* Automatically close modal and display success message on catalog page

![Catalog - billing success](https://user-images.githubusercontent.com/355014/187675531-82ef6759-a86c-446a-a557-513a50f06a41.png)

---

If Billing not setup and credits Not Claimed / Or Modal Closed

![Catalog](https://user-images.githubusercontent.com/355014/187676063-353d6d2f-aa41-4458-948d-f6fd57957bef.png)

---

Upon successfully redeeming credit...

![Catalog - success](https://user-images.githubusercontent.com/355014/187667061-0ef1fe51-4e12-44c2-bbf5-45e56c3334f1.png)

Closing the modal will simply display to the merchant the Catalog overview page [with available credit prompt updates ](https://github.com/woocommerce/pinterest-for-woocommerce/issues/518)

## Flow Chart

[...]

PR #565, opened 2022-09-27
Add catalog page ads notice

### LINKED ISSUE #554: Ads Credit Flow, suplement to 531, catalog overview page
This is a supplementary Issue to #531 

In #531 we have been asking for:

* Check for billing setup once every 24 hours
* If billing is setup, check eligibility and redeem credit
* Catalog overview page will update to reflect whether credit is redeem as per figma designs and https://github.com/woocommerce/pinterest-for-woocommerce/issues/516

The two last bullet points will not be handled in #531 for granularity reason.

We are moving:

* Catalog overview page will update to reflect whether credit is redeem as per figma designs and https://github.com/woocommerce/pinterest-for-woocommerce/issues/516

To this Issue.

alexiglesias31:

Closes #554 .

* Catalog overview page will update to reflect whether credit is redeem as per figma designs and https://github.com/woocommerce/pinterest-for-woocommerce/issues/516

1. Install a build version of this branch.
2. Do the onboarding.
3. The credits notice should appear after the onboarding modal is closed.
4. Once the notice is dismissed, it won't be rendered again.


alexiglesias31:
@budzanowski
The current code in the PR only contains the notice component. The logic for showing or hiding the notice is still pending, because I wanna be clear before continue with the development.

From my understanding the flow would be the following:
- The notice should be displayed every time that the onboarding modal is closed.
- Once that billing is setup and credits redeemed (at this stage the notice is [this](https://user-images.githubusercontent.com/355014/187675531-82ef6759-a86c-446a-a557-513a50f06a41.png)) if the user closes the notice we will use the `UserInteractions` API to permanently hide the notice, since credits were already redeemed and the notice is not needed anymore.

Please let me know if I'm missing something here.

budzanowski:
First, small changes to the designs:
If billing is not set we should show:
image

If billing was set but credits not claimed:
image

This is a bit fuzzy to me how we should achieve the logic but I think that it may actually be as simple as just displaying this when the billing is set. We don't need to focus on the `credits not claimed` and check anything. This will be handled by Pinterest side.

So the logic is:
When the user closes the modal show the banner. If no billing show `Spend $15 to get $125 in Pinterest ad credits. To claim the credits add your billing details.` if billing is not set show: `Spend $15 to claim $125 Pinterest ad credits. (Ad credits may take up to 24 hours to be credited to account).` 

If user clicks X dismiss permanently hide the notice.

alexiglesias31:
Thanks @budzanowski 

> This is a bit fuzzy to me how we should achieve the logic but I think that it may actually be as simple as just displaying this when the billing is set. We don't need to focus on the credits not claimed and check anything. This will be handled by Pinterest side.

:+1: 

> If user clicks X dismiss permanently hide the notice.

Double confirmation on this :slightly_smiling_face:. Do we want to show again the notice at some point? I was thinking on the `UserInteractions` API to set a parameter that indicates if the notice needs to be displayed or not. So, my question is: Do we need to re-enable the notice when the client adds the billing details? Or, in the other hand, if the client dismisses the notice before adding the billing details that will permanent hide it. That means that we are not displaying the notice anymore after the billing is set.
Sorry for the fuzzy question :slightly_smiling_face:

budzanowski:
@alexiglesias31 no worries. Ask all you need and how much you need:

>So, my question is: Do we need to re-enable the notice when the client adds the billing details? Or, in the other hand, if the client dismisses the notice before adding the billing details that will permanent hide it. That means that we are not displaying the notice anymore after the billing is set.

I had the same question earlier. The answer is that after dismissal the notice is not shown anymore.

Maybe we will adjust that after testing but for now after dismiss just hide forever ( ever if the user adds billing details ).

budzanowski:
This looks great. One last request: please disable ( don't show ) the ads notice when the campaign is not available. We already have that information and it is already used for Modals so this is now simple. For reference:
https://github.com/woocommerce/pinterest-for-woocommerce/pull/567/files#diff-ef17e4bc3a16add953750367022d90fdbd0e34dabdbabbc80aac306c7ab1a7ccR108

Please make that change, and merge, I am approving the PR.

budzanowski (approved):

Issue #572, opened 2022-09-29
Expand the gating functionality with Ads Countries, Currency and available credits coupons check.


budzanowski:
Currently the gating functionality introduced in [Ads Credit Flow: Feature Flag](https://github.com/woocommerce/pinterest-for-woocommerce/issues/514) only checks if the coupon campaign is globally available. We should expand this functionality to also verify if the merchant is in Ads supported countries and if there are coupons available for his set currency.

This depends on https://github.com/woocommerce/pinterest-for-woocommerce/issues/571

PR #573, opened 2022-09-29
Ads coupon class.

### LINKED ISSUE #571: Ads credit cupon class implementation.
We need to implement a class that does two things:

1. Storing of the coupons codes.
2. Allow retrieval of code/codes. 

The codes storage should be mapped to currencies or countries for which the codes are available.

The retrieval of code should check if the current merchant setup allows for redeeming of ads credits. If everything is fine, return the code. If there is configuration block ( no ads in that country, wrong currency ) return false.

budzanowski:

Adds a simple class for storing ( hardcoded ) coupons.

Closes #571 .

The implementation is based on our assumption on how the coupons work. This may be changed in the future.

1. No testing instructions for now.

### Additional details:
None.

### Changelog entry

>

### DIFF
--- a/src/AdCreditsCoupons.php
+++ b/src/AdCreditsCoupons.php
@@ -1,0 +1,80 @@
+<?php
+/**
+ * Pinterest for WooCommerce Ads Credits Coupons
+ *
+ * @package     Pinterest_For_WooCommerce/Classes/
+ * @version     x.x.x
+ */
+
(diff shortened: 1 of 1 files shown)


budzanowski:
Merging without review. This will be most probably changed in the future but I want to unblock https://github.com/woocommerce/pinterest-for-woocommerce/issues/572 which does not rely on details just output from this class.

PR #587, opened 2022-10-05
Ads User Switching Behavior

### LINKED ISSUE #574: Ads credit user switch behaviour.
This is an exploratory issue: Verify what happens when the merchant disconnects and connects again to a different account. Create a plan needed changes or just implement any necessary adjustments if they are needed. 

The merchant should be able to switch to another account and claim credits again ( if Pinterest API allows for it ).
Current implementation will block this bc the Modals and notices are hidden for ever after dismissal.

alexiglesias31:

Closes #574.

Add user switching behavior

1. Install a build of this PR and do the onboarding.
2. Dismiss the modal.
3. Disconnect the Pinterest account in the plugin.
4. Connect a new Pinterest account.
5. Modals should be displayed too to the new account.

### DIFF
--- a/assets/source/catalog-sync/components/OnboardingModals/index.js
+++ b/assets/source/catalog-sync/components/OnboardingModals/index.js
@@ -25,7 +25,7 @@ const OnboardingModals = ( { onCloseModal } ) => {
 	}
 
 	// Ads campaign modal no error.
-	if ( couponRedeemInfo?.error_code === false ) {
+	if ( ! couponRedeemInfo?.error_code ) {
 		return <OnboardingAdsModal onCloseModal={ onCloseModal } />;
 	}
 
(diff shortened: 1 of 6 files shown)


alexiglesias31:
@budzanowski 
Opened this PR, so we can discuss about the method.
My suggestion is to flush the options that dismisses the modals. When the user disconnects and connects another user the modals will be displayed. 
The "drawback" of this method is that modals will be displayed if the user reconnects the same user. However, imho, this may not be a problem, since only the success modal will be displayed. What are your thoughts on this?

budzanowski:
@alexiglesias31, yes I think that flushing the options is at least the first step and I am OK if we will only have that - switching accounts is a niche thing in general.

If it would turn out to be not sufficient then we could do a bit more. When we switch the user we check the discounts API so we know if the user has previously redeem a coupon. In that case we could skip the modals.

alexiglesias31:
Thanks @budzanowski.

Yes, I'd bet for starting with this approach and we can improve it later. This is ready for review then :slightly_smiling_face:

budzanowski:
@alexiglesias31 I have added two small additional reset features that are used during advertiser change flow.

budzanowski (approved):

alexiglesias31:
Nice catch. Thank you @budzanowski

PR #589, opened 2022-10-06
Hide ads related elements if ads credits is not available

### LINKED ISSUE #577: Ads Credit: connect gating functionality to Ads Credit UI elements.
When a merchant uses Pinterest plugin with Ads feature they should not see the ads nudges and banners if the gating functionality detects that the ads campaign is not possible for current merchant. For example the start screen should not show the ads message for merchants form outside ads allowed countries:

image

The aim of this issue is identification of all ads credits UI elements and not showing them in case Ads credits are not available.

alexiglesias31:

Closes #577

Hide Ads features components if the Ads campaign is not available for current merchant.

2. Connect a merchant with no available ads feature.
3. The ads features components should not be displayed.

### DIFF
--- a/TRACKING.md
+++ b/TRACKING.md
@@ -82,7 +82,7 @@ Clicking on an external documentation link.
 	- with `{ link_id: 'ad-data-terms', context: 'wizard'|'settings' }`
 	- with `{ link_id: 'ad-terms-of-service', context: 'wizard'|'settings' }`
 	- with `{ link_id: 'install-tag', context: 'wizard'|'settings' }`
-- [`WelcomeSection`](assets/source/setup-guide/app/views/LandingPageApp.js#L47) with `{ link_id: 'terms-of-service', context: 'welcome-section' }`
+- [`WelcomeSection`](assets/source/setup-guide/app/views/LandingPageApp.js#L48) with `{ link_id: 'terms-of-service', context: 'welcome-section' }`
 - [TermsAndConditionsModal]( assets/source/setup-guide/app/components/TermsAndConditionsModal.js#L24)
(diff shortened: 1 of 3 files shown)


alexiglesias31:
@budzanowski , please double check if I missed something:
- Card in landing page.
- Gift notice in Setup Account.
- Modals (already implemented)
- Ads Notice (already implemented)

budzanowski:
@alexiglesias31 I don't think that just checking for ads supported countries is enough here. We still want to hide the UI elements when the the campaign is disabled globally or there are no coupons for the merchant. This is why we have `check_if_ads_campaign_is_active` that is used to create the `useSettingsSelect()?.ads_campaign_is_active` which should be the correct parameter to check.

alexiglesias31:
@budzanowski
I think the `ads_campaign_is_active` parameter won't be available (always false) on landing step or the connect step, but perhaps I misunderstood something.

budzanowski:
@alexiglesias31 Please double check but I believe that it should be available right form the start.

We are hooked into `admin_init` so right after start we check with:
```PHP
add_action( 'admin_init', array( Pinterest\AdCredits::class, 'check_if_ads_campaign_is_active' ) );
```
And the checks inside that function don't rely on user connection to the Pinterest account.

For testing you have to alter the return value from `get_is_campaign_active_from_recommendations` since it always returns false ( until we will enable the feature ). But besides that it should work.

alexiglesias31:
Thanks for the clarification @budzanowski,
I missed the behavior of `get_is_campaign_active_from_recommendations`

budzanowski (approved):

budzanowski (approved):

PR #591, opened 2022-10-10
Track redeem credits event

### LINKED ISSUE #578: Ads Credit, track coupon redeem flow.
Coupon redeem operation should be tracked. 

Investigate the best place for coupon redeem action tracking and the format of tracking.
We want minimal info, we just want to check if the users are actually getting value from this flow.

alexiglesias31:

Closes #578.

Track redeem credits event.

2. Do the onboarding and redeem the credits.
3. The redeem credits event should be tracked.

### DIFF
--- a/TRACKING.md
+++ b/TRACKING.md
@@ -82,7 +82,7 @@ Clicking on an external documentation link.
 	- with `{ link_id: 'ad-data-terms', context: 'wizard'|'settings' }`
 	- with `{ link_id: 'ad-terms-of-service', context: 'wizard'|'settings' }`
 	- with `{ link_id: 'install-tag', context: 'wizard'|'settings' }`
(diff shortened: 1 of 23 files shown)


budzanowski:
We want to track bot success and failure scenario.
I think that we could move the tracking from the `redeem_credits` to the `add_redeem_credits_info_to_account_data` method. This way we will be able to also track status and the error parameters.

budzanowski (changes requested):
Please add tracking to all of the results.

alexiglesias31:
In that case, we can track all the `$redeem_information`, or do you think we should unset advertiser_id, username and id?

budzanowski:
@alexiglesias31 that is a good call! We should not track anything that allows us to identify the user. We want to be GDPR safe. Coupon id, redeem status ( success, failure ), error code. We can also drop error message because it is equivalent to the error code.

alexiglesias31:
@budzanowski done

budzanowski (approved):

PR #592, opened 2022-10-10
Add update procedure

### LINKED ISSUE #584: Ads credit, plugin update behaviour.
Investigate and implement how the code should behave when this is deployed on merchant site as a plugin update and not fresh setup. Things to consider:

- modals display behaviour
- redemption behaviour ( will cyclic check be OK or we should maybe try to do the setup during update procedure )
- everything else that comes to mind.

alexiglesias31:

Closes #584.

Integrate the ad credits flow to users that already has the plugin installed.

1. Install the live version of the plugin and do the onboarding.
2. Update to a build version of this PR.
3. The modals should be displayed.

### DIFF
--- a/src/PluginUpdate.php
+++ b/src/PluginUpdate.php
@@ -13,7 +13,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 	exit;
 }
 
-use Automattic\WooCommerce\Pinterest\AdCredits;
+use Automattic\WooCommerce\Pinterest\API\UserInteraction;
 use Exception;
 use Throwable;
 /**
@@ -105,6 +105,7 @@ class PluginUpdate {
 		return array(
 			'domain_verification_migration',
 			'feed_generation_migration',
+			'ads_credits_integration',
 		);
 	}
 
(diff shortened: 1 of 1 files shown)


alexiglesias31:
@budzanowski
This is a first approach for the plugin update.
As far as I understand, after the plugin update the modals will be displayed, since the `ads_modal_dismissed` parameter will not be set.
I'd bet for redeem the credits during the update procedure, but please let me know your thoughts.

budzanowski:
@alexiglesias31 I wonder if we should not hide the modal on update and rely only on banners. 
I will ask the designer.

budzanowski:
I have confirmed that the modals should not be shown for users that are updating the plugin. Only the post-onboarding messages on the Catalog page should be used. This means that we probably can just create dismiss options during the update procedure.

alexiglesias31:
Thanks for the clarification @budzanowski 
Should we keep the credits redeem during the update procedure too? Or just the dismiss options and the redeem can be scheduled normally with the daily heartbeat?

budzanowski:
I think that daily heartbeat is a good idea.

budzanowski (approved):

PR #605, opened 2022-10-18
Do not count inactive credits

### LINKED ISSUE #601: Donot count inactive coupon values while showing ads credit
### Describe the bug:
Screen Shot 2022-10-14 at 5 34 42 PM
Screen Shot 2022-10-14 at 5 34 35 PM

### Steps to reproduce:

1. If user's credit is inactive or expired and not redeemed, dont count it in the 

### Expected behavior:
Only count active credits

### Actual behavior:
Counts all credits

alexiglesias31:

Closes #601

Get only the available active credits.

2. Do the onboarding with an advertiser with active and inactive credits.
3. The available credits should only contain the active credits.

### DIFF
--- a/src/API/Base.php
+++ b/src/API/Base.php
@@ -639,13 +639,9 @@ class Base {
 	 * @return mixed
 	 */
 	public static function get_available_discounts( $advertiser_id ) {
-		return self::make_request(
-			"advertisers/{$advertiser_id}/discounts",
-			'GET',
-			array(
-				'active' => true,
-			),
-			'ads'
-		);
+		$request_url = "advertisers/{$advertiser_id}/discounts";
+		$request_url = add_query_arg( 'active', 'true', $request_url );
+
+		return self::make_request( $request_url, 'GET', array(), 'ads' );
 	}
 }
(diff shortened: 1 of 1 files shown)


src/API/Base.php
```diff
@@ -639,6 +639,13 @@ public static function validate_ads_offer_code( $advertiser_id, $offer_code ) {
...
 	public static function get_available_discounts( $advertiser_id ) {
-		return self::make_request( "advertisers/{$advertiser_id}/discounts", 'GET', array(), 'ads' );
+		return self::make_request(
+			"advertisers/{$advertiser_id}/discounts",
+			'GET',
+			array(
+				'active' => true,
+			),
+			'ads'
+		);
```
budzanowski: I think that this is supposed to be a query arg and not body parameter. Please double check but I think that this should be done like this. 
```suggestion
		$request_url = "advertisers/{$advertiser_id}/discounts";
		$request_url = add_query_arg( 'active', 'true', $request_url );

		return self::make_request( $request_url, 'GET', array(), 'ads' );
```

alexiglesias31: Of course :man_facepalming:. Thanks for the catch

budzanowski (approved):

PR #610, opened 2022-10-20
Back off mechanism for create merchant

### LINKED ISSUE #446: Investigate No relevant meta tag was found exception

alexiglesias31:

Closes #446.

Follow up of the PR #512.
Implement progressive backing off of calling the merchant connect API.

### Additional details:
This adds a caching mechanism to requests made to the /catalogs/partner/connect/ endpoint. The cache starts off at 10 minutes, and then will double with every call to the API.

### Changelog entry

> Tweak - Backoff merchant creation in case of failure.

### DIFF
--- a/src/Merchants.php
+++ b/src/Merchants.php
@@ -145,7 +145,7 @@ class Merchants {
 		$cache_key = PINTEREST_FOR_WOOCOMMERCE_PREFIX . '_request_' . md5( wp_json_encode( $args ) );
 		$cache     = get_transient( $cache_key );
 
-		if ( ! $cache ) {
+		if ( false !== $cache ) {
 			return $cache;
 		}
 
--- a/src/Merchants.php
+++ b/src/Merchants.php
@@ -153,7 +153,7 @@ class Merchants {
 			// The response only contains the merchant id.
 			$response = Base::update_or_create_merchant( $args );
 		} catch ( Throwable $th ) {
-			$delay = Pinterest_For_Woocommerce()::get_data( 'create_merchant_delay' ) ?? 10 * MINUTE_IN_SECONDS;
+			$delay = Pinterest_For_Woocommerce()::get_data( 'create_merchant_delay' ) ?? MINUTE_IN_SECONDS;
 
 			set_transient( $cache_key, $th->getCode(), $delay );
 
--- a/src/Merchants.php
+++ b/src/Merchants.php
@@ -110,7 +110,7 @@ class Merchants {
 		$advertiser = reset( $advertisers['data'] ); // All advertisers assigned to a user share the same merchant_id.
 
 		if ( empty( $advertiser->merchant_id ) ) {
-			throw new Exception( __( 'No merchant returned in the advertiser\'s response.', 'pinterest-for-woocommerce' ), 404 );
+			throw new Exception( __( "No merchant returned in the advertiser's response.", 'pinterest-for-woocommerce' ), 404 );
 		}
 
 		return $advertiser->merchant_id;
--- a/src/Merchants.php
+++ b/src/Merchants.php
@@ -146,7 +146,7 @@ class Merchants {
 		$cache     = get_transient( $cache_key );
 
 		if ( false !== $cache ) {
-			return $cache;
+			throw new Exception( __( 'There was a previous error trying to create or update merchant.', 'pinterest-for-woocommerce' ), (int) $cache );
 		}
 
 		try {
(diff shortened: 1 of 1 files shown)


src/Merchants.php
```diff
@@ -142,15 +142,36 @@ public static function update_or_create_merchant() {
 			'merchant_name'    => $merchant_name,
 		);
 
-		// The response only contains the merchant id.
-		$response = API\Base::update_or_create_merchant( $args );
+		$cache_key = PINTEREST_FOR_WOOCOMMERCE_PREFIX . '_request_' . md5( wp_json_encode( $args ) );
+		$cache     = get_transient( $cache_key );
+
+		if ( ! $cache ) {
```
budzanowski: I think that this is wrong. If transient expires then `get_transient` will return `false` and we will get inside the `if` resulting in returned `false` value. We should use `get_transient` API directly and check for `false` explicitly:
```suggestion
		if ( false !== $cache ) {
```

alexiglesias31: Of course :man_facepalming: 
Yes, that was the idea. Nice catch.

src/Merchants.php
```diff
@@ -142,15 +142,36 @@ public static function update_or_create_merchant() {
 			'merchant_name'    => $merchant_name,
 		);
 
-		// The response only contains the merchant id.
-		$response = API\Base::update_or_create_merchant( $args );
+		$cache_key = PINTEREST_FOR_WOOCOMMERCE_PREFIX . '_request_' . md5( wp_json_encode( $args ) );
+		$cache     = get_transient( $cache_key );
+
+		if ( false !== $cache ) {
+			return $cache;
```
budzanowski: If we have cache value then it means that we are delaying next fetch. We can't return something here because the rest of the code will expect a correct value. We need to throw an exception to signalise to the rest of the App what is going on. We should throw it with the original code and a message that this is a backoff attempt.

src/Merchants.php
```diff
@@ -142,15 +142,36 @@ public static function update_or_create_merchant() {
...
+		}
+
+		try {
+			// The response only contains the merchant id.
+			$response = Base::update_or_create_merchant( $args );
+		} catch ( Throwable $th ) {
+			$delay = Pinterest_For_Woocommerce()::get_data( 'create_merchant_delay' ) ?? 10 * MINUTE_IN_SECONDS;
+
+			set_transient( $cache_key, $response, $delay );
```
budzanowski: If there was an Exception then the `$response` is not defined. The only thing that we can store here is `$th->getCode`. We will be able to use it later in the cache check on line 149 for throwing an exception.

src/Merchants.php
```diff
@@ -142,15 +142,36 @@ public static function update_or_create_merchant() {
...
+
+			// Double the delay.
+			Pinterest_For_Woocommerce()::save_data( 'create_merchant_delay', $delay * 2 );
```
budzanowski: To prevent crazy values we should use max allowed value.
```suggestion
			Pinterest_For_Woocommerce()::save_data( 'create_merchant_delay', max( $delay * 2, HOUR_IN_SECONDS * 6 );
```

alexiglesias31:
@budzanowski thanks for the feedback. Just updated the changes according to your comments.

src/Merchants.php
```diff
@@ -142,15 +142,36 @@ public static function update_or_create_merchant() {
...
+
+		if ( false !== $cache ) {
+			throw new Exception( __( 'There was a previous error trying to create or update merchant.', 'pinterest-for-woocommerce' ), (int) $cache );
+		}
+
+		try {
+			// The response only contains the merchant id.
+			$response = Base::update_or_create_merchant( $args );
+		} catch ( Throwable $th ) {
+			$delay = Pinterest_For_Woocommerce()::get_data( 'create_merchant_delay' ) ?? 10 * MINUTE_IN_SECONDS;
```
budzanowski: I did tested it and it works. Though I think that 10 minutes is much too aggressive. I have counted a multiple retries in a matter of couple of seconds when I have disabled the delay, if the first one would be just a fluke then the whole site would be blocked for the 10 minutes. I think that we can start with 1 minute. Getting to 10 is just 4 failed retries.

budzanowski (approved):
Left one comment but otherwise this is good to go.

findingsimple:
@alexiglesias31 are your PR's / work still being reviewed/QA'd by a fellow sau/cal team member before getting to us?

I haven't seen anyone else around for a little bit or drafts.

alexiglesias31:
Hi @jconroy,
We were opening directly the PRs on the tasks related to the Ads Credits flow (tested by me) and this one (follow up of a PR opened by Jpry).
We are moving back to the usual flow now.

src/Merchants.php
alexiglesias31: Done :white_check_mark:

PR #612, opened 2022-10-21
Display both, remaining discounts and future discounts

### LINKED ISSUE #611: Ads Credits, future and marketing offer discounts available at the same time.
A merchant may have different coupons/discounts available at the same time.
Application should correctly display the UI elements for each combination of available discounts.

In case that both future credits ( related to the plugin coupon ) value is present and there are already discounts with available credits the following two UI elements should be displayed at the same time:
image
image

alexiglesias31:

Closes #611.

Display both, remaining and future discounts.

### Additional details:

In addition: Fix small issue with the active campaigns transient.


budzanowski:
I have made a small addition to the code. Since the transient stores strings we need to do some casting. Otherwise we may experience strange behaviour.

PR #613, opened 2022-10-24
Declare HPOS compatibility.

budzanowski:
Add snippet that declares the plugin to be HPOS compatible.

### DIFF
--- a/pinterest-for-woocommerce.php
+++ b/pinterest-for-woocommerce.php
@@ -46,6 +46,16 @@ if ( ! defined( 'ABSPATH' ) ) {
 define( 'PINTEREST_FOR_WOOCOMMERCE_PLUGIN_FILE', __FILE__ );
 define( 'PINTEREST_FOR_WOOCOMMERCE_VERSION', '1.2.4' ); // WRCS: DEFINED_VERSION.
 
+// HPOS compatibility declaration.
+add_action(
+	'before_woocommerce_init',
+	function() {
+		if ( class_exists( '\Automattic\WooCommerce\Utilities\FeaturesUtil' ) ) {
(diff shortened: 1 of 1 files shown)


pinterest-for-woocommerce.php
```diff
@@ -46,6 +46,16 @@
 define( 'PINTEREST_FOR_WOOCOMMERCE_PLUGIN_FILE', __FILE__ );
 define( 'PINTEREST_FOR_WOOCOMMERCE_VERSION', '1.2.4' ); // WRCS: DEFINED_VERSION.
 
+// HPOS compatibility declaration.
+add_action(
+	'before_woocommerce_init',
+	function() {
+		if ( class_exists( '\Automattic\WooCommerce\Utilities\FeaturesUtil' ) ) {
+			\Automattic\WooCommerce\Utilities\FeaturesUtil::declare_compatibility( 'custom_order_tables', 'pinterest-for-woocommerce/pinterest-for-woocommerce.php', true );
+		}
```
JPry: While the plugin code doesn't currently use imports, I think that new code should properly use imports rather than specifying the FQDN.

Also, we can use the `::class` constant with the import. This constant is safe to use whether the class exists or not. So I propose this block looks like the following, with the addition of `use Automattic\WooCommerce\Utilities\FeaturesUtil;` at the beginning of the file.

```suggestion
		if ( class_exists( FeaturesUtil::class ) ) {
			FeaturesUtil::declare_compatibility(
				'custom_order_tables',
				plugin_basename( PINTEREST_FOR_WOOCOMMERCE_PLUGIN_FILE )
			);
		}
```

JPry (changes requested):

PR #632, opened 2022-11-04
Support automatic enhanced match support

### LINKED ISSUE #385: Support Automatic enhanced match feature apart from email specific enhanced match

alexiglesias31:

Closes #385.

Support automatic enhanced match support.

### Screenshots:

image

image

2. The Automatic Enhanced Match legals will be available on the setup advertiser and tag setup step.
3. Disable the Automatic Enhanced Match setting in the Pinterest account. The setting will be enabled on the Pinterest account when the user presses the Complete Setup button during the onboarding.
4. The setting Automatic Enhanced Match will be read only in the plugin and synced on every tab load or manually click on the Sync button.

### Changelog entry

> Add - Automatic enhanced match support.


alexiglesias31:
@jconroy @adarshakb

I have a question related to the legal text:

> [Automatic Enhanced Match](https://www.pinterest.com/_/_/help/business/article/automatic-enhanced-match) is enabled by default to match more of your website visitors and conversions to people on Pinterest. You can manage this in Settings.

As discussed before, the setting is read-only. Is the setting enabled by default for all new merchants? 
Because the setting is synced on every tab load, if the setting is disabled by default for new merchants (or existing merchants as well) it can be confusing to claim that aem is enabled by default and display the setting disabled when the user visits the Settings page.

alexiglesias31:
@jconroy a kind reminder in case you missed my previous comment :point_up:

adarshakb:
The copy can read "Pinterest settings" and link to conversions tag manager https://ads.pinterest.com/advertiser/{{advertiserId}}/conversions/tag/

alexiglesias31:
@adarshakb,
Another question about the legal copy:
1. What about the step when a new merchant still doesn't have an advertiser? Is it ok to strong 'Pinterest Settings'?
image
2. Is AEM enabled by default on the Pinterest side? Because we let the user know that AEM is enabled by default, however, if the AEM setting is disabled for new merchants, once the user reaches the Settings page, the setting will be synced from Pinterest and disabled. I just want to be sure this is the expected behavior or if we need to modify something.

budzanowski:
@adarshakb ^ could you please provide any insight to @alexiglesias31 question?

adarshakb:
I think the feature here is to enable it for the user from Woo side through the API. We should link to appropriate legal notices (I see you have some links)  and they should accept the terms and conditions (In I accept there must be links too? UX/Legal can say)

sc-ecolasurdo (approved):

alexiglesias31:
Hey @message-dimke this is ready for review.

message-dimke:
Hey, @alexiglesias31 !

When I activate AEH at Pinterest side it checks all the boxes.

Pinterest

When I do it through the extension it does only `Email`. Is it ok?

Pinterest

alexiglesias31:
Nice call @message-dimke :thinking:
This feature has been around for a while now. At that point there was only one option available through the API (`aem_enabled`). There are more parameters available now, we could set all the options through the API call as you mentioned. However, I think that needs confirmation from @adarshakb.

message-dimke:
@adarshakb ,
could you advice which behaviour is preferred, enable all the options or the email only?
ref https://github.com/woocommerce/pinterest-for-woocommerce/pull/632#issuecomment-1436024121

ivanrios:
Please send all parameters in order to have them all enabled.

**aem_enabled** - Whether Automatic Enhanced Match email is enabled
**aem_fnln_enabled** - Whether Automatic Enhanced Match name is enabled.
**aem_ph_enabled** -  Whether Automatic Enhanced Match phone is enabled.
**aem_ge_enabled** - Whether Automatic Enhanced Match gender is enabled.
**aem_db_enabled** - Whether Automatic Enhanced Match birthdate is enabled.
**aem_loc_enabled** - Whether Automatic Enhanced Match location is enabled.

alexiglesias31:
Hey @message-dimke just added the extra options to the API call.

message-dimke (approved):
Now, when I activate AEM, everything is ON as expected!

Pinterest

PR #633, opened 2022-11-12
Update Tracking.php

alewolf:

I am the developer of the [Pixel Manager for WooCommerce](https://wordpress.org/plugins/woocommerce-google-adwords-conversion-tracking-tag/). The Pixel Manager pro version also offers event tracking for Pinterest. 

We have customers who want to use the Pinterest for WooCommerce plugin and the Pixel Manager for WooCommerce together. So essentially, they need Pinterest for WooCommerce for the feed and pins and the Pixel Manager for WooCommerce to handle the tracking (which has a few advantages over using the native implementation of Pinterest for WooCommerce. Let me know if you need more details on that.). 

In such a case, it would be great to have a filter in the Pinterest for WooCommerce plugin, which the Pixel Manager can use to disable the Pinterest for WooCommerce event tracking and take over. The filter will only be used if tracking for Pinterest gets enabled in the Pixel Manager. 

This pull request adds such a filter to the Pinterest for WooCommerce plugin.

### DIFF
--- a/src/Tracking.php
+++ b/src/Tracking.php
@@ -350,8 +350,15 @@ JS;
 	 */
 	private static function tracking_enabled() {
 		
-		if ( apply_filters( 'woocommerce_pinterest_disable_tracking', false ) ) {
-		  return false; 
+		/**
+		 * Allow third party plugins to disable the tracking pixel. 
+		 *
+		 * @since 1.2.6
+		 *
+		 * @param bool $is_disable Tracking is enabled if false, and disabled if true.
+		 */
+		if ( apply_filters('woocommerce_pinterest_disable_tracking', false) ) {
+			return false;
 		}
 
 		if ( ! Pinterest_For_Woocommerce()::get_setting( 'track_conversions' ) || ! self::get_active_tag() ) {
--- a/src/Tracking.php
+++ b/src/Tracking.php
@@ -349,6 +349,10 @@ JS;
 	 * @return boolean
 	 */
 	private static function tracking_enabled() {
+		
+		if ( apply_filters( 'woocommerce_pinterest_disable_tracking', false ) ) {
+		  return false; 
+		}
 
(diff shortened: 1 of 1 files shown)


budzanowski:
@alewolf Hi! Thank you for this PR

We will discuss this contribution internally. Before we can proceed please make sure that the Wordpress/WooCommerce PHP coding standards are applied and the documentation block is added.

For reference: https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#4-hooks-actions-and-filters

alewolf:
Hi @budzanowski ! 

Ok. I've added the documentation block. 

It also passes the PHPCS checks for WooCommerce.

budzanowski:
Thank you!

OK, we will take it from here.
I don't see a problem with the PR but I need to consult the rest of the team to double check. You should see if this will get merged or I will ping you directly if we will not decide to use it.

Please give us a review. Your feedback is welcome - we want to improve as much as possible.

Cheers!

alewolf:
Ok. 

There is such a filter for [Facebook for WooCommerce](https://github.com/woocommerce/facebook-for-woocommerce/blob/00186681a473470fb4a2fcda6db3567550bc1c22/facebook-commerce-events-tracker.php#L92) and the [Google Listing and Ads](https://github.com/woocommerce/google-listings-and-ads/blob/d4b2dfd8751ebadb096f2a974de0202d08c048d0/src/Google/GlobalSiteTag.php#L463) plugins. 

IMO it make sense to have it in Pinterest for WooCommerce too.

budzanowski:
This will be merged in but with a comment that the filter is not guaranteed to be there in the future.

alewolf:
Thank you for this!

I wonder, is that just a generic comment, or are there any specific ideas that would require removing the filter in the future?

budzanowski:
We may decide that the tracking will be implemented in a different way. This would make the filter, in its current form at least, unusable.

alewolf:
Thanks for that info. I'll keep an eye on it then :eyes:

PR #639, opened 2022-11-21
Use min instead of max

alexiglesias31:

Fix an error calculating the backoff time.


alexiglesias31:
@budzanowski, I think we should use min here instead of max.

budzanowski (approved):

PR #649, opened 2022-12-02
Tweak add to cart tracking data

### LINKED ISSUE #548: Tweak add to cart tracking event on AJAX
### Describe the bug:

There are some customers complaining about the add to cart tracked event on single product pages using AJAX. The plugin is not reading properly the product properties (id, value, quantity).
There is margin for an enhancement on the logic around add to cart AJAX hook.

### Steps to reproduce:
1. Install the plugin.
2. Add a product to the cart from the single product page with AJAX.
3. The tracked event only contains the currency but no other product info.

### Expected behavior:
The tracked add to cart event should contain all the correct information.

### Actual behavior:
The tracked add to cart event only contains the store's currency.

alexiglesias31:

Closes #548.

There were some inconsistencies with the data on the add to cart event.
- The data is broken if the site uses AJAX to add products to the cart on the single product pages (a plugin like [this](https://wordpress.org/plugins/woo-ajax-add-to-cart/)).
- The prices were not correct on add to cart event (archive page) and WooCommerce setup with show prices excluding taxes.

1. Install a build version of this PR and setup the plugin.
2. Install [this](https://wordpress.org/plugins/woo-ajax-add-to-cart/) plugin.
3. Add to cart a product from the single product page (the event's data should be correct).
4. Configure WooCommerce to display prices in the shop page excluding taxes.
5.  Add to cart a product from the shop page (the price in the event's data should be the same than the displayed price).

### Changelog entry

> Fix - Inconsistencies with the add to cart tracking event.

### DIFF
--- a/src/Tracking.php
+++ b/src/Tracking.php
@@ -244,7 +244,7 @@ class Tracking {
 		$object_id = empty( $variation_id ) ? $product_id : $variation_id;
 		$product   = wc_get_product( $object_id );
 
-		$product_price = self::get_product_price_with_taxes( $product );
+		$product_price = self::get_product_display_price( $product );
 
 		self::add_event(
 			'AddToCart',
@@ -285,7 +285,7 @@ class Tracking {
 
 			$product = $order_item->get_product();
 
-			$product_price = self::get_product_price_with_taxes( $product );
+			$product_price = self::get_product_display_price( $product );
 
 			$terms      = wc_get_object_terms( $product->get_id(), 'product_cat' );
 			$categories = ! empty( $terms ) ? wp_list_pluck( $terms, 'name' ) : array();
@@ -375,7 +375,7 @@ JS;
 
 		$product_id    = $product->get_id();
 		$product_name  = $product->get_name();
-		$product_price = self::get_product_price_with_taxes( $product );
+		$product_price = self::get_product_display_price( $product );
 
 		$wc_currency = get_woocommerce_currency();
 		$tracking    = <<< JS
@@ -658,7 +658,7 @@ JS;
 	private static function filter_add_to_cart_attributes( array $args, WC_Product $product ) {
 		$attributes = array(
 			'data-product_name' => $product->get_name(),
-			'data-price'        => self::get_product_price_with_taxes( $product ),
+			'data-price'        => self::get_product_display_price( $product ),
 		);
 
 		$args['attributes'] = array_merge( $args['attributes'], $attributes );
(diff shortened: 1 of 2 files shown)


sc-ecolasurdo (approved):

src/Tracking.php
```diff
@@ -243,7 +244,7 @@ public static function hook_add_to_cart_event( $cart_item_key, $product_id, $qua
 		$object_id = empty( $variation_id ) ? $product_id : $variation_id;
 		$product   = wc_get_product( $object_id );
 
-		$product_price = WC()->cart->display_prices_including_tax() ? wc_get_price_including_tax( $product ) : wc_get_price_excluding_tax( $product );
+		$product_price = self::get_product_price_with_taxes( $product );
```
message-dimke: I suggest renaming because the method can return both prices with and w/o taxes.

```suggestion
		$product_price = self::get_product_price( $product );
```

src/Tracking.php
```diff
@@ -284,7 +285,7 @@ public static function hook_checkout_event( $order_id ) {
 
 			$product = $order_item->get_product();
 
-			$product_price = WC()->cart->display_prices_including_tax() ? wc_get_price_including_tax( $product ) : wc_get_price_excluding_tax( $product );
+			$product_price = self::get_product_price_with_taxes( $product );
```
message-dimke:
```suggestion
			$product_price = self::get_product_price( $product );
```

src/Tracking.php
```diff
@@ -665,4 +719,16 @@ public static function get_third_party_installed_tags() {
...
 	}
+
+	/**
+	 * Get product's price including/excluding tax.
+	 *
+	 * @param WC_Product $product The product object.
+	 *
+	 * @return string
+	 */
+	protected static function get_product_price_with_taxes( $product ) {
```
message-dimke:
```suggestion
	protected static function get_product_price( $product ) {
```

alexiglesias31:
Thanks @message-dimke, changes applied.

message-dimke:
Also I suggest removing all the javascript events not to maintain both variants with and w/o attributes.

How about removing this

https://github.com/woocommerce/pinterest-for-woocommerce/blob/345553dd1f681186487882e7a3ebd95e6e625d5b/src/Tracking.php#L101-L112

and

https://github.com/woocommerce/pinterest-for-woocommerce/blob/345553dd1f681186487882e7a3ebd95e6e625d5b/src/Tracking.php#L237-L241

and all the related methods. This way we will not have to maintain JS snippets and will send add to cart events from the backend `hook_add_to_cart_event`.

message-dimke:
From Slack by @alexiglesias31 :

> The JS snippets are used to track the add to cart event when:
>
> - Enable AJAX add to cart buttons on archives is enabled;
> - Redirect to cart page after successful addition is disabled;
>
> The reason is to track the add to cart event immediately and not on the next page refresh (which may lead to potential missed tracked events due to cart abandonment for example).

message-dimke (approved):
Hey, @alexiglesias31 !

Ready to be merged, Thanks!

PR #650, opened 2022-12-05
Fix multiple registered feeds on site language change.

### LINKED ISSUE #643: The feed profile is duplicated after changing the Wordpress language.
### Describe the bug:

Every time a different language is set in the Wordpress settings, a new feed profile is created on Pinterest leaving the original one active.

Screen Shot 2022-11-28 at 12 19 12

### Steps to reproduce:
1.  Log into your account
2.  Go to Settings -> General
3.  Change site Language
3.  Wait 2 hours
4.  Go to Pinterest Ads portal and go to Ads -> Catalogs.
5. You will see more than 1 catalog record

### Expected behavior:
Since there is no multifeed feature, each Pinterest connection should only have 1 active feed profile.

### Actual behavior:
A new feed profile is generated if the user changes the Wordpress language.

budzanowski:

When the user changes the locale a new feed configuration is created. It now reflects the new locale ( language ). We can't delete the old feed configuration ( API limitation ) but we can disable it. 

Closes #643  .

- Correctly identify the registered feed configuration as local.
- Make sure that the additional old feed configurations are disabled.

### TODO:
- [x] cache invalidation.
- [x] remove TODOs comments.
- [x] feed enable feature when feed configuration detected but not enabled.
- [x] comments.

**Using develop**

( this is a copy from #643 )
1. Log into your account
2. Go to Settings -> General
3. Change Site Language
4. Wait 2 hours
5. Go to the Pinterest Ads portal and go to Ads -> Catalogs.
6. You will see more than 1 catalog record

**Switch to this branch**
Wait for some time ( < 10 minutes ) The automation should clean things up.
You should receive an email about that a data source has been deleted - Pinterest is working on making that message less scary.
If you now check the feed records in Pinterest UI there should be only one active.

### Additional details:

The code in this PR checks for all of the feed configurations and acts only on the ones that are registered as WooCommerce integrations. It also compares the URLs. This means that the cleanup procedure only works for multiple feed files that are coming from the same website. Connecting multiple ( different ) sites is still supported (whether that is good or bad is a different discussion and more a business decision )

### Changelog entry

> Fix - Prevent multiple active feed files.

### DIFF
--- a/src/FeedRegistration.php
+++ b/src/FeedRegistration.php
@@ -134,8 +134,11 @@ class FeedRegistration {
 
 		Pinterest_For_Woocommerce()::save_data( 'feed_registered', $feed_id );
 
-		self::feed_enable_status_maintenance( $merchant_id, $feed_id );
+		if ( ! $feed_id ) {
+			return false;
+		}
 
+		self::feed_enable_status_maintenance( $merchant_id, $feed_id );
 		return true;
 	}
 
--- a/src/FeedRegistration.php
+++ b/src/FeedRegistration.php
@@ -129,7 +129,7 @@ class FeedRegistration {
 		$merchant_id   = $merchant['data']->id;
 		$local_feed_id = self::get_locally_stored_registered_feed_id();
 
-		if ( ! $local_feed_id || ! Feeds::is_local_feed_registered( $merchant_id ) ) {
+		if ( ! $local_feed_id || ! Feeds::match_local_feed_configuration_to_registered_feeds( $merchant_id ) ) {
 
 			$response = Merchants::update_or_create_merchant();
 
--- a/src/FeedRegistration.php
+++ b/src/FeedRegistration.php
@@ -131,7 +131,7 @@ class FeedRegistration {
 			 * Update feed if we don't have a feed_id saved or if local feed is not properly registered.
 			 * for cases where the already existed in the API.
 			 */
-			$registered = self::get_registered_feed_id();
+			$registered = self::get_locally_stored_registered_feed_id();
 
 			if ( ! $registered || ! Feeds::is_local_feed_registered( $merchant['data']->id ) ) {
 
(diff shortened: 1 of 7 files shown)


budzanowski:
This is a Draft state since there are some issues with the API access and thus this PR is not testable.

src/FeedRegistration.php
```diff
@@ -112,37 +112,128 @@ public function handle_feed_registration() {
...
+		if ( ! $merchant_id ) {
+			return false;
+		}
+
+		$feed_id = self::get_locally_stored_registered_feed_id();
+
+		// If the feed is not registered, try to match it to a registered feed.
+		if ( ! $feed_id ) {
+			$feed_id = Feeds::match_local_feed_configuration_to_registered_feeds( $merchant_id );
+		}
```
message-dimke: If I have a locally registered feed option, then it will never search for a locale different feed and update it unless the `feed_registered` option is emptied.

My suggestion is let it search for a matching feed and create a new one or reactivate an old one if any.

```suggestion
		$feed_id = Feeds::match_local_feed_configuration_to_registered_feeds( $merchant_id );
```

Also there is a problem if new locale does not match any of Pinterest locales, but I think it is fixed with another PR. So for example if I set my locale to New Zealand then it does not update and just quits with an Exception (the exception has no meaningful message and just says something like Unsupported parameter). I assume this is a locale problem since I haven't found New Zealand from the list of supported locales by Pinterest and other locations like Australia, United Kingdom or US those do work well. I'll check this case when reviewing locale mapping PR once again.

message-dimke (changes requested):

src/FeedRegistration.php
budzanowski: > My suggestion is let it search for a matching feed and create a new one or reactivate an old one if any.

This is how it was actually done before. I think that I have tried to bee a bit too smart. Great catch.

> Also there is a problem if new locale does not match any of Pinterest locales, but I think it is fixed with another PR.
Yes this is handled by https://github.com/woocommerce/pinterest-for-woocommerce/pull/687 so we don't have to think about this here. But good point nevertheless.

budzanowski:
@message-dimke update made.

message-dimke (approved):
Good to go! Thank you, @budzanowski !

PR #660, opened 2023-01-12
Add additional parameters into WC tracker for Pinterest

message-dimke:

- Adding additional WC tracker data: feed generation time and products count, store connection status and product sync status;
- Refactoring feed file operations our of feed generator to unit test feed generation failures and proper tracker snapshot properties update;
- Adding more unit tests on involved classes;

#### Basic test
1. Navigate to `WooCommerce > Status > Scheduled Actions > Pending`.
2. Run `pinterest-for-woocommerce-start-feed-generation` action.
3. Using WP CLI run a command `wp wc tracker snapshot`4. 
4. Check the comment output for 
```
... "store":{"connected":"yes","actively_syncing":"yes"},"feed":{"generation_time":111,"products_count":20} ...
```
5. `store` and `feed` objects must be present (they were not before, added within current PR) and have properties with values of the store connection status and sync status as well as recent feed generation details: time it took to generate the feed and the number of products went into the feed.
#### Test when previous results are displayed if feed generation is in progress
-1. Add new simple product into your catalog.
0. Put exit code above the line 173 to emulate feed generation job in progress (like it is having hard times generating 1100500 products)

https://github.com/woocommerce/pinterest-for-woocommerce/blob/0f0da243ff1ceec15b3f90fe029a0bedc9a9566b/src/FeedGenerator.php#L166-L173

**p.s.** _you can also add `sleep(3);` for example, because feed generation may run less than a second an wall time can have 0. value, by adding `sleep(3);` you know that wall time must be at least 3 and check it._

```
$this->feed_file_operations->rename_temporary_feed_files_to_final();
ProductFeedStatus::set(
	array(
		'status'                                                     => 'generated',
		ProductFeedStatus::PROP_FEED_GENERATION_RECENT_PRODUCT_COUNT => ProductFeedStatus::get()['product_count'],
	)
);
exit;
sleep(3);
ProductFeedStatus::set_feed_file_generation_time( time() );
```
1. Navigate to `WooCommerce > Status > Scheduled Actions > Pending`.
2. Run `pinterest-for-woocommerce-start-feed-generation` action.
3. Using WP CLI run a command `wp wc tracker snapshot`
4. Check the comment output for 
```
... "store":{"connected":"yes","actively_syncing":"yes"},"feed":{"generation_time":111,"products_count":20} ...
```
5. `store` and `feed` objects must be present and have the values from the previous **Basic test** case.
6. Go back to the code and remove `exit;`
7. Navigate to `WooCommerce > Status > Scheduled Actions > In progress` and remove in progress action on feed generation chain.
8. Navigate to `WooCommerce > Status > Scheduled Actions > Pending`.
9. Run `pinterest-for-woocommerce-start-feed-generation` action again.
10. Using WP CLI run a command `wp wc tracker snapshot`
11. Check the comment output for 
```
... "store":{"connected":"yes","actively_syncing":"yes"},"feed":{"generation_time":111,"products_count":20} ...
```
12. `feed` object must be present and contain the values from the new feed generation run: generation time at least 3, product count to be one product more that it was before.

### Changelog entry

> Add - Plugin connection and feed registration status tracking.
> Add - Recent feed generation time and feed product count tracking.

### DIFF
--- a/src/FeedGenerator.php
+++ b/src/FeedGenerator.php
@@ -129,7 +129,7 @@ class FeedGenerator extends AbstractChainedJob {
 	protected function handle_start() {
 		self::log( __( 'Feed generation start. Preparing temporary files.', 'pinterest-for-woocommerce' ) );
 		/* Remember the time a feed generation has started. */
-		TrackerSnapshot::reset_feed_file_generation_time();
+		ProductFeedStatus::reset_feed_file_generation_time();
 		try {
 			$this->prepare_temporary_files();
 			ProductFeedStatus::set(
@@ -154,11 +154,11 @@ class FeedGenerator extends AbstractChainedJob {
 	 */
 	protected function handle_end() {
 		self::log( __( 'Feed generation end. Moving files to the final destination.', 'pinterest-for-woocommerce' ) );
-		/* Calculate the time it took a feed to generate. */
-		TrackerSnapshot::set_feed_file_generation_time( time() );
 		try {
 			$this->add_footer_to_temporary_feed_files();
 			$this->rename_temporary_feed_files_to_final();
+			/* Calculate the time it took a feed to generate. */
+			ProductFeedStatus::set_feed_file_generation_time( time() );
 		} catch ( Throwable $th ) {
 			$this->handle_error( $th );
 			throw $th;
@@ -415,6 +415,8 @@ class FeedGenerator extends AbstractChainedJob {
 				'error_message' => $th->getMessage(),
 			)
 		);
+		/* Marks feed as failed to generate. */
+		ProductFeedStatus::mark_feed_file_generation_as_failed();
 
 		self::log( $th->getMessage(), 'error' );
 		$this->schedule_next_generator_start( time() + self::WAIT_ON_ERROR_BEFORE_RETRY );
--- a/src/FeedGenerator.php
+++ b/src/FeedGenerator.php
@@ -1,4 +1,4 @@
-<?php //phpcs:disable WordPress.WP.AlternativeFunctions --- Uses FS read/write in order to reliable append to an existing file.
+<?php //phpcs:disable WordPress.WP.AlternativeFunctions --- Uses FS read/write in order to reliably append to an existing file.
 /**
  * Pinterest for WooCommerce Feed Files Generator
  *
@@ -36,12 +36,18 @@ class FeedGenerator extends AbstractChainedJob {
 	const WAIT_ON_ERROR_BEFORE_RETRY = HOUR_IN_SECONDS;
 
 	/**
+	 * Feed file operations class.
+	 *
+	 * @var FeedFileOperations
+	 */
+	private $feed_file_operations;
+
+	/**
 	 * Local Feed Configurations class.
 	 *
 	 * @var LocalFeedConfigs of local feed configurations;
 	 */
 	private $configurations;
-
 
 	/**
 	 * Location buffers. On buffer for each local feed configuration.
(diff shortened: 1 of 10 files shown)


src/TrackerSnapshot.php
```diff
@@ -57,8 +56,18 @@ protected static function include_snapshot_data( array $data = array() ): array
...
+
 		$data['extensions'][ PINTEREST_FOR_WOOCOMMERCE_TRACKER_PREFIX ] = array(
 			'settings' => self::parse_settings(),
+			'store'    => array(
+				'connected'        => wc_bool_to_string( Pinterest_For_Woocommerce::is_connected() ),
+				'actively_syncing' => wc_bool_to_string( ProductSync::is_product_sync_enabled() ),
+			),
+			'feed'     => array(
+				'generation_time' => $feed_generation_time,
+				'products_count'  => (int) ProductFeedStatus::get()['product_count'] ?? 0,
```
budzanowski: I think that the proper form is `product_count` and not `products...`.

budzanowski:
This looks good but I think that some modifications will be required.

The code will work correctly as long as the `WC_Tracker` snapshot will not happen during the feed generation process.
Since feed generation can take a lot of time, this can be a common occurrence.

This will lead to two things. The `PROP_FEED_GENERATION_WALL_TIME` will be equal to `0` and the `product_count` will be the number of products that have been processed since the start of the feed generation time.

src/FeedGenerator.php
```diff
@@ -128,6 +128,8 @@ private function start_generation() {
 	 */
 	protected function handle_start() {
 		self::log( __( 'Feed generation start. Preparing temporary files.', 'pinterest-for-woocommerce' ) );
+		/* Remember the time a feed generation has started. */
```
budzanowski: Please use single line comments for singular sentences.

In general I would omit comments like this. Thats because you have used a very descriptive name.

src/ProductFeedStatus.php
```diff
@@ -92,7 +98,47 @@ public static function set( $state ) {
 	public static function deregister() {
 
 		foreach ( self::STATE_PROPS as $key => $default_value ) {
+			self::$state[ $key ] = $default_value;
 			delete_transient( self::PINTEREST_FOR_WOOCOMMERCE_FEEDS_DATA_PREFIX . $key );
 		}
 	}
+
+	/**
+	 * Resets a feed generation start time.
```
budzanowski, line 109: Please add `@since` to the methods docblock.
You can use `@since x.x.x` and the release process will automatically populate the version.

budzanowski:
Fantastic job with the Unit tests!

message-dimke:
Hey, @budzanowski! I did update PR description with two test cases and detailed instructions on how to run those tests. As well as edited more files and added more tests.

src/TrackerSnapshot.php
```diff
@@ -57,8 +56,18 @@ protected static function include_snapshot_data( array $data = array() ): array
 			$data['extensions'] = array();
 		}
 
+		$feed_generation_time = ProductFeedStatus::get()[ ProductFeedStatus::PROP_FEED_GENERATION_WALL_TIME ];
+
 		$data['extensions'][ PINTEREST_FOR_WOOCOMMERCE_TRACKER_PREFIX ] = array(
 			'settings' => self::parse_settings(),
+			'store'    => array(
+				'connected'        => wc_bool_to_string( Pinterest_For_Woocommerce::is_connected() ),
+				'actively_syncing' => wc_bool_to_string( ProductSync::is_product_sync_enabled() ),
```
budzanowski: This will be included in the `settings` section. It has `product_sync_enabled` prop tracked.
It would be better to track if the feed is registered. You can use `Pinterest_For_Woocommerce()::get_data( 'feed_registered' )` but don't store the ID just check if it is not null.

src/ProductFeedStatus.php
```diff
@@ -90,9 +99,61 @@ public static function set( $state ) {
 	 * @return void
 	 */
 	public static function deregister() {
-
+		$properties_to_persist = array(
```
budzanowski: In case the feed is deregistered we should not send the count and time info - we should use the default values. This means that this logic can be eliminated.

budzanowski:
@message-dimke I have added some additional suggestions.

message-dimke:
@budzanowski , if deregistered is it ok to show `time: 0` and` products: 0` and since `actively_syncing` will container value `'no'` we will know this is not a sync, but a default which means nothing?

If the above is good, then I did the changes and committed. Ready for another review round.

budzanowski:
I think that it is OK.

budzanowski (approved):

PR #668, opened 2023-01-23
Fix price mismatch when taxable address is different than shipping zone

### LINKED ISSUE #638: Prices are showing without tax on the feed even if on the shop the prices includes tax
### Describe the bug:
There are some merchants that show prices including taxes on their sites. but on the feed, the price is without taxes.
For examples:
- Feed:  https://getstamped.co.uk/wp-content/uploads/pinterest-for-woocommerce-WlAhii.xml
- product: https://getstamped.co.uk/product/loyalty-card-stamp/

on the site the price is 10.00 on the feed 8.33

alexiglesias31:

Closes #638.

There is a price mismatch between the feed and the price displayed in the shop if the following conditions are met:
- Price entered inclusive of taxes.
- Display prices in the shop including taxes.
- Store address in XX country (e.g. Spain).
- Tax set only for the country where the store is set (e.g. Spain).
- The last shipping zone in the list do not include the store's address: e.g.:
   - Europe (Flat Rate...)
   - US (Flat Rate...)

For this particular kind of scenario the price in the feed does not take into account the tax, since it gets the taxable address in a country different than Spain.

2. Setup WC as described above (store address, tax region, shipping zone, etc).
3. The price in the feed should match with the price displayed in the shop.

### Changelog entry

> Fix - Price inconsistencies when tax region is different than shipping zone.

### DIFF
--- a/src/FeedGenerator.php
+++ b/src/FeedGenerator.php
@@ -495,7 +495,7 @@ class FeedGenerator extends AbstractChainedJob {
 	public function set_store_address_as_taxable_location( array $taxable_location ) {
 
 		if ( ! doing_action( $this->get_action_full_name( self::CHAIN_BATCH ) ) ) {
-			return;
+			return $taxable_location;
 		}
 
 		if ( isset( $taxable_location[0] ) ) {
--- a/src/FeedGenerator.php
+++ b/src/FeedGenerator.php
@@ -90,6 +90,9 @@ class FeedGenerator extends AbstractChainedJob {
 		if ( false === as_has_scheduled_action( self::ACTION_START_FEED_GENERATOR, array(), PINTEREST_FOR_WOOCOMMERCE_PREFIX ) ) {
 			$this->schedule_next_generator_start( time() );
 		}
+
+		// Set the store address as taxable location.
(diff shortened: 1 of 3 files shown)


sc-ecolasurdo (approved):

message-dimke:
Hey, @alexiglesias31 ! What if we set customer location on feed generation start and remove it on feed generation end instead of doing that for each product individually? What do you think?

alexiglesias31:
Hey @message-dimke, The feed generation can take quite long depending on the number of products, placing the filter at the start of the generation and removing at the end may interfere with other requests or methods of the site.
We could set the filter on every batch tho, e.g in the function `process_items`. Let me know your thoughts.

message-dimke:
> may interfere with other requests or methods of the site

Then we must check it, because if it is so then setting customer location when writing into the file may result in some odd website behaviour which will be hard to catch. Please, check if you find anything, I'll do that too and we need a test on that.

alexiglesias31:
Perfect. Will check and update :+1:

alexiglesias31:
Hey @message-dimke, I added the modifications and the test.

message-dimke (approved):
Thank you, @alexiglesias31 !

PR #672, opened 2023-01-24
Fix fatal error if get account data fails during onboarding

### LINKED ISSUE #671: Possible fatal error after saving the token
### Describe the bug:
There is a call to the endpoint `users/me` without an error handling hooked to the action `pinterest_for_woocommerce_token_saved`. If the call to the endpoint throwns an exception there is no catch block to avoid the fatal error and the onboarding of the plugin breaks.

### Steps to reproduce:
I don't have a clear way of reproducing the error, perhaps tweaking the code to ensure that the call to `users/me` returns throws an exception.

### Expected behavior:
The errors should be handled by the catch block to avoid the fatal error. 

### Actual behavior:
An API exception throws a fatal error and blocks the onboarding.

alexiglesias31:

Closes #671.

Handle properly errors that may occur getting the account data after saving the authorization token during the onboarding.

2. Add the following snippet:
`add_action( 'pinterest_for_woocommerce_token_saved', function() {
	throw new Exception( 'Test error during onboarding' );
}, 0 );`
3. An error notice should popup after connecting the account. Since we are simulating the error with an action the account is not disconnected. In a real scenario the account is disconnected too.

### Changelog entry

> Fix - Prevent fatal error if retrieving account fails during onboarding

### DIFF
--- a/assets/source/setup-guide/app/views/WizardApp.js
+++ b/assets/source/setup-guide/app/views/WizardApp.js
@@ -45,15 +45,16 @@ const WizardApp = ( { query } ) => {
 
 	const appSettings = useSettingsSelect();
 	const isDomainVerified = useSettingsSelect( 'isDomainVerified' );
+	const createNotice = useCreateNotice();
 
 	useEffect( () => {
 		if ( ! isConnected ) {
 			setIsBusinessConnected( false );
 		}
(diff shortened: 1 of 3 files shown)


sc-ecolasurdo (approved):

message-dimke (approved):
Works as described, thank you @alexiglesias31 !

PR #687, opened 2023-02-02
Add WordPress locale mapping to Pinterest API locale.

### LINKED ISSUE #683: Add WordPress locale mapping to Pinterest API locale.
The `https://api.pinterest.com/v3/catalogs/partner/connect/` endpoint supports only the following locale 

`it,es-419,ru-RU,hu-HU,hr-HR,sv-SE,de,nb-NO,th-TH,sk-SK,ro-RO,es-MX,uk-UA,nl,en-AU,he-IL,tr,es-ES,pl-PL,cs-CZ,en-CA,fi-FI,pt-PT,el-GR,ja,ar-SA,fr-CA,en-GB,es-AR,da-DK,zh-CN,en-US,vi-VN,id-ID,bg-BG,zh-TW,en-IN,tl-PH,ko-KR,af-ZA,ms-MY,bn-IN,te-IN,fr,hi-IN,pt-BR`

the task is to implement mapping of WordPress native locales that are required in different format by Pinterest. For example:
`nl-NL` should be mapped to `nl`

Throw an error when no mapping is possible.

budzanowski:

Closes #683  .

Pinterest API requires a different set of locales compared to what is used in WordPress. Most of the time the locales match. Sometimes 

Enable debug logging. 

**- supported locale**

1. Select one of the supported locales ( check #683 for the list ), WordPress language settings.
2. Perform the full onboarding flow.
3. Things should work.
4. ( logs ) call to https://api.pinterest.com/v3/catalogs/partner/connect/ should be OK

**- partially supported locale**

1. Select one of the partially supported locales ( check #683 for the list ), WordPress language settings.
2. Perform the full onboarding flow.
3. Things should work.
4. ( logs ) call to https://api.pinterest.com/v3/catalogs/partner/connect/ should be OK

**- unsupported locale** 

1. Select a locale that is not supported, ( not on the list in #683 ), WordPress language settings.
2. Perform the full onboarding flow.
3. Things should not work. 
4. ( logs ) **No** call to https://api.pinterest.com/v3/catalogs/partner/connect/ in logs
5. The following message should be visible in the UI ( Marketing -> Pinterest )
image

### Additional details:

#688 implements unit tests for this feature.

The error message copy ( last testing scenario ) is not final. We want to add a link to supported languages somewhere inside the Pinterest documentation. That page is not available yet. It will be a separate PR when the page will be created.

### Changelog entry

Add - WordPress locale to Pinterest locale mapping.

### DIFF
--- a/src/Exception/PinterestApiLocaleException.php
+++ b/src/Exception/PinterestApiLocaleException.php
@@ -14,4 +14,4 @@ use Exception;
 /**
  * Exception thrown when the the application locale is not supported by the API.
  */
-class PinterestApiLocaleException extends Exception {}
+class PinterestApiLocaleException extends Exception implements PinterestException {}
--- a/src/Exception/PinterestApiLocaleException.php
+++ b/src/Exception/PinterestApiLocaleException.php
@@ -1,0 +1,17 @@
+<?php
+/**
+ * PinterestApiLocaleException interface.
+ *
+ * @package Automattic\WooCommerce\Pinterest\Exception
+ */
+
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\Pinterest\Exception;
+
+use Exception;
+
+/**
+ * Exception thrown when the the application locale is not supported by the API.
+ */
+class PinterestApiLocaleException extends Exception {}
--- a/src/Feeds.php
+++ b/src/Feeds.php
@@ -12,8 +12,9 @@ if ( ! defined( 'ABSPATH' ) ) {
 	exit;
 }
 
-use Automattic\WooCommerce\Pinterest\API\Base;
 use \Exception;
+use Automattic\WooCommerce\Pinterest\API\Base;
+use Automattic\WooCommerce\Pinterest\Exception\PinterestApiLocaleException;
 
 /**
  * Class handling fetch methods for feed profiles.
@@ -127,7 +128,13 @@ class Feeds {
 		$configured_path = dirname( $feed->location_config->full_feed_fetch_location );
 		$local_path      = dirname( $config['feed_url'] );
 		$local_country   = Pinterest_For_Woocommerce()::get_base_country() ?? 'US';
-		$local_locale    = LocaleMapper::get_locale_for_api();
+
+		try {
+			$local_locale = LocaleMapper::get_locale_for_api();
+		} catch ( PinterestApiLocaleException $e ) {
+			// Local feed locale is not supported by Pinterest.
+			return '';
+		}
 
 		$registered_feed = '';
 
--- a/src/Feeds.php
+++ b/src/Feeds.php
@@ -128,14 +128,7 @@ class Feeds {
 		$configured_path = dirname( $feed->location_config->full_feed_fetch_location );
 		$local_path      = dirname( $config['feed_url'] );
 		$local_country   = Pinterest_For_Woocommerce()::get_base_country() ?? 'US';
-
-		try {
-			$local_locale = LocaleMapper::get_locale_for_api();
-		} catch ( PinterestApiLocaleException $e ) {
-			// Local feed locale is not supported by Pinterest.
-			return '';
-		}
-
+		$local_locale    = LocaleMapper::get_locale_for_api();
 		$registered_feed = '';
 
 		if ( $configured_path === $local_path && $local_country === $feed->country && $local_locale === $feed->locale ) {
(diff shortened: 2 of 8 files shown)


src/Exception/PinterestApiLocaleException.php
```diff
@@ -0,0 +1,17 @@
...
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\Pinterest\Exception;
+
+use Exception;
+
+/**
+ * Exception thrown when the the application locale is not supported by the API.
+ */
+class PinterestApiLocaleException extends Exception {}
```
JPry: This should `implement` the `PinterestException` interface.

src/Feeds.php
```diff
@@ -127,8 +128,7 @@ public static function is_local_feed_registered( $merchant_id ) {
 		$configured_path = dirname( $feed->location_config->full_feed_fetch_location );
 		$local_path      = dirname( $config['feed_url'] );
 		$local_country   = Pinterest_For_Woocommerce()::get_base_country() ?? 'US';
-		$local_locale    = str_replace( '_', '-', determine_locale() );
-
+		$local_locale    = LocaleMapper::get_locale_for_api();
```
JPry: Should we have a `try`/`catch` block around this to handle the possible `PinterestApiLocaleException` exception? It looks like only one place where this method is called is surrounded by a try/catch block.

src/LocaleMapper.php
```diff
@@ -0,0 +1,117 @@
...
+	 *
+	 * @since x.x.x
+	 * @return string
+	 * @throws PinterestApiLocaleException If no matching locale code is found.
+	 */
+	public static function get_locale_for_api() {
+		$locale = self::get_wordpress_locale();
+
+		// If the locale is in the list of Pinterest locales, return it.
+		if ( in_array( $locale, self::PINTEREST_LOCALE_CODES, true ) ) {
```
JPry: For this large of an array, it would be more performant to make the locales into keys, with simple values like `1` or `true`. Then we can use `array_key_exists()` instead of `in_array()` for the lookup.

This is especially relevant because we could possibly be examining the items in this array twice. Assuming that the item doesn't exist, that would lead to checking items in the array 94 times (47 * 2) instead of two times.

src/LocaleMapper.php
```diff
@@ -0,0 +1,117 @@
...
+			return $locale;
+		}
+
+		// If the locale is not in the list of Pinterest locales, try to find a match for just the language code.
+		$locale_parts = explode( '-', $locale );
+
+		if ( in_array( $locale_parts[0], self::PINTEREST_LOCALE_CODES, true ) ) {
+			return $locale_parts[0];
+		}
```
JPry: For better readability, we could assign this into meaningful variables instead of leaving it as a numerical index. This is easily done with array destructuring (example assumes the above-mentioned change to key-based codes is done):

```suggestion
		[ $language ] = explode( '-', $locale );

		if ( array_key_exists( $language, self::PINTEREST_LOCALE_CODES ) ) {
			return $language;
		}
```

src/LocaleMapper.php
```diff
@@ -0,0 +1,117 @@
...
+
+		// If no match was found, throw an exception.
+		throw new PinterestApiLocaleException( 'No matching Pinterest API locale found for ' . $locale );
```
JPry: Do we want this message to be translated?

JPry (changes requested):

src/LocaleMapper.php
budzanowski: I will replace if for the sake of moving forward. I though about it initially and decided that the cleanless of the array without artificial values is more favorable then potential speed improvement. For aprox 100 comparisions ( the worst case scenario ) the whole operation will take single microseconds and the API calls that are using this are called super rarely.

src/LocaleMapper.php
budzanowski: I needed to allow short array syntax for this - but that is OK imo.

src/LocaleMapper.php
budzanowski: Yes. Fixed.

src/Feeds.php
budzanowski: It is a bit convoluted but I think that at this stage all possible occurrences of this exception should be handled by one or another `try/catch` block.

We should probably create a separate issue and streamline exception handling because it is right now all over the place - but that should be a separate PR IMO.

budzanowski:
@JPry one more?

JPry (approved):
LGTM 👍🏻

PR #698, opened 2023-02-24
Provide external link EM tooltip

### LINKED ISSUE #267: Consider still providing users with the external link about the enhanced match
It's a follow-up mentioned in #253 by @ianlin:

> One suggestion regarding the current flow: as a customer I would like to know more information about each setting on the Pinterest site as I might not be able to understand it from a simple description on the tooltip. A better approach could be adding a `See more` link on the tooltip and it will open a new tab to the Pinterest site for more information about that setting when I click on it.

### Additional context 

We removed the external link https://help.pinterest.com/en/business/article/enhanced-match from #253. And Ian provides a suggestion that we could still provide the link to customers.

// cc @shivanarrthine

alexiglesias31:

Closes #267.

Adding back the external link to enhanced match documentation within the tooltip.

### Screenshots:

image

1. Install a build of this PR and do the onboarding.
2. The "See more" link should be available on the enhanced match tooltip.

### Changelog entry

> Add - Enhanced match documentation link.

### DIFF
--- a/TRACKING.md
+++ b/TRACKING.md
@@ -82,6 +82,7 @@ Clicking on an external documentation link.
 	- with `{ link_id: 'ad-data-terms', context: 'wizard'|'settings' }`
 	- with `{ link_id: 'ad-terms-of-service', context: 'wizard'|'settings' }`
 	- with `{ link_id: 'install-tag', context: 'wizard'|'settings' }`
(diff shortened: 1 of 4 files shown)


sc-ecolasurdo (approved):

message-dimke (approved):
LGTM! Link is there and working. Thank you, @alexiglesias31 !

PR #703, opened 2023-03-07
Fix/fatal error rest cart

alexiglesias31:

Fix fatal error for rest calls: https://wordpress.org/support/topic/fatal-error-4386/

### Changelog entry

> Fix - Fatal error if cart is not set.

### DIFF
--- a/src/Tracking.php
+++ b/src/Tracking.php
@@ -735,7 +735,7 @@ JS;
 	 * @return bool
 	 */
 	protected static function price_includes_tax() {
-		if ( isset( WC()->cart ) ) {
+		if ( isset( WC()->cart ) && method_exists( WC()->cart, 'display_prices_including_tax' ) ) {
 			return WC()->cart->display_prices_including_tax();
 		}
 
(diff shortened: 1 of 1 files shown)


message-dimke (approved):
Looks good! Thanks, @alexiglesias31 !

message-dimke:
Hey, @alexiglesias31 ! 
When `isset` successfully tests against `null` I am curious if it is a better idea to test `WC()->cart` to be a `WC_Cart` instance since there is a call to its method immediately after the condition is successful. `Null` is maybe better that instance of a class, since it will throw a Fatal when calling a method which does not exist rather than introducing an unexpected behaviour which could be harder to track down. In case of checking agains the instance of a `WC_Cart` class we must also log the mismatch not to crash user's website. What do you think?

Other than that looks good and I approved it already.

alexiglesias31:
Hey @message-dimke, as an extra layer of stability yes, I think we could use `method_exists` instead, that way we ensure there won't be any mismatch and if for some reason, the cart is not null but the method doesn't exist, then the value will be read from the WC settings.
Wdyt?

message-dimke (approved):
LGTM!

PR #721, opened 2023-04-10
Fix escape javascript product name attribute.

### LINKED ISSUE #716: Lack of escaping in the track function
### Describe the bug:
The code that sets up the pintrk() call in the JavaScript doesn't escape the product title, which means it will fail if the product title has an apostrophe in it.

### Steps to reproduce:
1. Create a product with an apostrophe in the title.
2. Visit that product page, observing the browser console.

### Expected behavior:
No console error and for the plugin to correctly track this with Pinterest.

### Actual behavior:
Uncaught SyntaxError: unexpected identifier 'Kursar'

This is related to the relevant snippet:
```
<script id='wc-add-to-cart-js-after'>
jQuery( function( $ ) {
	$( document.body ).on( 'added_to_cart', function( e, fragments, cart_hash, thisbutton ) {
		var quantity = document.querySelector( 'input.qty[name="quantity"]' ).value;
		pintrk( 'track', 'AddToCart', {
			'product_id': '23346',
			'product_name': 'Flowering cherry tree 'Kursar'',
			'value': 65 * quantity,
			'order_quantity': quantity,
			'currency': 'GBP'
		} );
	} );
} );
</script>
```

As you can see the product has Kursar in single quotes (which is the product name in the WooCommerce installation) but it is not escaped.

### Additional details:
Observed with 1.2.16.

Also appears to have been observed by another user: https://wordpress.org/support/topic/lack-of-escaping-in-javascript-causes-page-errors/

message-dimke:

Closes #716 .

When product name contains a quote tracker at product page fails.

1. Create a product with name which contains a single quote.

Edit_product_“Simple_product’s_name”_‹_WordPress_Pinterest_—_WordPress

2. Visit product page and see a JS error in a browser's console.

DevTools_-_pinterest_dima_works__product_simple-product

3. Checkout `fix/716-escape-js-product-name`.
4. Refresh the product page and see no JS console error and the product name properly escaped.

DevTools_-_pinterest_dima_works__product_simple-product

### Changelog entry

> Fix - Escape product name for JS tracker.

### DIFF
--- a/src/Tracking.php
+++ b/src/Tracking.php
@@ -371,7 +371,7 @@ JS;
 		}
 
 		$product_id    = $product->get_id();
-		$product_name  = $product->get_name();
+		$product_name  = esc_js( $product->get_name() );
 		$product_price = self::get_product_display_price( $product );
 
 		$wc_currency = get_woocommerce_currency();
--- a/src/Tracking.php
+++ b/src/Tracking.php
@@ -372,9 +372,9 @@ JS;
 
 		$product_id    = $product->get_id();
 		$product_name  = esc_js( $product->get_name() );
-		$product_price = self::get_product_display_price( $product );
+		$product_price = floatval( self::get_product_display_price( $product ) );
 
-		$wc_currency = get_woocommerce_currency();
+		$wc_currency = esc_js( get_woocommerce_currency() );
 		$tracking    = <<< JS
 jQuery( function( $ ) {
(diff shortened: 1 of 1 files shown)


ibndawood (approved):
Thank you, @message-dimke, for the fix. The changes look good to me. However, I am wondering if we should add similar escaping for `$product_price` and `$wc_currency` since they also pass through filters and can have incorrect string literals.

ibndawood (approved):
Thank you, @message-dimke. I just added `esc_js` to the `get_add_cart_snippet_archive` as well.

PR #729, opened 2023-04-17
Adds `esc_url` to `add_query_arg` function calls

ibndawood:

This PR adds `esc_url` to all `add_query_arg` function calls.

1. Verify WooCommerce is installed and activated.
2. Verify Pinterest for WooCommerce plugin is installed and activated.
3. Navigate to `Marketing > Pinterest > Connection` and verify that you are able to disconnect (if already connected) and reconnect successfully.
4. Navigate to the site store by clicking on the `Visit Store` link.
5. Hover over the product thumbnails and click on the `Save` icon from `Pinterest`. Verify the product is saved to Pinterest.
6. Install `semgrep` using `brew install semgrep`
7. Clone this repository in a top-level folder: `git clone git@github.com:Automattic/wpscan-semgrep-rules.git`
8. Run `semgrep --no-git-ignore --text -c audit ../pinterest-for-woocommerce` for `semgrep` results. The results should not return any findings.

### Changelog entry

> Fix - unescaped `add_query_arg` function calls

### DIFF
--- a/class-pinterest-for-woocommerce.php
+++ b/class-pinterest-for-woocommerce.php
@@ -830,7 +830,7 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 
 			set_transient( PINTEREST_FOR_WOOCOMMERCE_AUTH, $control_key, MINUTE_IN_SECONDS * 5 );
 
-			return self::get_connection_proxy_url() . 'login/' . PINTEREST_FOR_WOOCOMMERCE_WOO_CONNECT_SERVICE . '?' . $state;
+			return esc_url( self::get_connection_proxy_url() . 'login/' . PINTEREST_FOR_WOOCOMMERCE_WOO_CONNECT_SERVICE . '?' . $state );
 		}
 
 
--- a/src/API/Auth.php
+++ b/src/API/Auth.php
@@ -144,9 +144,11 @@ class Auth extends VendorAPI {
 			$query_args['view'] = sanitize_key( $view );
 		}
 
-		return add_query_arg(
-			$query_args,
-			admin_url( 'admin.php' )
+		return esc_url(
+			add_query_arg(
+				$query_args,
+				admin_url( 'admin.php' )
+			)
 		);
 	}
 }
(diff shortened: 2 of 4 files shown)


nima-karimi (commented):
Thanks, @ibndawood. The PR tests well, but I'm wondering if escaping is necessary in cases where the URL is neither stored in the DB nor displayed to the user. Only the `SaveToPinterest` file seems to be outputting the URL. In other cases, the URL is just fed to a redirect method. Do you think escaping those helps?

ibndawood:
Thank you, @nima-karimi, for your time and review.

> Do you think escaping those helps?

I agree that escaping should happen as late as possible, that is, when we output the string. However, there was a note in the original post read "_If the scanner picked it up, then really make sure that it’s not a risk, and even if it’s not, consider refactoring the code so that it doesn’t trigger a scan_". What do you recommend that we do here? Remove the escaping and let the code be as is as a false positive, or escape them to satisfy the scan results?

nima-karimi:
Hmm. The wpscan rule that is causing this false positive is this one: `wpscan-semgrep-rules/audit/php/wp/security/xss/query-arg.yml#L41`

This marks anything that returns an `add_query_arg` as unsafe. We could make a PR to remove that rule because it's too general. On the other hand, it is marked as a "warning" severity, so maybe it's good to keep it just in case the return value is outputted somewhere.

In this case, my opinion is to either leave it as a warning or add a comment so that semgrep ignores that line as a false positive:

```php
// nosemgrep: audit.php.wp.security.xss.query-arg
return add_query_arg(
	$query_args,
	admin_url( 'admin.php' )
);
```

ibndawood:
Thank you, @nima-karimi, for the explanation. I see the merit in your opinion. However, If it is okay with you, I'd like to keep the `esc_url` refactor to avoid any confusion about it being a false positive. 

More context: pcShBQ-MX-p2#comment-1679

nima-karimi (approved):
Thanks for the added context. That makes sense since I can't see any downside to escaping here. LGTM 👍

ibndawood:
Thank you, @nima-karimi.

PR #741, opened 2023-04-25
Refactor feed generator action backoff.

message-dimke:

Removing `action_scheduler_failed_execution` handler as the one which made feed generation to loop. Adding few other handlers (`action_scheduler_failed_execution` and `action_scheduler_unexpected_shutdown`) to properly handle action exceptions and action unexpected shutdown events.

#### Testing `action_scheduler_failed_execution`.
1. Checkout the PR.
2. Edit `FeedGenerator.php::handle_batch_action()` and make it throw an exception (preferably with a message you could remember to look for it in the logs). Search for a string like `Feed Generator pinterest/jobs/generate_feed/chain_batch Action failed to execute due to an error thrown **[Exception message here]**. A complete feed generation retry has been scheduled.`
3. Run a corresponding action and check the logs. Logs must contain records regarding the action failure and some information about retrying the action one hour later.

WooCommerce_status_‹_Wordpress_—_WordPress

#### Testing `action_scheduler_unexpected_shutdown`.
The handler is attached to a `shutdown` hook which runs on php shutdown. Don't have instructions on how to test this.

Both handlers are unit tested.

### Changelog entry

> Update - Pinterest feed generator backoff adjustments.

### DIFF
--- a/src/Exception/FeedFileOperationsException.php
+++ b/src/Exception/FeedFileOperationsException.php
@@ -14,7 +14,7 @@ use Exception;
 /**
  * An exception thrown then something went wrong writing into a feed file.
  */
-class FeedFileOperationsException extends Exception {
+class FeedFileOperationsException extends Exception implements PinterestException {
 	public const CODE_COULD_NOT_RENAME_ERROR = 10;
 
 	public const CODE_COULD_NOT_OPEN_FILE_ERROR = 20;
--- a/src/Exception/FeedFileOperationsException.php
+++ b/src/Exception/FeedFileOperationsException.php
@@ -1,10 +1,19 @@
 <?php
+/**
+ * Feed file operation exception.
+ *
+ * @package Automattic\WooCommerce\Pinterest\Exception
+ */
+
 declare( strict_types=1 );
 
 namespace Automattic\WooCommerce\Pinterest\Exception;
 
 use Exception;
 
+/**
+ * An exception thrown then something went wrong writing into a feed file.
+ */
 class FeedFileOperationsException extends Exception {
 	public const CODE_COULD_NOT_RENAME_ERROR = 10;
 
--- a/src/Exception/FeedFileOperationsException.php
+++ b/src/Exception/FeedFileOperationsException.php
@@ -1,0 +1,14 @@
+<?php
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\Pinterest\Exception;
+
+use Exception;
+
+class FeedFileOperationsException extends Exception {
+	public const CODE_COULD_NOT_RENAME_ERROR = 10;
+
+	public const CODE_COULD_NOT_OPEN_FILE_ERROR = 20;
+
+	public const CODE_COULD_NOT_WRITE_FILE_ERROR = 30;
+}
--- a/src/FeedGenerator.php
+++ b/src/FeedGenerator.php
@@ -659,10 +659,8 @@ class FeedGenerator extends AbstractChainedJob {
 	 * @since 1.2.14
 	 */
 	protected function is_timeout_error( array $error ): bool {
-		return isset( $error['type'] ) &&
-				E_ERROR === $error['type'] &&
-				isset( $error['message'] ) &&
-				strpos( $error ['message'], 'Maximum execution time' ) !== false;
+		return isset( $error['type'] ) && E_ERROR === $error['type'] &&
+			isset( $error['message'] ) && strpos( $error ['message'], 'Maximum execution time' ) !== false;
 	}
 
 	/**
--- a/src/FeedGenerator.php
+++ b/src/FeedGenerator.php
@@ -726,7 +726,7 @@ class FeedGenerator extends AbstractChainedJob {
 	 * @throws Exception Related to max retries reached or missing arguments on the action.
 	 */
 	public function maybe_handle_error_on_timeout( int $action_id ) {
-		wc_deprecated_function( __CLASS__ . '::' . __FUNCTION__, 'x.x.x' );
+		wc_deprecated_function( __METHOD__, 'x.x.x' );
 	}
 
 	/**
@@ -747,13 +747,12 @@ class FeedGenerator extends AbstractChainedJob {
 				'hook'         => $hook,
 				'args'         => $args,
 				'status'       => ActionSchedulerInterface::STATUS_FAILED,
-				'per_page'     => $threshold,
 				'date'         => gmdate( 'U' ) - $time_period,
 				'date_compare' => '>',
 			],
 			'ids'
 		);
 
-		return count( $failed_actions ) === $threshold;
+		return count( $failed_actions ) >= $threshold;
 	}
 }
--- a/src/FeedGenerator.php
+++ b/src/FeedGenerator.php
@@ -18,7 +18,6 @@ use Automattic\WooCommerce\ActionSchedulerJobFramework\Proxies\ActionSchedulerIn
 use Automattic\WooCommerce\Pinterest\Exception\FeedFileOperationsException;
 use Automattic\WooCommerce\Pinterest\Utilities\ProductFeedLogger;
 use ActionScheduler;
-use Error;
 use Exception;
 use Pinterest_For_Woocommerce;
 use Throwable;
@@ -681,7 +680,6 @@ class FeedGenerator extends AbstractChainedJob {
 			)
 		);
 	}
-
 
 	/**
 	 * Set the store address as taxable location.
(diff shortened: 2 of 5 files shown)


src/Exception/FeedFileOperationsException.php
```diff
@@ -0,0 +1,23 @@
...
+declare( strict_types=1 );
+
+namespace Automattic\WooCommerce\Pinterest\Exception;
+
+use Exception;
+
+/**
+ * An exception thrown then something went wrong writing into a feed file.
+ */
+class FeedFileOperationsException extends Exception {
```
JPry: This should also implement the `PinterestException` interface.

```suggestion
class FeedFileOperationsException extends Exception implements PinterestException {
```

src/FeedGenerator.php
```diff
@@ -629,39 +711,35 @@ public function handle_action_timeout( $action_id, $error ) {
 	 * @since 1.2.14
 	 */
 	protected function is_timeout_error( array $error ): bool {
-		return isset( $error['type'] ) &&
-				E_ERROR === $error['type'] &&
-				isset( $error['message'] ) &&
-				strpos( $error ['message'], 'Maximum execution time' ) !== false;
+		return isset( $error['type'] ) && E_ERROR === $error['type'] &&
+			isset( $error['message'] ) && strpos( $error ['message'], 'Maximum execution time' ) !== false;
```
JPry: To improve readability, I think it would make sense to rework this method a bit. Here's my idea:

```suggestion
		$is_error              = E_ERROR === $error['type'] ?? 0;
		$is_max_execution_time = strpos( $error['message'] ?? '', 'Maximum execution time' ) !== false;

		return $is_error && $is_max_execution_time;
```

src/FeedGenerator.php
```diff
@@ -629,39 +711,35 @@ public function handle_action_timeout( $action_id, $error ) {
...
+				'date_compare' => '>',
+			],
+			'ids'
+		);
 
-			$this->handle_generation_retries( $action_args[0], $action_args[1] );
-		} catch ( Throwable $th ) {
-			$this->handle_error( $th );
-		}
+		return count( $failed_actions ) >= $threshold;
```
JPry, line 756: If the `per_page` is set to `$threshold`, won't the result of `search()` never be more than `$threshold`?

src/FeedGenerator.php
```diff
@@ -101,9 +104,115 @@ function () {
...
+	 *
+	 * @param int       $action_id - Action Scheduler action id.
+	 * @param Throwable $throwable - Exception object.
+	 * @param string    $context - Action Scheduler context (e.g. WP CLI, Async Request or an empty string).
+	 *
+	 * @since x.x.x
+	 *
+	 * @return void
+	 */
+	public function handle_failed_execution( int $action_id, Throwable $throwable, string $context ) {
```
JPry: The `$context` parameter doesn't seem to be used here.

JPry (changes requested):

src/FeedGenerator.php
```diff
@@ -101,9 +104,114 @@ function () {
...
+					// Translators: 1. Action Scheduler hook name.
+					__(
+						'Feed Generator `%s` Action reschedule threshold has been reached. Quit.',
+						'pinterest-for-woocommerce'
+					),
+					$hook
+				)
+			);
+			return;
+		}
```
message-dimke, line 160: Hey, @JPry ! What do you think if we remove this. IMHO it does not make sense to have it since AS has its own retry mechanism.

src/FeedGenerator.php
message-dimke: Removed the parameter from the list and updated corresponding actions.

src/FeedGenerator.php
message-dimke: Changed the condition to `===`

src/Exception/FeedFileOperationsException.php
message-dimke: Done.

src/FeedGenerator.php
```diff
@@ -629,39 +710,36 @@ public function handle_action_timeout( $action_id, $error ) {
...
+	 * @param string $hook The job action hook.
+	 * @param ?array $args The job arguments.
 	 *
-	 * @param int $action_id The ID of the action marked as failed.
+	 * @return bool True if the action's error rate is above the threshold, and false otherwise.
 	 *
-	 * @throws Exception Related to max retries reached or missing arguments on the action.
+	 * @since x.x.x
 	 */
-	public function maybe_handle_error_on_timeout( int $action_id ) {
```
JPry, line 647: One small note here: While I believe it's *extremely* unlikely that any third-party code is calling this method, it's still `public`. With that in mind, I think we should remove it's content, mark it as `@deprecated`, and add a call to `wc_deprecated_function()` here.

src/FeedGenerator.php
```diff
@@ -629,39 +710,36 @@ public function handle_action_timeout( $action_id, $error ) {
...
-
-		try {
-			$action_args = $action->get_args();
-
-			if ( ! isset( $action_args[0] ) || ! is_int( $action_args[0] ) || ! isset( $action_args[1] ) || ! is_array( $action_args[1] ) ) {
-				throw new Exception( __( 'There was not possible to re-schedule the action, no args available.', 'pinterest-for-woocommerce' ) );
-			}
+	protected function is_failure_rate_above_threshold( string $hook, ?array $args = null ): bool {
+		$threshold   = apply_filters( 'pinterest_for_woocommerce_action_failure_threshold', 3 );
+		$time_period = apply_filters( 'pinterest_for_woocommerce_action_failure_time_period', 1800 ); // 30 minutes.
```
JPry: Minor: Let's change this to `30 * MINUTE_IN_SECONDS`, to make it more obvious what the number is without the need for the trailing comment.

```suggestion
		$time_period = apply_filters( 'pinterest_for_woocommerce_action_failure_time_period', 30 * MINUTE_IN_SECONDS );
```

src/FeedGenerator.php
```diff
@@ -629,39 +710,36 @@ public function handle_action_timeout( $action_id, $error ) {
...
+		$failed_actions = $this->action_scheduler->search(
+			[
+				'hook'         => $hook,
+				'args'         => $args,
+				'status'       => ActionSchedulerInterface::STATUS_FAILED,
+				'per_page'     => $threshold,
```
JPry: I've been giving this a bit more thought, and I think the combination of `'per_page' => $threshold` and the `count()` comparison below could use an adjustment. I don't think it won't work, but I do think it can be a bit challenging to be certain it's working as we expect.

I think we can safely eliminate the `per_page` parameter here, and then go back to the `>=` comparison in the `return` statement. This query is still bound by a date, so we're not at risk of querying too many actions and getting a higher number than we expect.

src/FeedGenerator.php
```diff
@@ -629,39 +710,36 @@ public function handle_action_timeout( $action_id, $error ) {
...
+				'date_compare' => '>',
+			],
+			'ids'
+		);
 
-			$this->handle_generation_retries( $action_args[0], $action_args[1] );
-		} catch ( Throwable $th ) {
-			$this->handle_error( $th );
-		}
+		return count( $failed_actions ) === $threshold;
```
JPry: With the comment above, I think we can change this back to `count( $failed_actions ) >= $threshold`.

JPry (changes requested):

src/FeedGenerator.php
message-dimke: Do we want to control the failure rate, let's say we don't want action to retry after two or four attempts (three by. default) and `per_page` will help us to do that when provided with a value from the `pinterest_for_woocommerce_action_failure_threshold ` filter. Unless we just do it up to three times and give up. 

I am still not sure if it is reasonable to have that level of control on the number of retries though.

I think we have two options here: 
- leave per_page and have a filter to redefine the number of attempts.
- remove per_page, remove the filter and have three attempts hardcoded.

What do you think, Jeremy?

src/FeedGenerator.php
```diff
@@ -715,6 +715,20 @@ protected function is_timeout_error( array $error ): bool {
...
+	 *
+	 * @since 1.2.14
+	 * @deprecated x.x.x
+	 *
+	 * @param int $action_id The ID of the action marked as failed.
+	 *
+	 * @throws Exception Related to max retries reached or missing arguments on the action.
+	 */
+	public function maybe_handle_error_on_timeout( int $action_id ) {
+		wc_deprecated_function( __CLASS__ . '::' . __FUNCTION__, 'x.x.x' );
```
JPry:
```suggestion
		wc_deprecated_function( __METHOD__, 'x.x.x' );
```

message-dimke:
Hey, @JPry ! 

Could you advice, please, which changes are still needed, since I do not see any unresolved requests.

JPry (approved):
LGTM 👍🏻

PR #753, opened 2023-05-05
Implement Pinterest API V5 support

budzanowski:
### Description

This PR introduces support for the Pinterest API V5 in the Pinterest for WooCommerce plugin by adding a new class `APIV5`. The PR also includes a refactoring of the `Base` class by moving the `prepare_request` method from `Base` to both `APIV5` and `Base` classes.

### Changes

1. Created a new file `APIV5.php` which defines the `APIV5` class that extends the `Base` class.
   - Set the API domain for V5 in the `APIV5` class
   - Implemented the `prepare_request` method in the `APIV5` class to handle requests to the V5 API

2. Refactored the `Base` class in `Base.php`.
   - Moved the `prepare_request` method out of the `send` method and made it a separate public static method.
   - Updated the `send` method to use the `prepare_request` method to prepare the request before handling it.

### Impact

This update will allow the plugin to communicate with the latest Pinterest API (V5) while still maintaining backward compatibility with previous versions of the API. This will enable users to take advantage of new features and improvements provided by Pinterest's API V5.

### Testing

1. Verify that existing functionality using the `Base` class is not affected by the refactoring.
2. Test the new `APIV5` class by making requests to the V5 API and ensuring they are processed correctly.

### DIFF
--- a/src/API/APIV5.php
+++ b/src/API/APIV5.php
@@ -1,0 +1,47 @@
+<?php
+/**
+ * Pinterest API V5 class
+ *
+ * @class       Pinterest_For_Woocommerce_API
+ * @version     x.x.x
+ * @package     Pinterest_For_WordPress/Classes/
+ */
+
+namespace Automattic\WooCommerce\Pinterest\API;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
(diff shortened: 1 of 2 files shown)


message-dimke (approved):
LGTM!

PR #768, opened 2023-06-06
Ensure feed IDs is an array, even if it has not been saved before

### LINKED ISSUE #767: website crashes when I try to install the pinterest plugin
### Describe the bug:
website crashes when I try to install the pinterest plugin

### Steps to reproduce:

1. Try to install pinterest pluggin for woocommerce
2. 
3. 

### Expected behavior:
Install completes and pinterest is connected to woocommerce...
### Actual behavior:
install crashes out and then your website goes down and you have to enter recovery mode and deactivate pinterest before your website comes back up

JPry:

Closes #767.

This ensures that the `$feed_ids` variable is always an array, so that there is no error when it is used within `array_key_exists()`.

1. With PHP 8+, install and activate Pinterest for WooCommerce
2. After activation, there should be no fatal error

### Changelog entry

> Fix - ensure there is no fatal error for PHP 8+ during activation

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -21,6 +21,7 @@
 		<exclude name="Generic.Files.LineEndings.InvalidEOLChar"/>
 		<exclude name="Generic.Commenting.DocComment.MissingShort"/>
 		<exclude name="Generic.Arrays.DisallowShortArraySyntax"/>
+		<exclude name="WordPress.PHP.DisallowShortTernary"/>
 	</rule>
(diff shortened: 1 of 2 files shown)


layoutd (approved):
👍🏻  This fixes [the fatal with PHP 8](https://github.com/woocommerce/pinterest-for-woocommerce/issues/767#issuecomment-1579322614) identified in the issue

PR #777, opened 2023-06-13
Abort Tracking if this is a REST API request

### LINKED ISSUE #776: REST API requests generate error when object cache in use
### Describe the bug:
On a site running on PHP 8.1 with Woocommerce and pinterest-for-woocommerce installed, **plus Object Cache Pro installed**, with tracking enabled, a request to the REST API generates this error:

```
PHP message: objectcache.error: Cache key must be integer or non-empty string, null given (require('wp-blog-header.php'), require_once('wp-load.php'), require_once('wp-config.php'), require_once('wp-settings.php'), do_action('init'), WP_Hook->do_action, WP_Hook->apply_filters, Automattic\\WooCommerce\\Pinterest\\Tracking::maybe_init, Automattic\\WooCommerce\\Pinterest\\Tracking::load_async_events, get_transient)
```

For example, the application generates a REST API call to `/wp-json/wc-admin/plugins/active?_locale=user` when the Woocommerce Analytics Overview page was visited (`/wp-admin/admin.php?page=wc-admin&path=/analytics/overview`), resulting in this error.

### Steps to reproduce:

1. Install Woocommerce, pinterest-for-woocommerce, and Object Cache Pro, on a Wordpress app running on PHP 8.1.
2. Visit /wp-admin/admin.php?page=wc-admin&path=/analytics/overview (or some other page that generates REST API requests).

### Expected behavior:
No error should be generated.

### Actual behavior:
It generates the error listed above, visible in the error log.

### Additional details:
This occurs because 

1. [`Tracking::maybe_init()` does not abort for REST requests](https://github.com/woocommerce/pinterest-for-woocommerce/blob/e02701cf9dc7acda359e91ddce0845a1a6b348e0/src/Tracking.php#L81). (Though it does abort for cron and admin pages.)
2. `Tracking::maybe_init()` only[ initializes `self::$deferred_conversion_events_transient_key` if `WC()->session()` exists](https://github.com/woocommerce/pinterest-for-woocommerce/blob/e02701cf9dc7acda359e91ddce0845a1a6b348e0/src/Tracking.php#L85-L87), which apparently it does not for a REST request.
3. When `Tracking::load_async_events()` is called during the REST request, it [passes that null version of `self::$deferred_conversion_events_transient_key` to get_transient()](https://github.com/woocommerce/pinterest-for-woocommerce/blob/e02701cf9dc7acda359e91ddce0845a1a6b348e0/src/Tracking.php#L142)

My best guess is that `Tracking::maybe_init()` should abort whenever a call is made to the REST API. I don't imagine there is a use case for tracking REST API requests. If there are edge cases where an application would like to run the Pinterest tracking code during REST API requests, we could wrap this in a hook.

REST requests may have been left out because of the apparent difficulty in checking for them during `init`. WP-Core does not (yet?) provide a way to check for REST requests until later actions. ... Luckily, Woocommerce offers a public API that allows us to check for REST requests during `init`: [`WP()->is_rest_api_request()`](https://woocommerce.github.io/code-reference/classes/WooCommerce.html#method_is_rest_api_request)

I propose adding a check for `WP()->is_rest_api_request()` in `Tracking::maybe_init()`. I will follow-up with a pull request.

### Alternatives
I suppose that, instead, you could do one of these  options to fix this error:

1. Abort whenever `WC->session()` is not an object.
2. Set `self::$deferred_conversion_events_transient_key` to something not null when `WC->session()` is not an object.

patrickmcelwee:
Closes #776  .

See the issue for an explanation and possible alternatives.

### Changelog entry

> Fix - Fixed error when getting transient for a key of null value.

### DIFF
--- a/src/Tracking.php
+++ b/src/Tracking.php
@@ -78,7 +78,7 @@ class Tracking {
 	 */
 	public static function maybe_init() {
 
-		if ( ! self::tracking_enabled() || wp_doing_cron() || is_admin() || WC()->is_rest_api_request() ) {
+		if ( ! self::tracking_enabled() || wp_doing_cron() || is_admin() ) {
 			return;
 		}
 
--- a/src/Tracking.php
+++ b/src/Tracking.php
@@ -78,7 +78,7 @@ class Tracking {
 	 */
 	public static function maybe_init() {
 
-		if ( ! self::tracking_enabled() || wp_doing_cron() || is_admin() ) {
+		if ( ! self::tracking_enabled() || wp_doing_cron() || is_admin() || WC()->is_rest_api_request() ) {
 			return;
 		}
 
--- a/src/Tracking.php
+++ b/src/Tracking.php
@@ -139,14 +139,14 @@ class Tracking {
 	 */
 	private static function load_async_events() {
 
-    if ( self::$deferred_conversion_events_transient_key ) {
-      $async_events = get_transient( self::$deferred_conversion_events_transient_key );
+		if ( self::$deferred_conversion_events_transient_key ) {
+			$async_events = get_transient( self::$deferred_conversion_events_transient_key );
 
-      if ( $async_events ) {
-        self::$events = array_merge( self::$events, $async_events );
-        delete_transient( self::$deferred_conversion_events_transient_key );
-      }
-    }
+			if ( $async_events ) {
+				self::$events = array_merge( self::$events, $async_events );
+				delete_transient( self::$deferred_conversion_events_transient_key );
+			}
+		}
 	}
 
 
--- a/src/Tracking.php
+++ b/src/Tracking.php
@@ -139,12 +139,14 @@ class Tracking {
 	 */
 	private static function load_async_events() {
 
-		$async_events = get_transient( self::$deferred_conversion_events_transient_key );
+    if ( self::$deferred_conversion_events_transient_key ) {
+      $async_events = get_transient( self::$deferred_conversion_events_transient_key );
 
-		if ( $async_events ) {
-			self::$events = array_merge( self::$events, $async_events );
-			delete_transient( self::$deferred_conversion_events_transient_key );
-		}
+      if ( $async_events ) {
+        self::$events = array_merge( self::$events, $async_events );
+        delete_transient( self::$deferred_conversion_events_transient_key );
+      }
+    }
 	}
 
 


budzanowski:
@patrickmcelwee Thank you! We will test and if everything works we will want to have it in the next release.

message-dimke:
Hi, @patrickmcelwee ! 

While reviewing your PR I found that the code you introduced will block REST API calls which are part of WordPress Blocks (WordPress block editor).

When using blocks to render a shop page we use tracking e.g. on Add to cart button. 

Of course with REST API session will not work and we will have to search for customer id elsewhere, but again we can not block some frontend REST API calls because of Blocks.

What if you update the function and wrap everything into a condition if key is something?

```php
private static function load_async_events() {
	if ( self::$deferred_conversion_events_transient_key ) {
		$async_events = get_transient( self::$deferred_conversion_events_transient_key );

		if ( $async_events ) {
			self::$events = array_merge( self::$events, $async_events );
			delete_transient( self::$deferred_conversion_events_transient_key );
		}
	}
}
```

Async events will not get into the storage because there is no session means there is no storage key and if there is no storage key nothing will get fetched from the storage hopefully not producing any Object Cache Pro errors.

patrickmcelwee:
Thanks to you @message-dimke! 

Yes, I see that your proposed solution has a smaller blast radius and I confirmed that it solved my Object Cache Pro error.

I updated my PR with it. I didn't squash commits, but I suppose you could if you prefer while merging. Or let me know and I can do that.

message-dimke:
@patrickmcelwee , could you fix code styling issues, please?

You can find errors here https://github.com/woocommerce/pinterest-for-woocommerce/pull/777/files

patrickmcelwee:
> @patrickmcelwee , could you fix code styling issues, please?
> 
> You can find errors here https://github.com/woocommerce/pinterest-for-woocommerce/pull/777/files

I believe I've now fixed them, though I'm not set up to run the check locally.

message-dimke (approved):
@patrickmcelwee , all good! Thank you!

message-dimke:
Congrats @patrickmcelwee !

https://github.com/woocommerce/pinterest-for-woocommerce/commit/e6a5953e90d0cc9df56d9a96d206e7238b1c2ed3

PR #777 ;-)

budzanowski:
Nice work @patrickmcelwee.

The update has been released with the latest version ( 1.3.4 ).

If you will find the time please give the plugin a review at https://wordpress.org/plugins/pinterest-for-woocommerce/

Thank you for your contribution!

patrickmcelwee:
Sweet ... thanks for getting this fix in there!

PR #792, opened 2023-07-25
Add `users/me/businesses` exception handling.

message-dimke:

Adding exception handling on a API call. Since there is still a check for businesses to be empty or not we can fallback to empty array in case of API call exception instead of throwing the exception and disconnect the user because of a failure up above in the code which is making the request.

### Additional details:

When user is onboarding, after the successful token save procedure we run two Pinterest API calls to fetch user information and businesses. If any of those two calls fail with exception we disconnect the user and reset the onboarding. Instead of throwing an exception when making a call to `users/me/businesses` we could handle the exception properly and introduce a fallback to an empty array of businesses. That is what this PR is doing.

### Changelog entry

> Tweak - Add Pinterest businesses API exception handling.

### DIFF
--- a/class-pinterest-for-woocommerce.php
+++ b/class-pinterest-for-woocommerce.php
@@ -1109,7 +1109,11 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 				array_key_exists( 'is_partner', $account_data ) &&
 				! $account_data['is_partner'];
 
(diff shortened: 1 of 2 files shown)


ibndawood (approved):
Thank you, @message-dimke, for the changes. I looked into the `Base::make_request` and compared its params with the `get_linked_businesses` method. I confirm the return values and the exceptions.

PR #805, opened 2023-08-07
Update/optimize calls to advertisers countries

### LINKED ISSUE #800: Optimize calls to "advertisers/countries"
Part of https://github.com/woocommerce/pinterest-for-woocommerce/issues/790

High-level tasks: Optimize calls to "advertisers/countries"

Next step - analyze the usage and present an optimization strategy.

budzanowski:

Closes #800 .

We have been notified that there are too many calls to the `advertisers/countries` endpoint. This PR helps to reduce the number of calls via two things:

1. Increase the time for which the response is cached from 1 to 2 days. This endpoint is basically static so we could increase the number even higher if we would be forced to.
2. Detect if we are connected and if not abort the API call - go straight to the fallback list.

The plugin should not be connected.

1. Set a breakpoint at `Pinterest_For_Woocommerce_Ads_Supported_Countries::get_countries`
3. Open the Marketing -> Pinterest page.
4. Call to the `get_countries` should not trigger the API call but the fallback should be used.
5. Connect the plugin
6. After connection the call to the `get_countries` should trigger an API call ( first time )
7. Refresh the page - the next call should use the cached value.

### Changelog entry

> Tweak - Don't fetch the allowed countries list when not connected.
> Tweak - Increase allowed countries cache lifetime to two days.

### DIFF
--- a/includes/class-pinterest-for-woocommerce-ads-supported-countries.php
+++ b/includes/class-pinterest-for-woocommerce-ads-supported-countries.php
@@ -7,7 +7,6 @@
  */
 
 use Automattic\WooCommerce\Pinterest\API\Base;
-use \Exception;
 
 if ( ! defined( 'ABSPATH' ) ) {
 	exit;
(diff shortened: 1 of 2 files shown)


includes/class-pinterest-for-woocommerce-ads-supported-countries.php
```diff
@@ -6,7 +6,8 @@
  * @since 1.0.5
  */
 
-use  Automattic\WooCommerce\Pinterest\API\Base;
+use Automattic\WooCommerce\Pinterest\API\Base;
+use \Exception;
```
JPry: This file isn't namespaced, so this import doesn't have any effect. I also see a PHP warning generated.

```suggestion
```

JPry (approved):
Works as described, just needs the PHPCS fix and the one suggestion I made before merging.

PR #806, opened 2023-08-08
Fix add to cart event is not fired with add to cart Ajax button.

### LINKED ISSUE #789: Add to cart events on Ajax add to cart rely on WooCommerce Add to cart behaviour
### Describe the bug:

Issue opened for discussion to help confirm the expected behaviour:

Add to cart events aren't being tracked on themes or plugins using AJAX on the single product page when **Enable AJAX add to cart buttons on archives** isn't enabled in WooCommerce Add to cart behaviour in **WooCommerce > Settings > Products**.

Since AJAX add to cart from plugins/themes work without the option being enabled, Pinterest for WooCommerce should be able to track add to cart events when that option isn't enabled. 

### Steps to reproduce:

1. Go to WooCommerce > Settings > Products and uncheck both "Redirect to the cart page after successful addition" and "Enable AJAX add to cart buttons on archives"

2. Install and activate the [Ajax add to cart for WooCommerce](https://wordpress.org/plugins/woo-ajax-add-to-cart/) plugin (or another plugin offering similar functionality)
3. Add a product to the cart from the single product page via AJAX.
4. Check Pinterest tag helper and you'll find the add to cart event isn't fired.
5. Go to WooCommerce > Settings > Products and enable only "Enable AJAX add to cart buttons on archives"
6. Add a product to the cart from the single product page via AJAX and notice the add to cart event is fired as expected. 

### Expected behavior:

Add to cart event should be fired on Ajax add to cart even when "Enable AJAX add to cart buttons on archives" isn't enabled in WooCommerce. As far as I know, Ajax add to cart functionalities don't rely on the setting, so the Pinterest add to cart event shouldn't rely on it. 

### Actual behavior:

Add to cart events aren't fired on Ajax add to cart when "Enable AJAX add to cart buttons on archives" isn't enabled in WooCommerce.

### LINKED ISSUE #793: Add to cart events aren't tracked for Ajax add to cart
### Describe the bug:

Add to cart events aren't being tracked on themes or plugins using AJAX add to cart on the single product page.

### Steps to reproduce:

1. Install and activate the [Ajax add to cart for WooCommerce](https://wordpress.org/plugins/woo-ajax-add-to-cart/) plugin (or another plugin or theme offering similar functionality)
2. Add a product to the cart from the single product page via AJAX.
5. Check Pinterest tag helper and you'll find the add to cart event isn't fired.

(Note: for [Ajax add to cart for WooCommerce](https://wordpress.org/plugins/woo-ajax-add-to-cart/) plugin, you can get the add to cart event to fire by selecting "Enable AJAX add to cart buttons on archives" in WooCommerce > Settings > Product. See the [related bug report](https://github.com/woocommerce/pinterest-for-woocommerce/issues/789).)

### Expected behavior:

Add to cart event should be fired on Ajax add to cart on the single product page. 

### Actual behavior:

Add to cart events aren't fired on Ajax add to cart on the single product page.

message-dimke:

Closes #789 .
Closes #793 .

Add to cart events aren't being tracked on themes or plugins using AJAX on the single product page when Enable AJAX add to cart buttons on archives isn't enabled in WooCommerce Add to cart behaviour in WooCommerce > Settings > Products.

Removing _Enable AJAX add to cart buttons on archives_ value check not to affect add to cart events.

1. Checkout `fix/add-to-cart-event-with-ajax` branch.
2. Install Chrome Pinterest Tag Helper extension.
3. Make sure WooCommerce settings for `Redirect to the cart page after successful addition` and `Enable AJAX add to cart buttons on archives` are both set to `OFF`.

255561257-89ad0e91-46bb-4724-b4d0-3c9db844180c

4. Install [Ajax add to cart for WooCommerce](https://wordpress.org/plugins/woo-ajax-add-to-cart/) and Activate the plugin.
5. Open any product page and hit Add to cart button.
6. Check the request was made with AJAX.
7. Check Pinterest Tag Helper and search for Add to Cart event tracked.

Cursor_and_Beanie_–_WordPress

8. Deactivate _**Ajax add to cart for WooCommerce**_
9. Open any product page and hit Add to cart button.
10. Check the request was not AJAX and the page was reloaded.
11. Check Pinterest Tag Helper and search for Add to Cart event tracked.

Cursor_and_Beanie_–_WordPress-2

### Changelog entry

> Fix - Make add to cart events independent from WooCommerce archive page settings.

### DIFF
--- a/src/Tracking.php
+++ b/src/Tracking.php
@@ -100,7 +100,7 @@ class Tracking {
 			}
 
 			// AddToCart - ajax.
-			if ( 'yes' === get_option( 'woocommerce_enable_ajax_add_to_cart' ) && 'yes' !== get_option( 'woocommerce_cart_redirect_after_add' ) ) {
+			if ( 'yes' !== get_option( 'woocommerce_cart_redirect_after_add' ) ) {
 				add_action( 'wp_enqueue_scripts', array( __CLASS__, 'ajax_tracking_snippet' ), 20 );
 				add_filter(
 					'woocommerce_loop_add_to_cart_args',
--- a/tests/Unit/TrackingTest.php
+++ b/tests/Unit/TrackingTest.php
@@ -2,6 +2,8 @@
 
 namespace Automattic\WooCommerce\Pinterest;
 
+use Pinterest_For_Woocommerce;
+
 class TrackingTest extends \WP_UnitTestCase {
 
 	function setUp() {
@@ -10,10 +12,12 @@ class TrackingTest extends \WP_UnitTestCase {
 	}
 
 	function test_ajax_tracking_snippet_action_added() {
-		add_option( 'woocommerce_enable_ajax_add_to_cart', 'yes' );
-		add_option( 'woocommerce_cart_redirect_after_add', 'no' );
-		\Pinterest_For_Woocommerce::save_setting( 'track_conversions', true );
-		\Pinterest_For_Woocommerce::save_setting( 'tracking_tag', 'some-tag-id' );
+		// Deleting the option to make sure it does not affect tracking.
+		delete_option( 'woocommerce_enable_ajax_add_to_cart' );
+		update_option( 'woocommerce_cart_redirect_after_add', 'no' );
+
+		Pinterest_For_Woocommerce::save_setting( 'track_conversions', true );
+		Pinterest_For_Woocommerce::save_setting( 'tracking_tag', 'some-tag-id' );
 
 		Tracking::maybe_init();
 
@@ -25,5 +29,19 @@ class TrackingTest extends \WP_UnitTestCase {
 			10,
 			has_filter( 'woocommerce_loop_add_to_cart_args', array( Tracking::class, 'filter_add_to_cart_attributes' ) )
 		);
+	}
+
+	function test_ajax_tracking_snippet_action_is_not_added() {
+		// Deleting the option to make sure it does not affect tracking.
+		delete_option( 'woocommerce_enable_ajax_add_to_cart' );
+		update_option( 'woocommerce_cart_redirect_after_add', 'yes' );
+
+		Pinterest_For_Woocommerce::save_setting( 'track_conversions', true );
+		Pinterest_For_Woocommerce::save_setting( 'tracking_tag', 'some-tag-id' );
+
+		Tracking::maybe_init();
+
+		$this->assertFalse( has_action( 'wp_enqueue_scripts', array( Tracking::class, 'ajax_tracking_snippet' ) ) );
+		$this->assertFalse( has_filter( 'woocommerce_loop_add_to_cart_args', array( Tracking::class, 'filter_add_to_cart_attributes' ) ) );
 	}
 }


message-dimke:
Hey, @sukafia ! Here is the first fix.

src/Tracking.php
```diff
@@ -100,7 +100,7 @@ public static function maybe_init() {
 			}
 
 			// AddToCart - ajax.
-			if ( 'yes' === get_option( 'woocommerce_enable_ajax_add_to_cart' ) && 'yes' !== get_option( 'woocommerce_cart_redirect_after_add' ) ) {
+			if ( 'yes' !== get_option( 'woocommerce_cart_redirect_after_add' ) ) {
```
ibndawood, line 103: I'd like to understand why this condition was there in the first place. I checked the `ajax_tracking_snippet` and can see that it attaches a JS function to the `added_to_cart` event. 

Do we need these conditions if this event is triggered without these options? Can you please provide some context here.

tests/Unit/TrackingTest.php
ibndawood, line 1: Excellent work with the test cases.

ibndawood (approved):
Hey @message-dimke, thank you for the changes. They test well and look good to me. I've raised a question related to the context of the condition to understand the necessity of the condition, which is not a blocker.

src/Tracking.php
message-dimke: Hi, @ibndawood !

I'll tell you what I know. 
- `woocommerce_enable_ajax_add_to_cart` option is a setting of WooCommerce with a label of `Enable AJAX add to cart buttons on archives`. Archives, not product pages. So it is, at bare minimum, confuses our users. And I do not know why it was decided to use it elsewhere except archive pages.
- `woocommerce_cart_redirect_after_add` option speaks for itself and controls the redirect after Add to cart was clicked.

Next. We have a JS code shipped with WooCommerce which first checks for `woocommerce_cart_redirect_after_add` option enabled and redirect immediately or, if not, triggers JS `added_to_cart` event. 

All the above is WooCommerce codebase.

Now to Pinterest extension. We have a tracker which is a JS tracker. Also we have two strategies how to call events: **1.** we render JS function call immediately, e.g. adding a listener to the above `added_to_cart` WooCommerce JS event, or **2.** we store the JS code into transient and render it on the next page load (any page and if the user will abandon the site we will not have it ever rendered and reported to Pinterest).

The code you were looking at is a conditional code to render JS event to trigger `pintrk()` tracker function within JS if `woocommerce_cart_redirect_after_add` is `no`. There is some code below which hooks into `woocommerce_add_to_cart` event but inside its callback there is a condition to exit if `wp_doing_ajax()` is TRUE which makes JS and the hook mutually exclusive (even if it is not obvious at first sign). 

When the solution is not elegant, it is what it is and I decided not to refactor it because we are moving to v5 and I have a PR ready introducing another tracker - `Conversions API` with which I have all the tracking code refactored (including the above JS code removed) and if there is a case of website abandonment then at least one tracker will catch an `add to cart` event, because we are going to have both trackers: a JS one and an API. The last is always realtime (except bulk, which is not yet implemented).

ibndawood: Thank you so much, @message-dimke, for the time to explain the context of the condition. I really appreciate it. I've already approved the PR so you are good to merge it.

PR #811, opened 2023-08-09
Fix caching.

### LINKED ISSUE #807: Caching in make_request is not working.
### Describe the bug:
Calling get_list_of_ads_supported_countries always results in hitting the endpoint and the cache is not used.

### Steps to reproduce:

1. Connect the plugin
2. Set a breakpoint at get_list_of_ads_supported_countries
3. Go to Marketing -> Pinterest
4. Observe that the make_request value is not cached ( inside the make_request )

### Expected behavior:

Calls to make_request should be cached.

### Actual behavior:

The API call is done each time.

budzanowski:

Closes #807 .

The main change implemented in this PR is moving the lines that modify the `$api` parameter above the retrieval from the cache line. This fixes the cache issue. Since the `$api` parameter is used to calculate the cache key it is crucial that the cache retrieval and cache storing procedures observe the same value of the `$api` parameter. This was not the case when cache retrieval was done before the line that potentially changes the `$api` parameter.

1. Connect the plugin
2. Set a breakpoint at the `make_request` in the `get_list_of_ads_supported_countries`
3. Open Marketing -> Pinterest
4. Step into `make_request` observe cache behaviour
5. Subsequent calls to `make_requst` via ``get_list_of_ads_supported_countries` should properly use cache.

### Additional details:

1. Cache storing code was extracted into a separate method.
2. None critical code lines have been extracted out of `try` o make the `try` more obivious.

### Changelog entry

Fix - Caching of API calls.

### DIFF
--- a/src/API/Base.php
+++ b/src/API/Base.php
@@ -112,10 +112,7 @@ class Base {
 
 			$response = self::handle_request( $request );
 
-			if ( ! empty( $cache_expiry ) ) {
-				$cache_key = self::get_cache_key( $endpoint, $method, $payload, $api );
-				set_transient( $cache_key, $response, $cache_expiry );
-			}
+			self::maybe_cache_api_response( $endpoint, $method, $payload, $api, $response, $cache_expiry );
 
 			return $response;
 		} catch ( ApiException $e ) {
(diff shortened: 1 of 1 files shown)


ibndawood (approved):
Thank you, @budzanowski, for the changes. I tested the changes and noticed that difference in `ads` and `ads/` was causing the cache key to be different. With this fix, I can the cache being used.

budzanowski:
@ibndawood Thank you for the review.

PR #815, opened 2023-08-24
Updates phpunit version, unit tests and matrix for PHP8.2 compatibility.

krutidugade:

As part of PHP 8.1+ compatibility testing, we are updating the unit test tools to a newer version of PHP. This PR has following changes:
- Updates minimum PHP version to 7.4 since we'll [no longer support PHP 7.3 in WC 8.2 in October 2023](https://github.com/woocommerce/woocommerce/pull/39820). 
- Updates the php-unit-tests.yml to run PHP Unit test checks for PHP 8.0, 8.1, and 8.2 in addition to previous checks.
- Bumped the PHPUnit dev dependency to ^9.6.
- Removed "scaffolded tests" notice code from the bootstrap file for unit tests to run with 8.2.
- Updates unit tests to run with the latest PHP version
- Fixes PHPCS issues - mostly related to adding docblock comments.

1. Verify all the git workflow checks for PHP unit tests and PHP coding standard pass.
2. Expand PHP Unit tests and verify all tests pass without any errors.

### Additional details:

- Updated PHPUnit with `composer require phpunit/phpunit --dev --update-with-all-dependencies`
- Changed setup function syntax in unit tests to `setUp(): void`. This eliminates the following error
```
PHP Fatal error:  Declaration of Automattic\WooCommerce\Pinterest\Tests\Unit\TrackerSnapshotTest::setUp() must be compatible with Yoast\PHPUnitPolyfills\TestCases\TestCase::setUp(): void in /abc/wp-content/plugins/pinterest-for-woocommerce/tests/Unit/TrackerSnapshotTest.php on line 24
```
- Return type for method `perform_plugin_update_procedure` is void so `$mock_plugin_update->method('perform_plugin_update_procedure')->willReturn( null );` gives the following errors. I was able to fix this by removing `->willReturn( null )` as we shouldn't be returning anything.
```
1) Automattic\WooCommerce\Pinterest\Tests\Unit\PluginUpdate\Pinterest_Test_Plugin_Update::testUpdateFlowNoThrow
PHPUnit\Framework\MockObject\IncompatibleReturnValueException: Method perform_plugin_update_procedure may not return value of type string, its declared return type is "void" /abc/wp-content/plugins/pinterest-for-woocommerce/tests/Unit/PluginUpdateTest.php:128

2) Automattic\WooCommerce\Pinterest\Tests\Unit\PluginUpdate\Pinterest_Test_Plugin_Update::testAfterUpdateTheUpdateIsNotExecutedAgain
PHPUnit\Framework\MockObject\IncompatibleReturnValueException: Method perform_plugin_update_procedure may not return value of type NULL, its declared return type is "void" /abc/wp-content/plugins/pinterest-for-woocommerce/tests/Unit/PluginUpdateTest.php:183
```
- `unset( $settings['local_feed_id'] );` resulted in the following warning during testing so added a condition to check if `$settings['local_feed_id']` is set and only then unset it.
```
PHP Deprecated:  Automatic conversion of false to array is deprecated in /home/runner/work/pinterest-for-woocommerce/pinterest-for-woocommerce/src/PluginUpdate.php on line 273
```
- Disabling WooCommerce.Commenting.CommentHooks.MissingSinceComment checks for existing hooks since it will be difficult to trace when they were added.

### Changelog entry

> Dev - Updates PHPUnit version, unit tests, and matrix in git workflow for PHP8.2 compatibility.

### DIFF
--- a/.github/workflows/php-coding-standards.yml
+++ b/.github/workflows/php-coding-standards.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Prepare PHP
         uses: woocommerce/grow/prepare-php@actions-v1
         with:
-          php-version: '7.3'
+          php-version: '8.2'
           tools: cs2pr
 
       - name: Log PHPCS debug information
--- a/.github/workflows/php-unit-tests.yml
+++ b/.github/workflows/php-unit-tests.yml
@@ -30,7 +30,7 @@ jobs:
       WP_TESTS_DIR: "/tmp/wordpress/tests/phpunit"
     strategy:
       matrix:
-        php: [7.3, 7.4, 8.0, 8.1, 8.2]
+        php: [7.4, 8.0, 8.1, 8.2]
         wp-version: [latest]
 
     steps:
--- a/.github/workflows/php-unit-tests.yml
+++ b/.github/workflows/php-unit-tests.yml
@@ -32,11 +32,6 @@ jobs:
       matrix:
         php: [7.3, 7.4, 8.0, 8.1, 8.2]
         wp-version: [latest]
-        include:
-          - php: 7.4
-            wp-version: 5.9
-          - php: 7.3
-            wp-version: 5.9
 
     steps:
       - name: Checkout repository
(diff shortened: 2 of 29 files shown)


message-dimke (approved):
Approved! Thank you, @krutidugade !

Issue #739, opened 2023-04-25
Pinterest v5 API Epic.


budzanowski:
This is the Epic for the Pinterest v5 API migration work:

The integration branch for all of the code work will be [pinterest-v5-integration-branch](https://github.com/woocommerce/pinterest-for-woocommerce/tree/pinterest-v5-integration-branch)

```[tasklist]
### Tasks
- [ ] #738
- [ ] https://github.com/woocommerce/pinterest-for-woocommerce/issues/743
- [ ] https://github.com/woocommerce/pinterest-for-woocommerce/issues/745
- [ ] https://github.com/Automattic/connect.woocommerce.com/pull/165
- [ ] https://github.com/woocommerce/pinterest-for-woocommerce/issues/748
- [ ] https://github.com/woocommerce/pinterest-for-woocommerce/issues/752
- [ ] https://github.com/woocommerce/pinterest-for-woocommerce/issues/757
- [ ] https://github.com/woocommerce/pinterest-for-woocommerce/issues/831
- [ ] https://github.com/woocommerce/pinterest-for-woocommerce/issues/833
- [ ] https://github.com/woocommerce/pinterest-for-woocommerce/issues/853
- [ ] https://github.com/woocommerce/pinterest-for-woocommerce/issues/848
- [ ] https://github.com/woocommerce/pinterest-for-woocommerce/issues/840
- [ ] https://github.com/woocommerce/pinterest-for-woocommerce/issues/844
- [ ] Display pinterest connection modal in a popover.
- [ ] https://github.com/woocommerce/pinterest-for-woocommerce/issues/856
```

PR #841, opened 2023-11-06
Add token exchange endpoint

### LINKED ISSUE #840: Add V3 to V5 update endpoint.
Create API request method that will allow triggering the token ( V3 -> V5 ) exchange procedure.

budzanowski:

This introduces a class that is able to exchange the V3 token for a V5 token.

Closes #840 .

This adds the API method for the token exchange procedure.
A separate class has been added because we need to ping the v3 endpoint 
so we need the constant to point to V3. We don't want to use the Base class bc it will be removed.
Besides the endpoint, this will also be the location for the update procedure method.

No testing for now. This will be testable when the update procedure will be implemented.

### Additional details:

This is a part of the #739 

### Changelog entry

> No changelog required.

### DIFF
--- a/src/API/TokenEchangeV3ToV5.php
+++ b/src/API/TokenEchangeV3ToV5.php
@@ -1,0 +1,69 @@
+<?php
+/**
+ * Pinterest API V5 class
+ *
+ * @class       Pinterest_For_Woocommerce_API
+ * @version     x.x.x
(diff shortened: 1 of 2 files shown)


message-dimke (approved):
LGTM!

Thank you, @budzanowski

PR #845, opened 2023-11-07
Add token exchange update procedure.

### LINKED ISSUE #844: Add V5 update procedure.

budzanowski:

When the plugin gets updated the token exchange procedure should be triggered.

Closes #844 .

Relies on #840.

1. Install the current production version and connect.
2. Complete the setup and connection.
3. Switch to this branch. Update manually `PINTEREST_FOR_WOOCOMMERCE_VERSION` to `1.4.0`
4. Add a breakpoint at the update method. Reload the plugin.
5. Observer the flow.

### DIFF
--- a/src/API/TokenExchangeV3ToV5.php
+++ b/src/API/TokenExchangeV3ToV5.php
@@ -43,5 +43,25 @@ class TokenExchangeV3ToV5 extends APIV5 {
 		$request_url = 'oauth/commerce_integrations/token/exchange/';
 		return self::make_request( $request_url );
 	}
+	/**
+	 * Update token from V3 to V5.
+	 *
+	 * @since x.x.x
+	 *
+	 * @return bool $success Whether the token was updated successfully.
+	 */
+	public static function token_update() {
+		$respone = self::exchange_token();
+
+		if ( 'success' != $respone['status'] ) {
+			return false;
+		}
+
+		$token_data = $respone['data'];
+
+		Pinterest_For_Woocommerce()::save_token_data( $token_data );
+
+		return true;
+	}
 
 }
--- a/src/API/TokenExchangeV3ToV5.php
+++ b/src/API/TokenExchangeV3ToV5.php
@@ -64,4 +64,15 @@ class TokenExchangeV3ToV5 extends APIV5 {
 		return true;
 	}
 
+	/**
+	 * Get the V3 token.
+	 *
+	 * @since x.x.x
+	 *
+	 * @return string $token The V3 token.
+	 */
+	public static function get_token() {
+		return Pinterest_For_Woocommerce()::get_data( 'token', true );
+	}
+
 }
--- a/src/API/TokenExchangeV3ToV5.php
+++ b/src/API/TokenExchangeV3ToV5.php
@@ -9,6 +9,9 @@
 
 namespace Automattic\WooCommerce\Pinterest\API;
 
+use Automattic\WooCommerce\Pinterest\Crypto;
+use Automattic\WooCommerce\Pinterest\Logger;
+
 if ( ! defined( 'ABSPATH' ) ) {
 	exit;
 }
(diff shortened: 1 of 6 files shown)


src/PluginUpdate.php
```diff
@@ -288,4 +293,30 @@ protected function ads_credits_integration(): void {
...
+	/**
+	 * Token update procedure. V3 to V5.
+	 *
+	 * @since 1.4.0
+	 *
+	 * @return void
+	 */
+	protected function token_update(): void {
+		// Update should only happen if the plugin is connected using the V3 token.
+		if ( ! Pinterest_For_Woocommerce()::is_connected() ) {
```
message-dimke: We have a problem here, since current branch is based of API v5 integration branch, `is_connected()` is a method to check if **v5** is connected and not **v3**. We need to add something like `is_v3_connected` because **v3** token is stored in a bit different data structure. 

**v3** token is stored as `[pinterest_for_woocommerce_data][token][access_token]`, when **v5** is `[pinterest_for_woocommerce_data][token_data][access_token]`

Not sure this is a good decision, but it is what it is =)

message-dimke: And because we store data between versions under different keys, it is totally possible to have both the same time while testing. It may explain why it worked for you and did not for me.

budzanowski: > And because we store data between versions under different keys, it is totally possible to have both at the same time while testing. It may explain why it worked for you and did not for me.

Most likely that. I was also changing some keys in the DB directly so maybe this made it work.

Good catch!

> v3 token is stored as [pinterest_for_woocommerce_data][token][access_token], when v5 is [pinterest_for_woocommerce_data][token_data][access_token]
>
> Not sure this is a good decision, but it is what it is =)

I have changed the name to indicate that now we store not only the token but the refresh token and other data as well.

I have updated the procedure. Please retry.

message-dimke:
Hey, @budzanowski ! 
There is one more issue with the API call headers.

Here:

https://github.com/woocommerce/pinterest-for-woocommerce/blob/9aed51be3cec77f6ab90bc48812b539d4942d6ab/src/API/Base.php#L288-L290

`self::get_token()` uses `Pinterest_For_Woocommerce()::get_access_token()` which returns no token since is looking for **v5** token keys not **v3**.

As a result of `Authorization: Bearer ` I am getting `401 Reconnect to your Pinterest account.`, access token is not added to the header.

We can redo `Base::get_token()` to be something like you did recently

```php
public static function get_token() {
	if ( is_null( self::$token ) ) {
		self::$token = Pinterest_For_Woocommerce()::get_data( 'token', true );
	}

	return self::$token;
}
```

budzanowski:
Sorry for the hassle. I was too lazy to constantly change branches so I was injecting the v3 token manually for testing.

This latest change https://github.com/woocommerce/pinterest-for-woocommerce/pull/845/commits/ae222589fa15dc74ad018b3fb46c0ba721213299 should fix the issue.

src/API/TokenExchangeV3ToV5.php
```diff
@@ -43,5 +43,36 @@ public static function exchange_token() {
...
+
+	/**
+	 * Get the V3 token.
+	 *
+	 * @since x.x.x
+	 *
+	 * @return string $token The V3 token.
+	 */
+	public static function get_token() {
+		return Pinterest_For_Woocommerce()::get_data( 'token', true );
```
message-dimke: Hey, @budzanowski 

The last change.

We need to `Pinterest/Crypto::decrypt()` our `[access_token]`.

```suggestion
		$token = Pinterest_For_Woocommerce()::get_data( 'token', true );
		try {
            $token['access_token'] = empty( $token['access_token'] ) ? '' : Pinterest\Crypto::decrypt( $token['access_token'] );
        } catch ( \Exception $th ) {
            /* Translators: The error description */
            Pinterest\Logger::log( sprintf( esc_html__( 'Could not decrypt the Pinterest API access token. Try reconnecting to Pinterest. [%s]', 'pinterest-for-woocommerce' ), $th->getMessage() ), 'error' );
        }
        return $token;
```

budzanowski: Good catch. I had my access token in the DB in a plane text.

message-dimke (approved):
Works as expected. Token exchange endpoint succeeds and we get v5 token from it. Also it gets saved into a database.

Thank you, @budzanowski !

PR #847, opened 2023-11-10
Check feed_location to contain absolute url.

### LINKED ISSUE #838: The feed_location should always be absolute URL
### Describe the bug:
v3_create_commerce_partner_merchant call made by the app. The Error is of type Api request failed. Cannot determine feed type from format [XML] and url [/wp-content/uploads/pinterest-for-woocommerce-t8OBKI.xml].
It looks like a small percentage of requests made are with relative URLs [during merchant creation](https://github.com/woocommerce/pinterest-for-woocommerce/blob/develop/src/Merchants.php#L128) where as the API expects absolute URL.

### Steps to reproduce:
1.  I am not sure on how to reproduce but you can easily reproduce by sending a relative URL to the API for a new merchant

### Expected behavior:
Convert relative URLs to absolute URLs. If unable to convert show a error to user to change settings

### Actual behavior:
Is sent as relative URL

### Additional details:
Currently call is made using args
```
$args = array(
			'merchant_domains' => get_home_url(),
			'feed_location'    => $config['feed_url'],
			'feed_format'      => 'XML',
			'country'          => Pinterest_For_Woocommerce()::get_base_country() ?? 'US',
			'locale'           => LocaleMapper::get_locale_for_api(),
			'currency'         => get_woocommerce_currency(),
			'merchant_name'    => $merchant_name,
		);
```
The feed_location needs to be always absolute URL because merchant_domains is an array and there can be more than one value and no other field currently is able to give the correct schema and domain.

message-dimke:

Closes #838 .

Adding extra check for `feed_location` to be a full URL and not just a relative path.
If `feed_location` does not contain a domain - add current domain to it.

There are no test instructions because the issue is not easily reproducible. 
Pinterest reported only some websites have this. 

### Changelog entry

> Tweak - Make sure `feed_location` has a full URL.

### DIFF
--- a/src/Merchants.php
+++ b/src/Merchants.php
@@ -139,9 +139,13 @@ class Merchants {
 		 */
 		$merchant_name = apply_filters( 'pinterest_for_woocommerce_default_merchant_name', esc_html__( 'Auto-created by Pinterest for WooCommerce', 'pinterest-for-woocommerce' ) );
 
+		// Check if the feed location is a full URL or a relative path and build the feed location accordingly.
+		$feed_location = parse_url( $config['feed_url'] );
+		$feed_location = ! empty( $feed_location['host'] ) ? $config['feed_url'] : get_home_url() . $feed_location['path'];
+
 		$args = array(
 			'merchant_domains' => get_home_url(),
-			'feed_location'    => $config['feed_url'],
+			'feed_location'    => $feed_location,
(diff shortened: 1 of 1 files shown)


budzanowski (approved):
I guess it may still fail in case that there is something really wonky with the setup and the manual creation of the URL will create something strange. But still this tries to improve things so 🛳️

PR #871, opened 2023-12-21
Fixing wrong feed items counter.

message-dimke:

Moving items counter one level up to prevent from being incremented when looping through the array of feed configurations.

Note: the issue is reproduceable when you have multiple local feed configurations. Also, in some cases, especially with fresh installations, we have a bug in the system when we always have multiple local feed configurations where the first one is `false`, the sign of this is when you have your feed file ending with `...-.csv`, w/o a random suffix. 

1. Being at develop branch go to WooCommerce - Status - Scheduled Actions - Pending section.
2. Run `pinterest-for-woocommerce-start-feed-generation` action.

Cursor_and_WooCommerce_status_‹_WordPress_Pinterest_—_WordPress

3. Go to Marketing - Pinterest tab and check counters. You should see items counter to show twice the amount of products added to the feed file when the original file has a proper number of products.

cursor_and_products_catalog_e280b9_pinterest_e280b9_marketing_e280b9_woocommerce_e280b9_wordpress_pinterest_e28094_woocommerce

4. Checkout `fix/wrong-feed-items-counter` branch and repeat all the steps above.
5. This time you should see the correct counter instead.

### Changelog entry

> Fix - Wrong feed generation items counter.

### DIFF
--- a/src/FeedGenerator.php
+++ b/src/FeedGenerator.php
@@ -394,8 +394,8 @@ class FeedGenerator extends AbstractChainedJob {
 					continue;
 				}
 				$this->buffers[ $location ] .= $product_xml;
-				++$processed_products;
 			}
+			++$processed_products;
 		}
 
 		// May throw write to file exception.


budzanowski (approved):
Nice. Clearly a mistake on my end :)

:shipit:

PR #872, opened 2023-12-26
Removing redundant custom transient notices.

### LINKED ISSUE #867: Notices styling is too generic - Affects all WooCommerce admin notices
I think this is self explanatory

Pinterest targets and applies styles to `.woocommerce-transient-notices`, a very generic class.

This interferes with the new product editor, as the notices are hidden behind the menu

image

If we deactivate those styles, the notices appear in the correct position

image
 

Consider adding a parent to the custom styles, so that Tik Tok targets only notices on their settings pages, and not everywhere in WooCommerce admin

message-dimke:

Closes #867 .

Removing TransientNotice component. Having it causes transient notice appear twice and Pinterest extension styles for the notice override WooCommerce global styles for other pages making notices shift left. 

The reason for that is `TransientNotice` component and `createNotice()` JS API method both dispatch the same action, my personal belief is that `TransientNotice` was added because of a lack of knowledge or before `createNotice()` became available.

Removing the component shots two issues at once: double notice bug ( when at Pinterest extension pages ) and new experimental Product Editor notices shifted to the left.

#### Double notice issue.

Settings_‹_Pinterest_‹_Marketing_‹_WooCommerce_‹_WordPress_Pinterest_—_WooCommerce

1. Install recent Pinterest plugin version and connect.
2. Being at Settings tab click 'Sync' ( top right ). 
3. Observe the notice about sync status appears bottom left. 
4. Check with Web Inspector there are two notices on a page ( even visually it is darker than the one inside Product Editor, because there are actually two the same notices one above the other ). Also those notices aren't next to each other in the source code, so finding both occurrences requires some search while the notices are still visible. 
5. Checkout `fix/remove-custom-transient-notices` branch 
6. run `nvm use && npm run build` to have new sources built. 
7. Being at Settings tab click 'Sync' ( top right ). 
8. Check there is only a single notice copy on the page.

#### Experimental Product Editor transient notice shift.

To reproduce this one and test the fix you will need a `wpcomstaging.com` website. 

1. Setup the website.
2. Enable experimental Product Editor.

WooCommerce_settings_‹_TheWayWeBook_—_WordPress

3. Add new or update an existing product.
4. Observe the transient notice bottom left. It must not overlap with the sidebar.

5. Install Pinterest extension ( no need to connect to Pinterest, we are testing CSS only )
6. Add new or update an existing product. 
7. Observe the transient notice bottom left. It overlaps with the sidebar.

‹_TheWayWeBook-2

8. Checkout `fix/remove-custom-transient-notices` branch.
9. Run `nvm use && npm run build:zip` to produce a ZIP file.
10. Uninstall the existing Pinterest plugin and install the one out the ZIP file.
11. Add or update a product using new experimental Product Editor.
12. Observe the transient notice bottom left. It must not overlap with the sidebar.

‹_TheWayWeBook-3

### Changelog entry

> Fix - Product Editor transient notice shift.
> Fix - Pinterest duplicated notices.


budzanowski (approved):
This is a great change :)

Just remove the .pot changes and this is good to go.

PR #885, opened 2024-01-14
Integrating v5 billing endpoints.

message-dimke:
### Pinterest Billing API

The task is to research the possibility to integrate newly suggested Billing APIs into the extension. The main question is can we easily replace all the places where we have calls to the old API with the new ones considering that we have four old API endpoints and Pinterest suggests only three new.

### The research.

`POST /ads/v4/advertisers/<advertiser-id>/marketing_offer/<offer-code>/redeem`
Used in two places: to validate if the code is valid (when called with parameter ‘validate only’) and to actually redeem the code.

Studying the codebase I did not manage to find a single place where validation is happening which makes me think of a validate_ads_offer_code function being redundant and not used anymore.

However, it is still used for redeeming a code and what we need from it is a status field to check if the action was successful and if not – error details from the response, which are going to be present if the request did not succeed.

According to the Pinterest new endpoint description both scenarios are supported by the new endpoint so the endpoint can be easily replaced.

The new endpoint’s response is missing success property which the old endpoint provides, instead an HTTP status code is used and is 200 if the action was successful or 4xx with the details on the error otherwise.

`GET /ads/v4/advertisers/<advertiser-id>/discounts`
Current endpoint is not described under either v3 or v4 endpoint documentations ( [https://developers.pinterest.com/docs/api/v3/](https://href.li/?https://developers.pinterest.com/docs/api/v3/), [https://developers.pinterest.com/docs/api/v4/](https://href.li/?https://developers.pinterest.com/docs/api/v4/) ).

From the code I can only guess why we search for specific response elements under indexes of 5 and 16 to work with. The new endpoint returns an array of items and from the document I can tell that each item returned has the properties we need to integrate the call into the extensions’s codebase.

`GET /ads/v4/advertisers/<advertiser-id>/partners/billing_profiles`
We use the endpoint to detect if billing was setup. We do not. use any other information returned by the endpoint and the new endpoint suggested by Pinterest has the data we need to determine the same (new API has a flag to include only active profiles and depending on the number of returned billing profile we know what we need).

`GET /ads/v4/advertisers/<advertiser-id>/billing_data`
Billing data endpoint is an analog of billing profiles endpoint and according to my research is not used by the extension. get_advertiser_billing_data can be removed from the codebase.

### DIFF
--- a/src/API/APIV5.php
+++ b/src/API/APIV5.php
@@ -543,4 +543,33 @@ class APIV5 extends Base {
 			'GET'
 		);
 	}
+
+	/**
+	 * Attempts to redeem the offer code for the given advertiser.
+	 *
+	 * @link https://developers.pinterest.com/docs/api/v5/#operation/ads_credit/redeem
+	 *
+	 * @since x.x.x
+	 *
+	 * @param string $advertiser_id Advertiser ID.
+	 * @param string $offer_code    Offer code (hash).
+	 *
+	 * @return array {
+	 *     Credits redeem status.
+	 *
+	 *     @type bool    $success       Whether the offer code was successfully redeemed or not.
+	 *     @type ?int    $errorCode     Error code type if error occurs.
+	 *     @type ?string $errorMessage  Reason for failure.
+	 * }
+	 * @throws PinterestApiException When unable to redeem the offer code or any unexpected error occur.
+	 */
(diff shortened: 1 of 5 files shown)


message-dimke:
Merging into `pinterest-v5-integration-branch` branch.

PR #876, opened 2024-01-02
Adding currency credit info to account data.

message-dimke:
### Changes proposed in this Pull Request:

Currency credit information data is required for the UI to display promo using proper currencies and amounts per locale. For domain verification step the information was missing causing page to go blank once verification finishes. That was all due to domain verification step was also fetching account data from Pinterest and updated it into options database table vanishing previously set currency credit information. 

Further research showed that there is a function used as a callback to init action to fetch currency information and save into account data. Debugging the function calls showed it was called up to 10 times per page load, updated it to alter account data only if the key is missing from the data set.

The solution is to initialise currency credit information setup to token save and domain verification steps, the same place other account data keys are initialised and assigned.

### Detailed test instructions:

1. Being at `develop` branch.
2. Get to `wp_options` database table and search for `pinterest_for_woocommerce` option name
3. Update the option value `verified_user_websites` subset by removing tested website domain from the list.
4. Refresh the page to see `Start verification` button.

Onboarding_Guide_‹_Pinterest_‹_Marketing_‹_WooCommerce_‹_WordPress_Pinterest_—_WooCommerce

5. Press the button and see blank page display.

Onboarding_Guide_‹_Pinterest_‹_Marketing_‹_WooCommerce_‹_WordPress_Pinterest_—_WooCommerce

Note: There is a console error _Cannot read properties of undefined (reading 'creditsGiven')_

DevTools_-_pinterest_dima_works_wp-admin_admin_php_page_wc-admin_path__2Fpinterest_2Fonboarding_step_claim-website_view_wizard

Note: payload does not contain `currency_credit_info`

DevTools_-_pinterest_dima_works_wp-admin_admin_php_page_wc-admin_path__2Fpinterest_2Fonboarding_step_claim-website_view_wizard

6. Checkout `fix/currency-credit-info-missing` branch.
7. Get to `wp_options` database table and search for `pinterest_for_woocommerce` option name
8. Update the option value `verified_user_websites` subset by removing tested website domain from the list.
9. Refresh the page to see `Start verification` button.
10. This time you should see verification process successfully finish, no console error and the response payload to have proper data subset under `account_data` key.

Onboarding_Guide_‹_Pinterest_‹_Marketing_‹_WooCommerce_‹_WordPress_Pinterest_—_WooCommerce

### Changelog entry

> Fix - Currency and credit information missing at domain verification step.

PR #902, opened 2024-02-15
Make sure currency_credit_info is always added to account_data

### LINKED ISSUE #899: JS error after URL verification
### Describe the bug:
After I verify my URL, I'm dumped onto a blank page:

   - The error is `TypeError: Cannot read properties of undefined (reading 'creditsGiven')`
   image
   `("As a new Pinterest customer, you can get %1$s in free ad credits when you successfully set up Pinterest for WooCommerce and spend %2$s on Pinterest Ads. <a>Pinterest Terms and conditions</a> apply.", "pinterest-for-woocommerce"), a.creditsGiven, a.spendRequire)`

Refreshing shows a verified site and I can continue with the onboarding.

### Steps to reproduce:
1. Do onboarding. WC 8.5 or 8.6. Store not fully set up. I was using `ddev share` to get an ngrok site.
2. Verify a domain.

### Expected behavior:
Continue onboarding flow

### Actual behavior:
Blank screen until referesh

layoutd:

Closes https://github.com/woocommerce/pinterest-for-woocommerce/issues/899.

Reported and incorrectly labeled as fixed in https://wordpress.org/support/topic/start-verification-issue/ and https://wordpress.org/support/topic/black-after-claim-your-site/.

As seen in #899, the response to the API call to `/wp-json/pinterest/v1/domain_verification` causes a JS error because the `currency_credits_info` is missing.

That info is added to the existing `account_data` on `init`: https://github.com/woocommerce/pinterest-for-woocommerce/blob/81b703bcf42682bbe07abb5c22b8c144f1434052/class-pinterest-for-woocommerce.php#L287-L288

But during DomainVerification, this data is updated again:
https://github.com/woocommerce/pinterest-for-woocommerce/blob/81b703bcf42682bbe07abb5c22b8c144f1434052/src/API/DomainVerification.php#L82

Which rebuilds the account data WITHOUT the `currency_credits_info`:
https://github.com/woocommerce/pinterest-for-woocommerce/blob/81b703bcf42682bbe07abb5c22b8c144f1434052/class-pinterest-for-woocommerce.php#L912-L942

That breaks the onboarding process. But refreshing runs the init hook again, so the settings are updated and the onboarding can finish.

This PR ensures that, any time the `account_data` is updated **after** the `init` hook has already been executed, the `currency_credits_info` is included. That way the `domain_verification` call returns all the necessary info. (The only other time the `update_account_data` method is used is right after connecting the account, between Steps 1 and 2 of onboarding, so there's no downside to adding the data there).

1. Go through onboarding and ensure that domain verification works correctly.
2. Quick smoke test (update user settings, etc) to make sure everything runs smoothly.
3. 

### Changelog entry

> Fix - Blank screen in onboarding after domain verification.

### DIFF
--- a/class-pinterest-for-woocommerce.php
+++ b/class-pinterest-for-woocommerce.php
@@ -939,7 +939,7 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 					$data['coupon_redeem_info'] = array( 'redeem_status' => false );
 
 					Pinterest_For_Woocommerce()::save_setting( 'account_data', $data );
-					return $data;
+					return Pinterest_For_Woocommerce()::add_currency_credits_info_to_account_data();
 				}
 
 				self::get_linked_businesses( true );
@@ -1051,13 +1051,14 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 		 *
 		 * @since 1.3.9
 		 *
-		 * @return void
+		 * @return array The updated account data.
 		 */
 		public static function add_currency_credits_info_to_account_data() {
 			$account_data                         = self::get_setting( 'account_data' );
 			$currency_credit_info                 = AdsCreditCurrency::get_currency_credits();
 			$account_data['currency_credit_info'] = $currency_credit_info;
 			self::save_setting( 'account_data', $account_data );
+			return $account_data;
 		}
 
 		/**


budzanowski (commented):
Looks good!

layoutd:
This is a repeat of #876, that one was first so closing in favor of that solution.

PR #911, opened 2024-02-25
Defer events affected by WC settings only.

### LINKED ISSUE #910: Delay only some of the Pinterest Tag events.
### Describe the bug:

Pinterest Tag events are events that are sent to Pinterest using JavaScript APIs. For that to happen, the JavaScript code must get rendered during a page render. Render is set to happen in the page header, so the events collected on the page can't be rendered. Even if we move render to the page footer, we are still saving all the collected page events at the `shutdown` WordPress event to prevent multiple database updates per page load. 

Some WooCommerce flows can be configured to use `Redirect to the cart page after successful addition`, which, if applied, will perform a redirect and prevent events from rendering. 

The issue reported by the Pinterest team is that events sometimes do not render when they should. This is because all Pinterest Tag events are delayed and so the behavior seems confusing. Not all events require a delay since the setting clearly states the redirect relates to the `Add to Cart` event and must delay only it when rendering the others immediately.

message-dimke:

Closes #910 .

Not all Pinterest Tag events have to be deferred. Since the only event affected by WC settings is the 'Add to Cart' event - defer it and fire the others asap.

Whether these are old shortcodes used for the Cart and the Checkout page or the new Block ones, both scenarios should work the same. As a prerequisite, Pinterest Tag Chrome extension to track events must be installed and used.

#### Tests with 'Redirect to the cart page after successful addition' is ON:

1. Shop page.

	- Click 'Add to Cart' next othe any product on the page.
	- Observe browser navigates to the cart page. The Add to Cart event is fired.
	- Click 'Procceed to Checkout' button.
	- At checkout page fill in all the required fields and press 'Place Order'.
	- Observe browser navigates to 'Thank You' page and 'Checkout' event is fired.

2. Product page.

	- Click 'Add to Cart' next to the product.
	- Observe 'Add to Cart' event is fired.
	- Navigate to the Cart page.
	- Click 'Procceed to Checkout' button.
	- At checkout page fill in all the required fields and press 'Place Order'.
	- Observe browser navigates to 'Thank You' page and 'Checkout' event is fired.

#### Tests with 'Redirect to the cart page after successful addition' is OFF:

1. Shop page.

	- Click 'Add to Cart' next othe any product on the page.
	- Observe the Shop page reloaded and 'Add to Cart' event fired.
	- Navigate to the Cart page.
	- Click 'Procceed to Checkout' button.
	- At checkout page fill in all the required fields and press 'Place Order'.
	- Observe browser navigates to 'Thank You' page and 'Checkout' event is fired.

2. Product page.

	- Click 'Add to Cart' next to the product.
	- Observe 'Add to Cart' event is fired.
	- Navigate to the Cart page.
	- Click 'Procceed to Checkout' button.
	- At checkout page fill in all the required fields and press 'Place Order'.
	- Observe browser navigates to 'Thank You' page and 'Checkout' event is fired.

Both cases should also work for the Product Archive page.

### Changelog entry

> Update - Defer only the 'Add to Cart' event when redirect to cart is on.

### DIFF
--- a/src/Tracking.php
+++ b/src/Tracking.php
@@ -53,13 +53,13 @@ class Tracking {
 		$this->trackers = $trackers;
 
 		// Tracks page visit events.
-		add_action( 'wp_footer', array( $this, 'handle_page_visit' ) );
+		add_action( 'wp_footer', array( $this, 'handle_page_visit' ), 0 );
 
 		// Tracks category visit events.
-		add_action( 'wp_footer', array( $this, 'handle_view_category' ) );
+		add_action( 'wp_footer', array( $this, 'handle_view_category' ), 0 );
 
 		// Tracks search events.
-		add_action( 'wp_footer', array( $this, 'handle_search' ) );
+		add_action( 'wp_footer', array( $this, 'handle_search' ), 0 );
 
 		// Tracks add to cart events.
 		add_action( 'woocommerce_add_to_cart', array( $this, 'handle_add_to_cart' ), 10, 6 );
--- a/src/Tracking.php
+++ b/src/Tracking.php
@@ -38,7 +38,7 @@ class Tracking {
 	const EVENT_VIEW_CATEGORY = 'ViewCategory';
 
 	/**
-	 * @var array $trackers A list of available trackers.
+	 * @var Tracker[] $trackers A list of available trackers.
 	 */
 	private $trackers = array();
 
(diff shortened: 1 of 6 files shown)


budzanowski (approved):
The priority 0 is quite agressive. If the script is in the footer does it matter when? Would any priority be enough? 

Approving in case we need 0. if not please adjust and merge.

budzanowski (approved):
Looks good after the updates.

Thank you! :shipit:

PR #914, opened 2024-02-25
CAPI calls with Ad Account ID is missing.

### LINKED ISSUE #913: CAPI events are sent even is Ad Account ID is missing.
### Describe the bug:

Reported by Pinterest team.

After disconnecting Pinterest account CAPI events are sent to Pinterest even if Ad Account ID information was deleted and there is no ID required to perform the request.

message-dimke:

Closes #913 .

Pinterest detected some CAPI requests with missing Ad Account ID (which is a required parameter). Not sure what were the circumstances of those calls, but making sure they will never happen is Ad Account ID is missing.

### Changelog entry

> Fix - CAPI calls with empty Ad Account ID.

### DIFF
--- a/src/Tracking/Conversions.php
+++ b/src/Tracking/Conversions.php
@@ -266,6 +266,9 @@ class Conversions implements Tracker {
 	 */
 	private function send_request( string $event_name, array $data ) {
 		$ad_account_id = Pinterest_For_WooCommerce()::get_setting( 'tracking_advertiser' );
+		if ( empty( $ad_account_id ) ) {
+			return;
+		}
 
 		/* translators: 1: Conversions API event name, 2: JSON encoded event data. */
 		$messages = sprintf(


src/Tracking/Conversions.php
```diff
@@ -266,6 +266,9 @@ private function get_search_data( Search $data ) {
 	 */
 	private function send_request( string $event_name, array $data ) {
 		$ad_account_id = Pinterest_For_WooCommerce()::get_setting( 'tracking_advertiser' );
+		if ( empty( $ad_account_id ) ) {
```
budzanowski, line 269: Should this maybe happen at the `Tracking:track_event` level? The tag should also not be sent if there is no ad_account_id.

message-dimke: Hey, @budzanowski !

Tracking class is responsible for notifying exact trackers about events happening. Pinterest Tag tracker does not required Ad Account ID when sending events for example (it initialises with Tracking Tag ID). If I block events at this level I block all the trackers. Adding conditional statement to filter out specific trackers IMHO is ugly. So I silence only those trackers which have requests dependent on Ad Account ID. What do you think?

budzanowski: I thought about stopping any tracking in case account_id is missing. I thought that even if Tag does not require account_id it would still make sense to stop the events because there is something wrong.

But it looks like this is good so we can roll with that.

budzanowski (approved):

PR #916, opened 2024-02-28
Disable a tracker hooks on removal.

### LINKED ISSUE #915: Hooks are not disabled when a tracker is removed.
### Describe the bug:

The tracking system supports multiple trackers, which may or may not hook into WP/WC actions and filters. Individual trackers can be added/removed from the system. When added, a tracker may set its hooks, but when removed, no hooks get disabled. Adding disable hooks procedure on tracker removal.

This is a pure dev thing.

message-dimke:

Closes #915 .

The tracking system supports multiple trackers, which may or may not hook into WP/WC actions and filters. Individual trackers can be added/removed from the system. When added, a tracker may set its hooks, but when removed, no hooks get disabled. Adding disable hooks procedure on tracker removal.

### Changelog entry

> Dev - Disable tracker hooks on the tracker removal.

### DIFF
--- a/src/Tracking.php
+++ b/src/Tracking.php
@@ -284,7 +284,11 @@ class Tracking {
 		$this->trackers = array_filter(
 			$this->trackers,
 			function ( $item ) use ( $tracker ) {
-				return get_class( $item ) !== $tracker;
+				if ( get_class( $item ) === $tracker ) {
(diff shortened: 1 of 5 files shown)


budzanowski (approved):
Looks good. Good cleanup always makes sense.

PR #925, opened 2024-03-06
922 retry oauth

### LINKED ISSUE #922: Token exchange procedure should retry after a failure.
Whenever automatic plugin update procedure fails we should retry in a couple of minutes.

budzanowski:

Closes #922 .

When token procedure fails retry 3 more times in  5, 10 and 15 minutes.

Testing:
- Connect v3 version and authenticate.
- Update with this version but use broken version ( tamper with the oauth url ).
- After the first failure happens an action for retry will be scheduled
- Replace the broken version with the working version and wait until the action gets triggered, or trigger it manually
- Verify that the token exchange worked.

### DIFF
--- a/class-pinterest-for-woocommerce.php
+++ b/class-pinterest-for-woocommerce.php
@@ -18,6 +18,7 @@ use Automattic\WooCommerce\Pinterest\FeedRegistration;
 use Automattic\WooCommerce\Pinterest\Heartbeat;
 use Automattic\WooCommerce\Pinterest\Logger;
 use Automattic\WooCommerce\Pinterest\Notes\MarketingNotifications;
+use Automattic\WooCommerce\Pinterest\Notes\TokenExchangeFailure;
 use Automattic\WooCommerce\Pinterest\PinterestApiException;
 use Automattic\WooCommerce\Pinterest\Tracking;
 use Automattic\WooCommerce\Pinterest\Tracking\Conversions;
@@ -299,6 +300,7 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 			add_action( 'pinterest_for_woocommerce_token_saved', array( $this, 'create_commerce_integration' ) );
 			add_action( 'pinterest_for_woocommerce_token_saved', array( $this, 'update_account_data' ) );
 			add_action( 'pinterest_for_woocommerce_token_saved', array( $this, 'update_linked_businesses' ) );
+			add_action( 'pinterest_for_woocommerce_token_saved', array( $this, 'post_update_cleanup' ) );
 
 			// Handle the Pinterest verification URL.
 			add_action( 'parse_request', array( $this, 'verification_request' ) );
@@ -613,6 +615,17 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 		 */
 		public static function set_api_version( $version ) {
 			return update_option( PINTEREST_FOR_WOOCOMMERCE_PINTEREST_API_VERSION, $version );
+		}
+
+		/**
+		 * Get API version used by the plugin.
+		 *
+		 * @since x.x.x
+		 *
+		 * @return string The API version.
+		 */
+		public static function get_api_version() {
+			return get_option( PINTEREST_FOR_WOOCOMMERCE_PINTEREST_API_VERSION, '' );
 		}
 
 		/**
@@ -1115,6 +1128,21 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 		 */
 		public static function update_linked_businesses() {
 			self::get_linked_businesses( true );
+		}
+
+		/**
+		 * Cleanup after the token update.
+		 *
+		 * @since x.x.x
+		 *
+		 * @return void
+		 */
+		public static function post_update_cleanup() {
+			// Mark note as actioned so it will be removed from the UI in case it was added.
+			TokenExchangeFailure::possibly_action_note();
+
+			// Update completed successfully.
+			Pinterest_For_Woocommerce()::set_api_version( 'v5' );
 		}
 
 		/**
--- a/class-pinterest-for-woocommerce.php
+++ b/class-pinterest-for-woocommerce.php
@@ -1138,8 +1138,7 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 		 * @return void
 		 */
 		public static function post_update_cleanup() {
-			// Mark note as actioned so it will be removed from the UI in case it was added.
-			TokenExchangeFailure::possibly_action_note();
+			TokenExchangeFailure::delete_failure_note();
 
 			// Update completed successfully.
 			Pinterest_For_Woocommerce()::set_api_version( 'v5' );
--- a/src/API/Auth.php
+++ b/src/API/Auth.php
@@ -9,7 +9,6 @@
 namespace Automattic\WooCommerce\Pinterest\API;
 
 use Automattic\WooCommerce\Pinterest\Logger;
-use Automattic\WooCommerce\Pinterest\Notes\TokenExchangeFailure;
 use Throwable;
 use WP_HTTP_Response;
 use WP_REST_Request;
--- a/src/API/Auth.php
+++ b/src/API/Auth.php
@@ -8,7 +8,8 @@
 
 namespace Automattic\WooCommerce\Pinterest\API;
 
-use Automattic\WooCommerce\Pinterest\Logger as Logger;
+use Automattic\WooCommerce\Pinterest\Logger;
+use Automattic\WooCommerce\Pinterest\Notes\TokenExchangeFailure;
 use Throwable;
 use WP_HTTP_Response;
 use WP_REST_Request;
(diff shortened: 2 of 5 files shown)


src/Notes/TokenExchangeFailure.php
```diff
@@ -74,4 +75,21 @@ public static function possibly_add_note() {
...
+
+	/**
+	 * Set note to actioned.
+	 *
+	 * @since x.x.x
+	 *
+	 * @return void
+	 */
+	public static function possibly_action_note() {
+		if ( ! self::note_exists() ) {
```
message-dimke: I had to do few tests: success case when sooner or later retry will work out, the case when I click on a banner's action button and the case if all retry actions fail. 

I could not see the note after the first test case because note is in the database but actioned. What if we change note check to check for a note and a status instead. This way if a user will install and rollback multiple times the banner will show again.

```suggestion
		if ( ! self::note_exists() && ! self::has_note_been_actioned() ) {
```

ecgan (commented):
I did some testing, and I have a finding that I think pretty similar to @message-dimke's comment in https://github.com/woocommerce/pinterest-for-woocommerce/pull/925#discussion_r1521109753.

My testing is done with these steps:

> - Connect v3 version and authenticate.
> - Update with this version but use broken version ( tamper with the oauth url ).
> - After the first failure happens an action for retry will be scheduled

When the first failure happens, I see the following: 

Screenshot 2024-03-12 at 6 38 49 PM

I clicked on the "Authenticate with Pinterest" button, and proceeded to successfully connect my store with Pinterest. This is done before the retry scheduled date time. 

However, when the retry kicks in, the next retry is still scheduled, even after the store is connected:

Screenshot 2024-03-12 at 6 44 37 PM

💡 I think during the action run, we need to check whether the store is already connected with v5 token. If yes, we can stop the action and no need to schedule another retry.

src/Notes/TokenExchangeFailure.php
budzanowski: I really don't want to support rolling back multiple times. You never know what can happen.

In `possibly_action_note` we want to hide note in case the procedure was successful. With this change the note will never be hidden unless somebody will click the button on the notice.

message-dimke: Successful scheduled action execution will also hide it. But ok, maybe my case is synthetic, and we do not need this.

ecgan (commented):
> My testing is done with these steps:
> 
> > * Connect v3 version and authenticate.
> > * Update with this version but use broken version ( tamper with the oauth url ).
> > * After the first failure happens an action for retry will be scheduled
> 
> When the first failure happens, I see the following: [screenshot with "action required" banner]

I did another round of testing that is similar to the above, but this time I do not see the "action required" banner after updating the plugin.

I looked into the code and I figured this is because previously I have already actioned on the note. I can see this in the database table `wp_wc_admin_notes` with name `pinterest-for-woocommerce-v5-token-exchange-failure` and status `actioned`.

💡 I think we need to reset or delete the note when we are disconnecting Pinterest connection or when we are uninstalling the plugin (using uninstall.php file).

budzanowski:
@ecgan 
> I think we need to reset or delete the note when we are disconnecting Pinterest connection or when we are uninstalling the plugin (using uninstall.php file).

We don't because disconnecting or deleting ( installing later ) always puts merchants on the path of regular manual connection requirement which is not an error scenario.

budzanowski:
@ecgan 
> 💡 I think during the action run, we need to check whether the store is already connected with v5 token. If yes, we can stop the action and no need to schedule another retry.

Correct! This should now be handled by [ee4038f](https://github.com/woocommerce/pinterest-for-woocommerce/pull/925/commits/ee4038fc5117f8252e4123093a11da14b2330487)

Please retry.

ecgan:
@budzanowski ,

> > I think we need to reset or delete the note when we are disconnecting Pinterest connection or when we are uninstalling the plugin (using uninstall.php file).
> 
> We don't because disconnecting or deleting ( installing later ) always puts merchants on the path of regular manual connection requirement which is not an error scenario.

But what if users did what I did, e.g. disconnecting Pinterest connection, uninstall Pinterest plugin (with v5 API), install old Pinterest plugin (with v3 API), and then update Pinterest plugin (with v5 API), and then for some reasons there is an error scenario there, but the note is not shown because it is already actioned in the database?

Also, according to https://developer.wordpress.org/plugins/plugin-basics/uninstall-methods/:

> When your plugin is uninstalled, you’ll want to clear out any plugin options and/or settings specific to the plugin, and/or other database entities such as tables.

I suppose this is also applicable to WC Admin notes too?

budzanowski:
@ecgan 
> But what if users did what I did, e.g. disconnecting Pinterest connection, uninstall Pinterest plugin (with v5 API), install old Pinterest plugin (with v3 API), and then update Pinterest plugin (with v5 API), and then for some reasons there is an error scenario there, but the note is not shown because it is already actioned in the database?

That is a very obscure scenario. The moment you disconnect, uninstall and go to install old version just to install new version sounds sketchy. We can reinforce the plugin for  this scenario but it starts too feel like an overkill.

budzanowski:
@ecgan please check [2cd9dcf](https://github.com/woocommerce/pinterest-for-woocommerce/pull/925/commits/2cd9dcffb5c53bdac55f19c353182fb812ee7b2f) it deletes the note on success and uninstall making the code ready to generate a new note in case it will be needed.

class-pinterest-for-woocommerce.php
```diff
@@ -615,6 +617,17 @@ public static function set_api_version( $version ) {
...
 
+		/**
+		 * Get API version used by the plugin.
+		 *
+		 * @since x.x.x
+		 *
+		 * @return string The API version.
+		 */
+		public static function get_api_version() {
+			return get_option( PINTEREST_FOR_WOOCOMMERCE_PINTEREST_API_VERSION, '' );
```
ecgan, line 628: With reference to https://developer.wordpress.org/plugins/plugin-basics/uninstall-methods/, I'm thinking we should delete the option (via https://github.com/woocommerce/pinterest-for-woocommerce/blob/30897feb5ada1962a61f6e1c9f98021e50381580/uninstall.php) when we are uninstalling the plugin. We can do this in a separate PR.

ecgan (approved):
Tested based on testing instructions, things seem to work as expected. 👍 

I'm leaving one minor comment below which can be done in a separate PR.

PHP code change looks good to me, though I'd defer that to @message-dimke for his opinion and approval.

ecgan:
> @ecgan please check https://github.com/woocommerce/pinterest-for-woocommerce/pull/925/commits/2cd9dcffb5c53bdac55f19c353182fb812ee7b2f it deletes the note on success and uninstall making the code ready to generate a new note in case it will be needed.

I just noticed this message. I'll test it out.

class-pinterest-for-woocommerce.php
budzanowski: https://github.com/woocommerce/pinterest-for-woocommerce/issues/931

ecgan:
> > @ecgan please check [2cd9dcf](https://github.com/woocommerce/pinterest-for-woocommerce/commit/2cd9dcffb5c53bdac55f19c353182fb812ee7b2f) it deletes the note on success and uninstall making the code ready to generate a new note in case it will be needed.
> 
> I just noticed this message. I'll test it out.

Something's weird. I tried out the latest code here with a "broken" version, and I don't see the retry scheduled action, no "action required" banner, the token exchange did not happened, and I need to manually connect Pinterest account in the plugin settings page. 

I was expecting to see retry in scheduled action, and "action required" banner.

I'll try this out again tomorrow.

ecgan:
> Something's weird. I tried out the latest code here with a "broken" version, and I don't see the retry scheduled action, no "action required" banner, the token exchange did not happened, and I need to manually connect Pinterest account in the plugin settings page.
> 
> I was expecting to see retry in scheduled action, and "action required" banner.

The reason why it was not working like how I expected is because the `pinterest_for_woocommerce_pinterest_api_version` option is still in the database with the value of `'v5'` even after uninstallation.

Screenshot 2024-03-13 at 3 51 18 PM

When I was doing the test with the "broken" version, the code sees the `'v5'` option value, and decides not to do the retry. See the relevant code here: https://github.com/woocommerce/pinterest-for-woocommerce/pull/925/files#diff-9087ae5f2a84d37a915583363d0231c72410aa9b1ac515734de92a14b191e8e4R331-R334. That's why there was no retry scheduled action.

I manually removed the options from the database and tested it again, it works as expected now, I can see retry in scheduled action, and the "action required" banner. 

Note that removing the option upon uninstallation is being addressed in PR https://github.com/woocommerce/pinterest-for-woocommerce/pull/932.

src/API/Auth.php
```diff
@@ -8,7 +8,8 @@
 
 namespace Automattic\WooCommerce\Pinterest\API;
 
-use Automattic\WooCommerce\Pinterest\Logger as Logger;
+use Automattic\WooCommerce\Pinterest\Logger;
+use Automattic\WooCommerce\Pinterest\Notes\TokenExchangeFailure;
```
ecgan: `TokenExchangeFailure` is not being used here, so we should remove this.

```suggestion
```

ecgan (approved):
> @ecgan please check https://github.com/woocommerce/pinterest-for-woocommerce/commit/2cd9dcffb5c53bdac55f19c353182fb812ee7b2f it deletes the note on success and uninstall making the code ready to generate a new note in case it will be needed.

Tested the above, works as expected. 👍 

Leaving one minor comment that I just noticed below.

PR #934, opened 2024-03-13
Delete feed on disconnect/uninstall.

message-dimke:

Delete feed instead of deactivating it on disconnect/uninstall action.

### Changelog entry

> Update - Delete feed instead of deactivate.

### DIFF
--- a/src/Feeds.php
+++ b/src/Feeds.php
@@ -321,48 +321,6 @@ class Feeds {
 	}
 
 	/**
-	 * Enabled the feed.
-	 *
-	 * @since x.x.x
-	 *
-	 * @param string $feed_id The ID of the feed.
-	 *
-	 * @return bool True if the feed is has been enabled, false otherwise.
-	 */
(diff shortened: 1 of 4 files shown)


src/Feeds.php
```diff
@@ -354,7 +354,27 @@ public static function enabled_feed( string $feed_id ): bool {
 	public static function disable_feed( string $feed_id ): bool {
```
budzanowski: It looks like we should able to remove this method now since we are not using diable_feed.

budzanowski (changes requested):
Nice. I like that you have fixed the parameters order!

budzanowski:
Looks good. We can merge. Please just remove the disable_feed in case we don't need it.

src/Feeds.php
message-dimke: Removing enable/disable feed methods according to logic changes (replacing with delete feed method).

PR #939, opened 2024-03-16
Use options and not transients for feed status manipulation.

budzanowski:
Use options and not transients for feed status manipulation.

This is done to prevent issues with persisten object cache.

### DIFF
--- a/src/PluginUpdate.php
+++ b/src/PluginUpdate.php
@@ -130,7 +130,7 @@ class PluginUpdate {
 			),
 			'1.4.0'  => array(
 				'token_update',
-				'feed_status_migration_1_4_0',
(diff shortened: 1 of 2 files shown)


message-dimke (approved):
Looks good! I did a test by faking a lower extension version in the database to run the update procedure to 1.4.0 and check if transients were successfully migrated.

_MySQL_5_5_5-10_4_32-MariaDB-1_10_4_32_maria_ubu2004-log__wordpress-pinterest_db_wp_options

All the corresponding transients were successfully migrated, and the extension still functioned properly. 

After running the feed generator action, properties were updated, but transients were not created, which is the expected new behavior.

Thank you, @budzanowski !

PR #941, opened 2024-03-18
Remove dead code in FeedState.php

budzanowski:
Remove not reachable code. This block is covered by the if block starting at line 242.


ecgan (approved):
LGTM 👍

PR #957, opened 2024-03-19
Fix feed_creation cache key calculation.

### LINKED ISSUE #955: Feed creation delay on error is not working.
The Feeds::create feed transient used for preventing too soon feed creation when previous error was detected will not work because inside the $data used to calculate md5 transient cache we use time() method which makes caching not very useful.

### LINKED ISSUE #956: Feed creation transient cache key should use ad_account_id.
When we create cache key that stops feed creation attempt from being executed too often we should also use ad_account_id to generate the cache key. Without it we still are blocking feed creation even if the ad_account_id was changed by the merchant.

The code is located in Feeds::create_feed.

budzanowski:
Add ad_accound_id to the cache key calculation and remove time() from the cache key calculation.

Fixes: #956 
Fixes: #955

Testing requires debugging the methods.

### DIFF
--- a/src/Feeds.php
+++ b/src/Feeds.php
@@ -121,21 +121,17 @@ class Feeds {
 		);
 
 		$data = array(
-			'name'                          => $feed_name,
-			'format'                        => 'XML',
-			'location'                      => $config['feed_url'],
-			'catalog_type'                  => 'RETAIL',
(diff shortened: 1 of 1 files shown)


message-dimke (approved):
LGTM

PR #959, opened 2024-03-20
Always add processing status to results.

### LINKED ISSUE #958: Feed processing status should be added in all cases and not only for sucess.
If the feed ingesting status is different than `success` we are not adding the feed process which results in only default values being present in the overview.

We should add it in all cases. Later we should also add a UI warning that the presented numbers may not be correct if the status is different than `success` - API docs mention that.

budzanowski:
Fixes: #958 

Always add feed processing status.


ecgan (approved):
Code looks good to me. 

Before this PR, this is what I see in the Catalog page:

Screenshot 2024-03-20 at 7 27 33 PM

In this PR, this is what I see - the stats are still 0, and there is now a "Remote sync status":

Screenshot 2024-03-20 at 7 35 12 PM

The stats are 0 because Pinterest API does not return the numbers, and its status is "PROCESSING" (I checked this in the log file).

PR #961, opened 2024-03-20
Decouple feed processing from setup status.

budzanowski:
Decouple feed processing status from feed setup status.

With this new approach we see that Feed is correctly configured also when the Feed ingestion status is `Failed` this is done because `Failed` mens that processing has issues and not configuration.

Other change is that without the configure feed we don't add processing workflow - because it does not have sense to fetch issues when we don't have feed id.

Other changes that has been implemented is that feed processing status DISAPPROVED is now 'error' and not 'pending' 

Processing stat has now a notice that overview numbers may be inaccurate until processing finishes.

### Testing

Best way to test this is to setup the plugin from scratch and observe different stages of feed and setup and see if they make sense.

### DIFF
--- a/src/API/FeedState.php
+++ b/src/API/FeedState.php
@@ -463,7 +463,7 @@ class FeedState extends VendorAPI {
 			case 'PROCESSING':
 				return sprintf(
 					/* Translators: %1$s Time difference string, %2$s number of products */
-					esc_html__( 'overview numbers can be inaccurate untill processing finishes.', 'pinterest-for-woocommerce' ),
+					esc_html__( 'Overview numbers may be inaccurate until processing finishes.', 'pinterest-for-woocommerce' ),
 				);
 			case 'DISAPPROVED':
 			case 'QUEUED_FOR_PROCESSING':
--- a/src/API/FeedState.php
+++ b/src/API/FeedState.php
@@ -235,11 +235,6 @@ class FeedState extends VendorAPI {
 			);
 		}
 
-		if ( FeedStatusService::FEED_STATUS_ERROR_FETCHING_FEED === $status ) {
-			$status       = 'error';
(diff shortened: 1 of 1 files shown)


src/API/FeedState.php
```diff
@@ -465,21 +446,25 @@ private static function map_status_into_extra_info( array $processing_results ):
...
-					esc_html__( 'Last attempt: %1$s ago', 'pinterest-for-woocommerce' ),
+					esc_html__( '%1$s ago', 'pinterest-for-woocommerce' ),
 					human_time_diff( $processing_date->getTimestamp() )
 				);
 				$global_error = Pinterest\FeedStatusService::get_processing_results_global_error( $processing_results );
 				return $info . ( $global_error ? ' - ' . $global_error : '' );
+			case 'PROCESSING':
+				return sprintf(
+					/* Translators: %1$s Time difference string, %2$s number of products */
+					esc_html__( 'overview numbers can be inaccurate untill processing finishes.', 'pinterest-for-woocommerce' ),
```
ecgan: Edit for the notice sentence:

```suggestion
					esc_html__( 'Overview numbers may be inaccurate until processing finishes.', 'pinterest-for-woocommerce' ),
```

ecgan (commented):
Tested based on the test instructions (I disconnect Pinterest account, uninstall plugin, reinstall plugin and connect account), it seems working as expected so far and makes sense for me. 

Here are some screenshots on the different stages:

Screenshot 2024-03-21 at 12 23 51 AM

Screenshot 2024-03-21 at 12 24 51 AM

Screenshot 2024-03-21 at 12 25 08 AM

I have not fully understood the change in the code logic yet. I can look into it tomorrow. Alternatively, @message-dimke may be a better person to look into the logic changes.

I'm leaving one comment on the new notice sentence below.

ecgan:
Adding a screenshot for when the processing is complete:

image

PR #963, opened 2024-03-21
Fix small issues found while looking at the file.

budzanowski:
Please review the changes. It looks like testing is not required this is only style and spelling.

### DIFF
--- a/class-pinterest-for-woocommerce.php
+++ b/class-pinterest-for-woocommerce.php
@@ -464,7 +464,7 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 		}
 
 		/**
-		 * Load Localization files.
+		 * Load localization files.
 		 *
 		 * Note: the first-loaded translation file overrides any following ones if the same translation is present.
 		 *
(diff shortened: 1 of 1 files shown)


class-pinterest-for-woocommerce.php
```diff
@@ -463,7 +464,7 @@ public function maybe_update_plugin() {
 		}
 
 		/**
-		 * Load Localisation files.
+		 * Load Localization files.
```
ecgan: Should it be uppercase or lowercase?

```suggestion
		 * Load localization files.
```

class-pinterest-for-woocommerce.php
```diff
@@ -343,7 +345,7 @@ public static function init_tracking() {
 		}
 
 		/**
-		 * Init Pinterest_For_Woocommerce when WordPress Initialises.
+		 * Init Pinterest_For_Woocommerce when WordPress Initializes.
```
ecgan: Lowercase "i"?

```suggestion
		 * Init Pinterest_For_Woocommerce when WordPress initializes.
```

ecgan (approved):
The changes look okay. I did not test them. Leaving two minor comments about casing below.

PR #972, opened 2024-03-26
Fix refresh token response log message.

message-dimke:

Before the change, the output was just the `Array`. Now, we can see some details, with security measures applied to prevent sensitive data from appearing in logs.

### Changelog entry

> Tweak - Update refresh token request's response message.


ibndawood (approved):
Thank you, @message-dimke.

PR #975, opened 2024-03-27
Add Twenty Twenty-Four compatibility.

### LINKED ISSUE #973: Twenty Twenty-Four Save Pin not working
The add to Pin feature is not working correctly with the Twenty Twenty-Four theme.

To reproduce open shop page and hover over on of the products.

In Storefront the button is clearly visible:
image

In Twenty Twenty-Four the button is being shown not over the product button the top left corner of page, barely visible ( for each product ):
image

budzanowski:

Closes #973  .

The absolute positioning of the element does not work with Twenty Twenty-Four because the immediate parent is not a positioned element in this theme.

> The element is positioned relative to its closest positioned ancestor (if any) or to the initial [containing block](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#identifying_the_containing_block). Its final position is determined by the values of top, right, bottom, and left. 

From https://developer.mozilla.org/en-US/docs/Web/CSS/position#absolute

For this reason we need to detect the new parent `.wp-block-post.product` and unset position values. We are sill leaving the `absolute` label to keep the button stacked over the image on hover. 

### Screenshots:

image

1. Switch to Twenty Twenty-Four
2. Hover over a produc
3. Pin button should be visible an clickable
4. Repeat with Storefront to confirm that there is no regression

### DIFF
--- a/assets/source/sass/frontend/pinterest-for-woocommerce-pins.scss
+++ b/assets/source/sass/frontend/pinterest-for-woocommerce-pins.scss
@@ -25,7 +25,9 @@
 
 // Twenty Twenty-Four compatibility.
 .wp-block-post {
+
 	&.product {
+
 		.pinterest-for-woocommerce-image-wrapper {
 			top: unset;
 			left: unset;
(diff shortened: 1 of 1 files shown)


rawdreeg (approved):
The pins looks good and work well both with Twenty Twenty-Four and Storefront.

ibndawood (approved):
Thank you, @budzanowski. I tested the PR on storefront and twentytwentyfour. The pins work as expected. 

Just a note about the cache: we need to clear it before viewing the changes.

PR #977, opened 2024-03-27
Revert version before the release.

budzanowski:
Getting ready for release. Revert version so the release script can update all of the necessary versions automatically.


ibndawood (approved):

ecgan:
Adding a note that in this PR, only the version number in `pinterest-for-woocommerce.php` was reverted; the version number in `package.json` was not reverted. This was causing a conflict as mentioned in https://github.com/woocommerce/pinterest-for-woocommerce/pull/980#issuecomment-2029877150.

PR #987, opened 2024-04-09
Fix typo.

budzanowski:
Wrong value was deleted from the DB.


message-dimke (approved):
LGTM! Thank you, @budzanowski !

PR #988, opened 2024-04-09
Removing unused `feature_flags` key from connection info data.

message-dimke:

Removing unused connection info data `feature_flags` key. 
Does not affect anything since tracking switch is stored at another options key, and CAPI feature flag is not used anywhere.

### Changelog entry

> Tweak - Remove `feature_flag` connection info data key.

### DIFF
--- a/src/API/TokenExchangeV3ToV5.php
+++ b/src/API/TokenExchangeV3ToV5.php
@@ -102,11 +102,6 @@ class TokenExchangeV3ToV5 extends APIV5 {
 			'advertiser_id' => Pinterest_For_Woocommerce()::get_setting( 'tracking_advertiser' ),
 			'tag_id'        => Pinterest_For_WooCommerce()::get_setting( 'tracking_tag' ),
 			'merchant_id'   => Pinterest_For_Woocommerce()::get_data( 'merchant_id' ),
-			'feature_flags' => array(
-				'tags'    => true,
-				'CAPI'    => true,
-				'catalog' => true,
-			),
 		);
 		Pinterest_For_Woocommerce()::save_connection_info_data( $info_data );
 


budzanowski:
Is this good to go?

message-dimke:
Hey, @budzanowski , it is!

src/API/TokenExchangeV3ToV5.php
```diff
@@ -102,11 +102,6 @@ public static function token_update() {
 			'advertiser_id' => Pinterest_For_Woocommerce()::get_setting( 'tracking_advertiser' ),
 			'tag_id'        => Pinterest_For_WooCommerce()::get_setting( 'tracking_tag' ),
 			'merchant_id'   => Pinterest_For_Woocommerce()::get_data( 'merchant_id' ),
-			'feature_flags' => array(
```
budzanowski, line 105: Just a question. We use the featgure_flats to set settings via `apply_oauth_flow_features` method. Are those settings preserved correctly from the V3 version in case of migration?

budzanowski (approved):
I have left a question. It the answer is that everything is preserved then please proceed with the merge. Pre-approving.

src/API/TokenExchangeV3ToV5.php
message-dimke: They are. Once connected, features are saved inside Pinterest settings under different keys.

```php
Pinterest_For_Woocommerce()::save_setting( 'track_conversions', $features['tags'] ?? false );
Pinterest_For_Woocommerce()::save_setting( 'track_conversions_capi', false );
Pinterest_For_Woocommerce()::save_setting( 'product_sync_enabled', $features['catalog'] ?? false );
```

message-dimke: When doing a token exchange, there is a settings update procedure. It looks for settings and if those do not exist, updates the settings with defaults. We are also safe here, since the settings are in the database. 

https://github.com/woocommerce/pinterest-for-woocommerce/blob/c85534a9f5d6f4f76fe673de2ffef84ba98999ab/class-pinterest-for-woocommerce.php#L1283-L1289

PR #997, opened 2024-04-11
Fix/fetch correct coupons information

### LINKED ISSUE #996: Existing discounts are not recognised.
There is something wrong with the existing discounts processing. They are not recognized.

This is super hard to test. One needs to have a valid connection, a correct billing setup and a discount applied to the account. Next the procedure to claim the coupon should be triggered again. Because already claimed coupons are not recognised the code will try to redeeme to coupon again - which should not happen.

budzanowski:

Closes #996 .

Get proper data for discounts processing.

1. Set a breakpoint in `process_available_discounts`
2. trigger hourly action.
3. step through the code to verify that correct array is fetched for discounts processing.

### DIFF
--- a/src/AdCredits.php
+++ b/src/AdCredits.php
@@ -246,14 +246,13 @@ class AdCredits {
 			return false;
 		}
 
-		$advertiser_id = Pinterest_For_Woocommerce()::get_setting( 'tracking_advertiser' );
+		$advertiser_id      = Pinterest_For_Woocommerce()::get_setting( 'tracking_advertiser' );
+		$discounts          = APIV5::get_ads_credit_discounts( $advertiser_id )['items'] ?? array();
+		$coupon             = AdCreditsCoupons::get_coupon_for_merchant();
+		$remaining_discount = 0;
+		$found_discounts    = array();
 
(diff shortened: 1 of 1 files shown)


src/AdCredits.php
```diff
@@ -246,14 +246,13 @@ public static function process_available_discounts() {
 			return false;
 		}
 
-		$advertiser_id = Pinterest_For_Woocommerce()::get_setting( 'tracking_advertiser' );
+		$advertiser_id      = Pinterest_For_Woocommerce()::get_setting( 'tracking_advertiser' );
+		$discounts          = APIV5::get_ads_credit_discounts( $advertiser_id )['items'] ?? array();
```
rawdreeg, line 252: APIV5::get_ads_credit_discounts might throw a `PinterestApiException`.  can this handled?

rawdreeg (approved):
Testing well. I left a comment on handling PinterestApiException. But I understand if we want the exception to thrown.

src/AdCredits.php
budzanowski: Good catch. It is handled downstream.

I have added `throws` to the dockblock for better visibility.

PR #998, opened 2024-04-12
Removing mandatory condition on tracking.

message-dimke:

The New Pinterest Connection Wizard has Tag as an optional step that can be skipped. Compared to the old plugin version (v3 API), where Tag was a mandatory condition to pass the Connection Wizard, the current implementation does not require it to function. However, some application bits check for Tag to be present to sync the catalog and slow landing page. 

Removing Tag presence as a mandatory condition to pass the landing page and enable product sync. 

The symptoms of this are after successfully connecting to Pinterest (with the Tag step skipped), you are faced with a landing page instead of the connection page and the Catalog tab saying, "Product sync is disabled." 

#### Landing Page

Pinterest_‹_WordPress_Pinterest_—_WooCommerce

After the "Get started" button is clicked, you actually see the connected user account information.

Onboarding_Guide_‹_Pinterest_‹_Marketing_‹_WooCommerce_‹_WordPress_Pinterest_—_WooCommerce

#### Product Sync

Products_Catalog_‹_Pinterest_‹_Marketing_‹_WooCommerce_‹_WordPress_Pinterest_—_WooCommerce

When the Catalog and the Feed are actually present.

Pinterest

### Changelog entry

> Fix - Removing Tag ID mandatory condition.

### DIFF
--- a/src/ProductSync.php
+++ b/src/ProductSync.php
@@ -11,9 +11,7 @@ namespace Automattic\WooCommerce\Pinterest;
 if ( ! defined( 'ABSPATH' ) ) {
 	exit;
 }
-use \Automattic\WooCommerce\ActionSchedulerJobFramework\Proxies\ActionScheduler as ActionSchedulerProxy;
-use Automattic\WooCommerce\Pinterest\FeedRegistration;
-use Automattic\WooCommerce\Pinterest\API\FeedIssues;
+use Automattic\WooCommerce\ActionSchedulerJobFramework\Proxies\ActionScheduler as ActionSchedulerProxy;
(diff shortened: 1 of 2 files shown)


src/ProductSync.php
```diff
@@ -136,11 +134,9 @@ private static function initialize_feed_components() {
 	 * @return boolean
 	 */
 	public static function is_product_sync_enabled() {
-
 		$domain_verified  = Pinterest_For_Woocommerce()::is_domain_verified();
```
budzanowski:
```suggestion
		$domain_verified = Pinterest_For_Woocommerce()::is_domain_verified();
```

budzanowski (approved):
Looks good. Left one cleanup suggestion. :shipit:

PR #1001, opened 2024-04-15
Adding missing Order ID into `custom_data` array for Checkout CAPI event.

message-dimke:

Pinterest reported Order ID is missing from Checkout CAPI events.
Adding Order ID into the `custom_data` event data array.

### Changelog entry

> Fix - Add Order ID into Checkout CAPI event.

### DIFF
--- a/src/Tracking/Conversions.php
+++ b/src/Tracking/Conversions.php
@@ -50,8 +50,6 @@ class Conversions extends Tracker {
 		$data = $this->prepare_request_data( $event_name, $data );
 
 		try {
-			$this->send_request( $event_name, $data );
-
 			/* translators: 1: Conversions API event name, 2: JSON encoded event data. */
 			$messages = sprintf(
 				'Sending Pinterest Conversions API event %1$s with a payload: %2$s',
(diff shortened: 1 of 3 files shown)


src/Tracking/Conversions.php
```diff
@@ -50,15 +50,15 @@ public function track_event( string $event_name, Data $data ) {
...
-
 			/* translators: 1: Conversions API event name, 2: JSON encoded event data. */
 			$messages = sprintf(
 				'Sending Pinterest Conversions API event %1$s with a payload: %2$s',
 				$event_name,
 				wp_json_encode( $data )
 			);
 			Logger::log( $messages, 'debug', 'conversions' );
+
+			$this->send_request( $event_name, $data );
```
budzanowski: Why do we have this send request moved to after the `logging`?

message-dimke: I want logs to show what the code is attempting to do before it may fall into an exception, either a connection exception or a Pinterest API error in response, so I moved the logging to the top before the call.

budzanowski: I would argue that this is unnecessary. We should always have one log entry for CAPI event.

In case of success a success message, in case of failure the failure message.

With the approach that you are proposing, we get 2 logs for the failure scenario and 1 log for the success scenario.
This also means that you get failure confirmation and don't get success confirmation.

message-dimke: Ok then. I did put the logging back after the API call.

budzanowski (approved):
Looks good!

PR #1004, opened 2024-04-15
Improve cohesion for billing methods by moving the to Billing class.

budzanowski:
This is a simple refactor/tidy PR. No functionality should be affected. Only improved code readability.

This is done as a step towards #994


ibndawood (approved):
Thank you, @budzanowski, for the changes. I searched for the usage of `update_billing_information` and `add_billing_setup_status_to_account_data` methods. I confirm that all the occurrences are updated to use `Billing`.  

I also verified that the `Billing` class is imported wherever these new methods are called.

I added a couple of minor `phpcs` fixes.

budzanowski:
@ibndawood thank you!

PR #1006, opened 2024-04-17
Add billing status info in the Settings UI

### LINKED ISSUE #994: Add billing status information to settings page.

budzanowski:
This will be useful to merchants but I have quickly developed it to know if the advertiser_id I am using for testing has correct billing setup.

### Screenshots:

Invalid billing status:
image

Valid billing status
image

1. Go to wp-admin -> Marketing -> Pinterest -> Connection
2. Observe billing status with Advertiser that has correct billing and with advertiser that has no billing status set 
3. Make sure that the button takes you to the billing settings.

### Changelog entry

> Add - Billing status UI in Connection tab.

### DIFF
--- a/assets/source/setup-guide/app/components/Billing/Billing.js
+++ b/assets/source/setup-guide/app/components/Billing/Billing.js
@@ -1,0 +1,136 @@
+/**
+ * External dependencies
+ */
+import { useState } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+import apiFetch from '@wordpress/api-fetch';
+import { Spinner } from '@woocommerce/components';
+import { getNewPath } from '@woocommerce/navigation';
+import { recordEvent } from '@woocommerce/tracks';
+import {
+	Button,
+	CardBody,
+	Flex,
+	FlexItem,
+	FlexBlock,
+	__experimentalText as Text, // eslint-disable-line @wordpress/no-unsafe-wp-apis --- _experimentalText unlikely to change/disappear and also used by WC Core
+} from '@wordpress/components';
+
+
+/**
+ * Clicking on "Connect" Pinterest account button.
+ *
+ * @event wcadmin_pfw_account_connect_button_click
+ */
+/**
+ * Clicking on "Disconnect" Pinterest account button.
+ *
+ * @event wcadmin_pfw_account_disconnect_button_click
+ * @property {string} context `'settings' | 'wizard'` In which context it was used?
+ */
+/**
+ * Opening a modal.
+ *
+ * @event wcadmin_pfw_modal_open
+ * @property {string} name Which modal is it?
+ * @property {string} context `'settings' | 'wizard'` In which context it was used?
+ */
+/**
+ * Closing a modal.
+ *
+ * @event wcadmin_pfw_modal_closed
+ * @property {string} name Which modal is it?
+ * @property {string} context `'settings' | 'wizard'` In which context it was used?
+ * @property {string} action
+ * 				`confirm` - When the final "Yes, I'm sure" button is clicked.
+ * 				`dismiss` -  When the modal is dismissed by clicking on "x", "cancel", overlay, or by pressing a keystroke.
+ */
+
+/**
+ * Pinterest account connection component.
+ *
+ * This renders the body of `SetupAccount` card, to connect or disconnect Pinterest account.
+ *
+ * @fires wcadmin_pfw_account_connect_button_click
+ * @fires wcadmin_pfw_account_disconnect_button_click with the given `{ context }`
+ * @fires wcadmin_pfw_modal_open with `{ name: 'account-disconnection', … }`
+ * @fires wcadmin_pfw_modal_closed with `{ name: 'account-disconnection', … }`
+ * @param {Object} props React props.
+ * @param {boolean} props.isConnected
+ * @param {Function} props.setIsConnected
+ * @param {Object} props.accountData
+ * @param {string} props.context Context in which the component is used, to be forwarded to fired Track Events.
+ * @return {JSX.Element} Rendered element.
+ */
+const Billing = ( {
+	accountData,
+} ) => {
+
+	const isBillingSetup = accountData?.is_billing_setup;
+
+	return (
+		<CardBody size="large">
+			<Flex direction="row" className="billing-info">
+				{ isBillingSetup === true ? ( // eslint-disable-line no-nested-ternary --- Code is reasonable readable
+					<>
+						<FlexBlock className="billing-label">
+							<Text variant="body">
+								{ __(
+									'Billing Setup Correctly',
+									'pinterest-for-woocommerce'
+								) }
+							</Text>
+						</FlexBlock>
+
+						<FlexItem>
+							<Button
+								href={
+									wcSettings.pinterest_for_woocommerce
+										.billingSettingsUrl
+								}
+								onClick={ () =>
+									recordEvent(
+										'pfw_go_to_billing_button_click'
+									)
+								}
+							>
+								{ __( 'Go to billing settings', 'pinterest-for-woocommerce' ) }
+							</Button>
+						</FlexItem>
+					</>
+				) : isBillingSetup === false ? (
+					<>
+						<FlexBlock>
+							<Text variant="body">
+								{ __(
+									'No Valid Billing Setup Found',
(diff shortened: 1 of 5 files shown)


includes/admin/class-pinterest-for-woocommerce-admin.php
```diff
@@ -367,6 +367,7 @@ private function get_component_settings() {
 				'serviceLoginUrl'          => $this->get_service_login_url(),
 				'createBusinessAccountUrl' => $this->get_create_business_account_url(),
 				'switchBusinessAccountUrl' => $this->get_switch_business_account_url(),
+				'billingSettingsUrl'       => $this->get_advertiser_billing_settings_url(),
```
ecgan, line 370: I think we don't need this. We already have access to the advertiser ID in the React JS code. Look for `appSettings.tracking_advertiser`.

assets/source/setup-guide/app/views/ConnectionApp.js
```diff
@@ -58,6 +59,11 @@ const SettingsApp = () => {
 						isConnected={ isConnected }
 						isBusinessConnected={ isBusinessConnected }
 					/>
+					<BillingStatus
+						setIsConnected={ setIsConnected }
+						isConnected={ isConnected }
+						isBusinessConnected={ isBusinessConnected }
```
ecgan: `BillingStatus` does not have these props.

```suggestion
```

assets/source/setup-guide/app/steps/BillingStatus.js
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Internal dependencies
+ */
+import Billing from '../components/Billing/Billing';
+import StepOverview from '../components/StepOverview';
+import { useSettingsSelect } from '../helpers/effects';
+
+const BillingStatus = () => {
+	const appSettings = useSettingsSelect();
```
ecgan: Since `appSettings` is not really used within this `BillingStatus` component but is being passed down into `Billing` child component, we can just move this line into `Billing`. So `Billing` does not need to have the `accountData` prop.

assets/source/setup-guide/app/components/Billing/Billing.js
```diff
@@ -0,0 +1,119 @@
...
+						{ __(
+							'Billing Setup Correctly',
+							'pinterest-for-woocommerce'
+						) }
+					</Text>
+				</FlexBlock>
+
+				<FlexItem>
+					<Button
+						variant="link"
```
ecgan: We are using older version of WordPress components, so this one isn't right. We should use `isLink` instead.

```suggestion
						isLink
```

assets/source/setup-guide/app/components/Billing/Billing.js
```diff
@@ -0,0 +1,119 @@
...
+						{ __(
+							'No Valid Billing Setup Found',
+							'pinterest-for-woocommerce'
+						) }
+					</Text>
+				</FlexBlock>
+
+				<FlexItem>
+					<Button
+						variant="primary"
```
ecgan: With this code, it would produce this UI:

Screenshot 2024-04-18 at 4 19 41 PM

I see two issues with this:

- The button text is black and it is not obvious that it is a button.
- It has a bit more padding on the right and it is not aligned with the "Disconnect" button above it.

We should use `isLink` without "Primary".

```suggestion
						isLink
```

It would produce the following UI:

image

assets/source/setup-guide/app/components/Billing/Billing.js
```diff
@@ -0,0 +1,119 @@
...
+	const isBillingSetup = accountData?.is_billing_setup;
+	const statusLabe = classnames( 'pfw-billing-info', {
+		'pfw-billing-info--status-success': isBillingSetup === true,
+		'pfw-billing-info--status-error': isBillingSetup === false,
+	} );
+
+	let element = null;
+
+	if ( isBillingSetup === undefined ) {
+		element = <Spinner className="billing-info__preloader" />;
```
ecgan: We don't have any class name with `billing-info__preloader`. 

I suppose you would want to add this into the CSS:

```css
	.billing-info__preloader {
		flex: 1;
	}
```

This would make the spinner show up as center aligned.

And you might want to change `billing-info` to `pfw-billing-info`?

assets/source/setup-guide/app/components/Billing/Billing.js
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * Clicking on "Go to billing settings" button.
+ *
+ * @event wcadmin_pfw_go_to_billing_button_click
+ */
+
+const Billing = ( { accountData } ) => {
+	const isBillingSetup = accountData?.is_billing_setup;
+	const statusLabe = classnames( 'pfw-billing-info', {
```
ecgan: Spelling mistake.

```suggestion
	const statusLabel = classnames( 'pfw-billing-info', {
```

Actually, we don't need this `statusLabel`. We can just put them inline into the `<FlexBlock className='...'>` JSX, because we already check for `isBillingSetup` in the rendering there.

assets/source/setup-guide/app/components/Billing/Billing.js
```diff
@@ -0,0 +1,119 @@
...
+		'pfw-billing-info--status-success': isBillingSetup === true,
+		'pfw-billing-info--status-error': isBillingSetup === false,
+	} );
+
+	let element = null;
```
ecgan: I'd prefer not to use `let element` and assign JSX to it. 

I'd prefer something like this:

```js
	const getElement = () => {
		if (isBillingSetup === undefined) {
			return "spinner here";
		}

		if (isBillingSetup === false) {
			return "no valid billing here";
		}

		// when isBillingSetup === true
		return "true condition";
	};

	// then when render:
	<Flex direction="row" className="pfw-billing-info">
		{getElement()}
	</Flex>;
```

With this, we don't need to keep track of what is `element`.

ecgan (changes requested):
Tested, looks okay and it works, but it can be better improved, in terms of code and user experience. Requesting changes.

assets/source/setup-guide/app/components/Billing/Billing.js
budzanowski: With isLink I see:
image

Do you maybe know why?

includes/admin/class-pinterest-for-woocommerce-admin.php
budzanowski: I will leave it as it is here for symmetry with other Urls.

budzanowski:
@ecgan than you for the review 👍🏻 

Everything has been adjusted.

assets/source/setup-guide/app/components/Billing/Billing.js
ecgan: I'm seeing the same thing. I think it is just TypeScript / code editor resolving to wrong type definition files, i.e. we are using and bundling old version of @wordpress/components, but it is resolved to newer version of definition files.

assets/source/setup-guide/app/components/Billing/Billing.js
```diff
@@ -0,0 +1,118 @@
...
+						{ __( 'Setup Billing', 'pinterest-for-woocommerce' ) }
+					</Button>
+				</FlexItem>
+			</>
+		);
+	};
+
+	return (
+		<CardBody size="large">
+			<Flex direction="row" className="pfw-billing-info">
```
ecgan, line 106: I just noticed that the class `pfw-billing-info` is in both `<Flex>` here and in `statusLabe`. I suppose it should be here in `<Flex>` and should not be in `statusLabe`?

assets/source/setup-guide/app/components/Billing/Billing.js
ecgan: It seems like the comment above was not addressed?

ecgan (changes requested):
Adding two comments related to `statusLabe`.

assets/source/setup-guide/app/components/Billing/Billing.js
budzanowski: It should be OK now.

ecgan (approved):
Tested, working as expected, LGTM 🙂 🚀 

Screenshot when billing setup is loading:

image


Screenshot of invalid billing status:

image


Screenshot of valid billing status:

image

PR #1015, opened 2024-05-01
Don't sync aems when tracking is not enabled.

### LINKED ISSUE #1014: Error in logs when tracking is disabled.
During onboarding, the merchant can opt out from tracking. If that happens, we can't sync settings for AEMS because it requires tracking id which is not present in the onboarding flow with disabled tracking.

The code was still trying to synchronise the settings but it was failing with and Error log. This should not happen. There should not be an attempt to sync AEMS settings.

budzanowski:

Closes #1014 .

Do not attempt AEMS setting sync when tracking is not enabled.

1. Connect with this PR, opt out from the tracking
2. wp-admin -> Marketing -> Pinterest -> Settings
3. For settings sync:
image

4. Go to logs and look for `Tracking advertiser or tag missing` it should not be there 😉

### DIFF
--- a/src/PinterestSyncSettings.php
+++ b/src/PinterestSyncSettings.php
@@ -95,6 +95,14 @@ class PinterestSyncSettings {
 	 * @throws Exception PHP Exception.
 	 */
 	private static function automatic_enhanced_match_support() {
+		/*
+		 * Tracking needs to be enabled in order to use automatic enhanced match support.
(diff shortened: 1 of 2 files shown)


ibndawood (approved):
Thank you, @budzanowski, for this PR. I followed the test instructions and I confirm the changes test well.

PR #1019, opened 2024-05-07
Revert woo.com → woocommerce.com in URLs

layoutd:

Revert `woo.com` → `woocommerce.com` in several places, especially for documentation links (pb0Spc-5mO-p2). Leaves `WooCommerce.com` camel-cased in visible text.

Also updates some README text and removes reference to unified nav in testing flows documentation.

Not sure if the extension author and "Made with 💜 by" should be Woo or WooCommerce.

1. No real code changes – just the recommendations API and icon image URLs.
2. Double check all new URLs actually go to a page.

### Changelog entry

> Tweak - Revert to WooCommerce.com domain.

### DIFF
--- a/.github/CODE_OF_CONDUCT.md
+++ b/.github/CODE_OF_CONDUCT.md
@@ -34,7 +34,7 @@ This Code of Conduct applies both within project spaces and in public spaces whe
 
 ## Enforcement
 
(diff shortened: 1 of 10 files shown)


rawdreeg (approved):
I have verified that we replaced all occurrences of woo.com. And that all links are correct.

PR #1026, opened 2024-06-21
Remove CAPI tracker by Pinterest request.

message-dimke:

Disabling Conversion API tracker according to some temporary legal aspects from Pinterest.

### How to test.

1. Install Pinterest extension.
2. Following the Connect Wizard, set up conversion tracking.

Pinterest

3. Setting up conversion tracking will send the extension corresponding boolean flags to indicate Pinterest Tag tracking and Conversions API tracking must be enabled.
4. Check `pinterest_for_woocommerce` option inside `wp_options` database table.
5. The option must include `s:22:"track_conversions_capi";b:0;` serialized data.
6. The above will indicate the despite Pinterest is sending us Conversions API is `ON`, we force it `OFF` into options.
e.g. `s:17:"track_conversions";b:1;s:22:"track_conversions_capi";b:0;`: Pinterest Tag on `ON`, but Conversions API is `OFF`.

### Changelog entry

> Update - Disabling CAPI tracker.

### DIFF
--- a/class-pinterest-for-woocommerce.php
+++ b/class-pinterest-for-woocommerce.php
@@ -335,11 +335,9 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 				return false;
 			}
 
-			$tag_tracker         = new Tag();
-			$user                = new User( WC_Geolocation::get_ip_address(), wc_get_user_agent() );
-			$conversions_tracker = new Conversions( $user );
+			$tag_tracker = new Tag();
 
-			return new Tracking( array( $tag_tracker, $conversions_tracker ) );
+			return new Tracking( array( $tag_tracker ) );
 		}
 
 		/**
--- a/class-pinterest-for-woocommerce.php
+++ b/class-pinterest-for-woocommerce.php
@@ -328,18 +328,15 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 			 *
 			 * @param bool $disable_tracking Whether to disable tracking.
 			 */
-			$is_tracking_disabled = apply_filters( 'woocommerce_pinterest_disable_tracking', false );
-			$is_not_a_site        = wp_doing_cron() || is_admin();
-			if ( $is_tracking_disabled || $is_not_a_site ) {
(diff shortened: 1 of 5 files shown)


budzanowski (changes requested):
We should do it via separate setting for CAPI. We should make the capi tracker option false. No UI for it would mean that we don't have to do anything else for now.

During the testing you just switch the CAPI tracker option to true and you can run the tests. We don't want to delete them because we will have to reintroduce them back very soon.

message-dimke:
> We should do it via a separate setting for CAPI. We should make the capi tracker option false. No UI for it would mean that we don't have to do anything else for now.

When connecting to Pinterest, Pinterest replies back with a set of features and their corresponding statuses. The set looks like this:
```
tags => bool
CAPI => bool
catalog => bool
```
 
The three flags above indicate what was chosen during the onboarding wizard (a Pinterest-hosted wizard). For example, choosing to sync catalog will send the `catalog` flag set to `true`.

The problem is tracking. Pinterest has a single wizard screen for tracking and a single yes/no button that turns ON/OFF both `tags` and `CAPI` flags synchronously. So, without forcing CAPI tracking to `false`, we can not disable that type of tracking.

Introducing the new `track_conversions_capi` setting will default to `false` for all the existing plugin users, but for the new ones, it will be set to the value selected while onboarding. Taking this into account, I will force the setting to false when the new user is connecting, no matter what the wizard returns to us.

Does it make sense?

budzanowski:
>  I will force the setting to false when the new user is connecting, no matter what the wizard returns to us.

For now this is good enough. We can change it later when they will have separate consents.

budzanowski:
Looks good. Do we need to add plugin update procedure that sets CAPI tracking to false explicitly? Or not set will default to false?

message-dimke:
If the setting is empty or there is no such setting, it will resolve to `false`.

https://github.com/woocommerce/pinterest-for-woocommerce/blob/c85534a9f5d6f4f76fe673de2ffef84ba98999ab/class-pinterest-for-woocommerce.php#L557-L562

If I read it correct =)
Anyway, I have unit tests that should test this as well.

budzanowski (approved):
OK lets do it. Please remember that this should be a GitHub only release.

PR #1027, opened 2024-06-22
Fix pagination on feed_issues endpoint

### LINKED ISSUE #936: Issues table pagination does not respect the page number and page size, and always return all issues
### Describe the bug:

Related to issue https://github.com/woocommerce/pinterest-for-woocommerce/issues/882.

In the Issues table, I have pagination in the table footer like this:

image

There are issues with the pagination:

- When I go to page 2, I noticed that the data is the same as page 1.
- The "Rows per page" is set to 25, but there are 40 issues in the table. 
- This means 40 issues in page 1, and the same 40 issues in page 2.

Here's a screenshot of the API call:

image

From the client side, we pass `paged` and `per_page` parameters to the API.

However, on the server side, those parameters are not really used:

https://github.com/woocommerce/pinterest-for-woocommerce/blob/8251408b26fc5838f76b76518e2942efea8efa84/src/API/FeedIssues.php#L39-L74

Specifically, in the following line, we set the limit to be 250, instead of the `per_page` value:

https://github.com/woocommerce/pinterest-for-woocommerce/blob/8251408b26fc5838f76b76518e2942efea8efa84/src/API/FeedIssues.php#L59

I believe the above is the cause of issue.

### Steps to reproduce:

1. Go to Issues table in the Catalog page.
2. Test pagination in the Issues table footer.

### Expected behavior:

Pagination should work as expected. 

### Actual behavior:

The table will always show all data, regardless of how you set pagination.

brezocordero:

Closes #936 

Fix the pagination on the Feed Issues table by slicing the results based on `per_page` and `paged` parameters.

1. Requires a site connected to Pinterest and a synched catalog with issues. 
2. Go to wp-admin/Marketing/Pinterest
3. If your catalog has more than 25 issues, check the pagination works
4. If your catalog has less than 25 issues, add the parameters to the query and check the number of issues displayed is correct. (Ex: on a catalog with 18 issues `page=wc-admin&path=%2Fpinterest%2Fcatalog&paged=2&per_page=10` should display 8 issues)

Screenshot 2024-06-21 at 7 17 16 PM

### Changelog entry

> Fix - Pagination on Feed Issues table.

### DIFF
--- a/assets/source/catalog-sync/sections/SyncIssues.js
+++ b/assets/source/catalog-sync/sections/SyncIssues.js
@@ -8,12 +8,12 @@ import {
 	ExternalLink,
 	__experimentalText as Text, // eslint-disable-line @wordpress/no-unsafe-wp-apis --- _experimentalText unlikely to change/disappear and also used by WC Core
 } from '@wordpress/components';
+import { __, sprintf } from '@wordpress/i18n';
 /**
  * Internal dependencies
  */
 import { REPORTS_STORE_NAME } from '../data';
 import SyncIssuesTable from './SyncIssuesTable';
-import { __, sprintf } from '@wordpress/i18n';
 import { useSettingsSelect } from '../../setup-guide/app/helpers/effects';
 
 const SyncIssues = () => {
--- a/assets/source/catalog-sync/sections/SyncIssues.js
+++ b/assets/source/catalog-sync/sections/SyncIssues.js
@@ -1,6 +1,7 @@
 /**
  * External dependencies
  */
+import { __, sprintf } from '@wordpress/i18n';
 import { useSelect } from '@wordpress/data';
 import { createInterpolateElement, useState } from '@wordpress/element';
 import { getHistory, getQuery, onQueryChange } from '@woocommerce/navigation';
@@ -8,7 +9,7 @@ import {
 	ExternalLink,
 	__experimentalText as Text, // eslint-disable-line @wordpress/no-unsafe-wp-apis --- _experimentalText unlikely to change/disappear and also used by WC Core
 } from '@wordpress/components';
-import { __, sprintf } from '@wordpress/i18n';
+
 /**
  * Internal dependencies
  */
--- a/assets/source/catalog-sync/sections/SyncIssues.js
+++ b/assets/source/catalog-sync/sections/SyncIssues.js
@@ -2,16 +2,24 @@
  * External dependencies
  */
 import { useSelect } from '@wordpress/data';
-import { useState } from '@wordpress/element';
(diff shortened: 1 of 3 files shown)


message-dimke:
Hey, @brezocordero !

The issue is a bit deeper than @ecgan described within his #936 ticket.

With the Pinterest API v3 endpoint we were downloading a file with all the feed item issues and parsing it. We could find a total number of records and build a proper pagination. However, this is not possible anymore. Pinterest v5 API removed the file and replaced it with an API call [https://developers.pinterest.com/docs/api/v5/#operation/items_issues/list](https://developers.pinterest.com/docs/api/v5/#operation/items_issues/list).

The fix you did does a proper paging for 250 first records, which is the maximum amount of records available in one call. When the solution may fit most of our users, there might be others, who have more records. With the current Pinterest v5 API implementation, it is not possible to know the total amount of items and build the paging. We could build a Prev/Next kind of paging.

Another issue with paging is that we would like to have error type of records go before all the warnings ones. That is why you could see sorting made before the output. And that is the reason I am fetching 250 records with the API call.

brezocordero:
> Another issue with paging is that we would like to have error type of records go before all the warnings ones. That is why you could see sorting made before the output. And that is the reason I am fetching 250 records with the API call. 

> it is not possible to know the total amount of items and build the paging. 

Thanks for the explanation @message-dimke . I noticed the limitations when looking at the code. It is not easy to have pagination + errors first with those constraints.

> When the solution may fit most of our users, there might be others, who have more records.

Do you know if they could see all the issues on `Pinterest > Catalog and Product Groups > Diagnostics`?

If they can, one option would be to display a message for those users, letting them know that we only display up to 250 issues and that they can see them all on `Pinterest > Catalog and Product Groups > Diagnostics`. What do you think?

message-dimke:
Hi, @brezocordero !

I spoke with @budzanowski about this, and we think your idea of having a message and a link is good. However, I can not yet confirm that `Pinterest > Catalog and Product Groups > Diagnostics` has the required records. My Pinterent account got disapproved, and I need it recovered first. It may take some time.

message-dimke:
The link to the feed diagnostics page on Pinterest is 

```
https://www.pinterest.com/business/catalogs/[Catalog ID]/diagnosticsv2/?advertiserId=[Advertiser ID]
```

message-dimke:
Added a message. In case we have 250 (which is a limit of items per page), show a message and a link to the Pinterest Diagnostics page.

Products_Catalog_‹_Pinterest_‹_Marketing_‹_WooCommerce_‹_WordPress_Pinterest_—_WooCommerce

PR #1028, opened 2024-06-25
Release/1.4.3

message-dimke:
Release 1.4.3 (WPCOM Atomic only)


rawdreeg:
I think this should have gone into develop, and trunk should remain in 1.3.x in case we need to release a hotfix to WP.org users, right? @message-dimke 

Edited to add: Ow we merged 1.4.0 already, so I suppose that ship has sailed :-)

PR #1046, opened 2024-08-02
Disconnect plugin on deactivation.

### LINKED ISSUE #1047: Deactivating plugin should disconnect the user from pinterest
Disconnect the user on the plugin deactivation event.

message-dimke:

Closes #1047

### Changelog entry

> Add - Call to disconnect from Pinterest on deactivation.


budzanowski (approved):
Looks good if completed.

I understand that the try/catch block was removed because disconnect handles it on its own now? Correct?

message-dimke:
> Looks good if completed.
> 
> I understand that the try/catch block was removed because disconnect handles it on its own now? Correct?

Yes. Disconnect does not throw any exceptions, it returns a boolean flag. So try/catch around it is not useful.

PR #1055, opened 2024-08-23
Fixing Pinterest Save button positioning for Twenty Twenty-Four theme.

### LINKED ISSUE #1054: Twenty-Twenty Four theme Pinterest save button is out of product card.
### Describe the bug:

Shop and Shop Category product cards have Pinterest Save button outside the card.

Cursor_and_Shop_–_Wildly_Involved

Music_–_Wildly_Involved

### Steps to reproduce:

1. Install Pinterest extension with Twenty Twenty-Four.
2. Navigate to the Shop.
3. Hover any product card.
4. Observe Pinterest Save button is out of bounds.

### Expected behavior:

Pinterest Save button is inside the product card when you hover the card.

### Actual behavior:

Pinterest Save button is out of the product card boundaries.

message-dimke:

Closes #1054.

Pinterest Save button has CSS absolute positioning when the product card has no CSS relative position rule on it which makes Pinterest Save button align relatively to the document, instead of a product card element.

### Screenshots:

Cursor_and_Shop_–_Wildly_Involved

Music_–_Wildly_Involved

1. Install Pinterest extension with Twenty Twenty-Four.
2. Navigate to the Shop.
3. Hover any product card.
4. Observe Pinterest Save button is out of bounds.

### Changelog entry

> Fix - Pinterest Save button positioning.

### DIFF
--- a/assets/source/sass/frontend/pinterest-for-woocommerce-pins.scss
+++ b/assets/source/sass/frontend/pinterest-for-woocommerce-pins.scss
@@ -28,12 +28,6 @@
 .wc-block-product {
 
 	&.product {
-
 		position: relative;
-
-		.pinterest-for-woocommerce-image-wrapper {
-			top: unset;
-			left: unset;
-		}
 	}
 }
--- a/assets/source/sass/frontend/pinterest-for-woocommerce-pins.scss
+++ b/assets/source/sass/frontend/pinterest-for-woocommerce-pins.scss
@@ -24,9 +24,11 @@
 
 
 // Twenty Twenty-Four compatibility.
-.wp-block-post {
+.wp-block-product {
 
 	&.product {
+
+		position: relative;
 
 		.pinterest-for-woocommerce-image-wrapper {
 			top: unset;
--- a/assets/source/sass/frontend/pinterest-for-woocommerce-pins.scss
+++ b/assets/source/sass/frontend/pinterest-for-woocommerce-pins.scss
@@ -24,7 +24,8 @@
 
 
 // Twenty Twenty-Four compatibility.
-.wp-block-product {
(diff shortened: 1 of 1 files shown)


assets/source/sass/frontend/pinterest-for-woocommerce-pins.scss
```diff
@@ -24,10 +24,12 @@
 
 
 // Twenty Twenty-Four compatibility.
-.wp-block-post {
+.wp-block-product {
```
ecgan: I believe you mean this?

```suggestion
.wc-block-product {
```

ecgan (changes requested):
@message-dimke , leaving one finding below.

assets/source/sass/frontend/pinterest-for-woocommerce-pins.scss
ecgan: @message-dimke , do you happen to know when and where `wp-block-post` is changed to `wc-block-product`? I was searching around but couldn't find the info.

Also, maybe for backward compatibility, we need to have both, like the following, in case some users are using some older versions? (need to test this out.)

```suggestion
.wp-block-post,
.wc-block-product {
```

ecgan:
@message-dimke , I was testing around and I noticed that the Pinterest "Save" button does not show up in the product page with Twenty Twenty-Four theme.

Screenshot of product image in product page:

image

Screenshot of related products in product page - note that for this one, it is using `wp-block-post` CSS class, not `wc-block-product`, and it does not have the `pinterest-for-woocommerce-image-wrapper` div:

image

Maybe you could consider opening separate issues / PRs to fix them.

PR #1061, opened 2024-08-29
Fix no product in page visit tracking

### LINKED ISSUE #1059: Call to a member function `get_id()` on bool
### Describe the bug:

The plugin is causing a critical WordPress error. 

`wp-content/plugins/pinterest-for-woocommerce/src/Tracking.php:94` Call to a member function get_id() on bool. 

All was working before the update (to the 1.4.7 version), so I am not sure if it is a product issue, but please let me know how we can fix it (the plugin has been deactivated for now to prevent the error from breaking the site).

### Additional details:

My wild guess is that it is a product-related issue. The check for `if is_product()` passes, but then `wc_get_product()` returns `bool` value. And calling `get_id()` on `bool` obviously fails.

budzanowski:

Closes #1059 .

There is no good reason behind this PR except that I am adding a defensive code to check if the product has been fetched.

In theory `is_product()` and `wc_get_product()` should return consistent results. Since we are observing the issue it means that they are not. Without access to the merchant site and proper debugging I don't see a way to getting to the root cause of this issue.

### Screenshots:
No screenshots.

1. Tracking should still work, especially for page_visit

I don't know how to reproduce this issue so we can only check if there is no regression and hope that the defensive code will fix the problem.

### Changelog entry

Fix - Detect no product error in the page_visit tracking.

### DIFF
--- a/src/Tracking.php
+++ b/src/Tracking.php
@@ -77,6 +77,7 @@ class Tracking {
 	 * Used as a callback for the wp_footer hook.
 	 *
 	 * @since 1.4.0
+	 * @since x.x.x Added checkf for product page.
 	 *
 	 * @return void
 	 */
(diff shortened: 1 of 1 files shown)


message-dimke (approved):
Looks good! 

Thank you, @budzanowski !

PR #1067, opened 2024-09-16
Reuse existing feed

### LINKED ISSUE #1060: `25` Feed Deletion Failed messages
### Describe the bug:

A message from the forum thread https://wordpress.org/support/topic/pinterest-for-woocommerce-feed-deletion-failed/#post-17982223

> _I’ve got the same message. In fact I have 25 of them. It does not help to dismiss them. They just come back on next reload…_

ViG9EQx

message-dimke:

Closes #1060

The PR aims to fix multiple things related to the feed deletion issue.

- The update procedure will remove duplicates of the Feed Deletion Failure notice.
- Feed deletion failure notice also got an updated text.
- Existing feeds to be fetched from Pinterest and reactivated, if any.
- Removing the condition to auto-disconnect the plugin in case of 403 Pinterest API error response (Pinterest returns 403 when an unapproved merchant can not delete their feed, which is not an access token-related issue, and we do not need to disconnect in this case).

### Changelog entry

> Fix - Feed Deletion Failure notice duplicates removal.
> Fix - Reuse existing feed, if any.
> Fix - 403 Pinterest API error response is not the reason to auto-disconnect.
> Dev - Tests suits update.

### DIFF
--- a/src/Feeds.php
+++ b/src/Feeds.php
@@ -315,8 +315,8 @@ class Feeds {
 	private static function does_feed_match( $feed ): bool {
 		$local_country = Pinterest_For_Woocommerce::get_base_country();
 		$local_locale  = LocaleMapper::get_locale_for_api();
-		$does_match = $local_country === $feed['default_country'] ?? '';
-		$does_match = $does_match && $local_locale === $feed['default_locale'] ?? '';
+		$does_match    = $local_country === $feed['default_country'] ?? '';
+		$does_match    = $does_match && $local_locale === $feed['default_locale'] ?? '';
 		return $does_match && 0 === strpos( $feed['location'] ?? '', get_site_url() );
 	}
 
--- a/src/Feeds.php
+++ b/src/Feeds.php
@@ -329,7 +329,7 @@ class Feeds {
 	}
 
 	/**
-	 * Tries to match remote feeds against local website configuration to find an existing feed, if any.
+	 * Compare remote feeds to local configuration to find a matching feed.
 	 *
 	 * @since x.x.x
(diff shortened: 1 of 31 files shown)


src/Feeds.php
```diff
@@ -311,6 +308,46 @@ public static function match_local_feed_configuration_to_registered_feeds( array
...
+	 * @return bool
+	 */
+	private static function does_feed_match( $feed ): bool {
+		$local_country = Pinterest_For_Woocommerce::get_base_country();
+		try {
+			$local_locale = LocaleMapper::get_locale_for_api();
+		} catch ( PinterestApiLocaleException $e ) {
+			$local_locale = LocaleMapper::PINTEREST_DEFAULT_LOCALE;
+		}
+		$does_match    = $local_country === $feed['default_country'] ?? '';
```
budzanowski: My personal preference is to have early exit on each condition.

src/Feeds.php
```diff
@@ -311,6 +308,46 @@ public static function match_local_feed_configuration_to_registered_feeds( array
...
+		$does_match    = $does_match && $local_locale === $feed['default_locale'] ?? '';
+		return $does_match && 0 === strpos( $feed['location'] ?? '', get_site_url() );
+	}
+
+	/**
+	 * Tries to match remote feeds against local website configuration to find an existing feed, if any.
```
budzanowski:
```suggestion
	 * Compare remote feeds to local configuration to find a matching feed.
```

budzanowski (approved):
Left two comments, a nitpicks, looks good. Approving.

PR #1077, opened 2024-10-22
Making `integration_data` data not mandatory.

message-dimke:

Integration data appeared not mandatory for the extension to function. However, we were using it to identify if the user connected status. Since the Pinterest Commerce Integration request may fail, we do not want the extension to rely on its presence.

```php
add_filter(
	'pre_http_request',
		if (
		) {
			return array(
				'headers' => array(
				),
				'body' => json_encode(
					array(
						'code'    => 911,
					)
				),
				'response' => array(
					'code'    => 500,
				),
				'cookies'  => array(),
				'filename' => '',
			);
		}

		return $response;
	},
	10,
	3
);
```

3. You should not see any changes. Nothing must fail.

### Changelog entry

> Update - Make `integration_data` optional for the extension.


budzanowski (approved):
Looks good!

PR #1078, opened 2024-10-22
Retrying Commerce Integration Creation on Falure.

message-dimke:

This PR introduces the initial create Commerce Integration call retry procedure. It may happen that the Create Integration API will fail. We have to retry its creation since the integration data is crucial for assigning customers to a proper Pinterest advertisement platform on the Pinterest side.

The code will try three times: in a minute after the initial failure; in an hour; in a day.

```php
add_filter(
	'pre_http_request',
		if (
		) {
			return array(
				'headers' => array(
				),
				'body' => json_encode(
					array(
						'code'    => 911,
					)
				),
				'response' => array(
					'code'    => 500,
				),
				'cookies'  => array(),
				'filename' => '',
			);
		}

		return $response;
	},
	10,
	3
);
```

3. During the connection procedure, the commerce integration call will fail. Look into Pending Scheduled Actions for a retry scheduled to run within a minute of the connection time.

WooCommerce_status_‹_WordPress_Pinterest_—_WordPress-2

It should also have an `attempt` number parameter set to `1`, meaning this is the first attempt of a retry.

4. Look into `pinterest-for-woocommerce` log for a message about the failure and the retry procedure.

WooCommerce_status_‹_WordPress_Pinterest_—_WordPress-3

5. Wait until the action triggers itself or force-run it. Since we still have a code to make the endpoint fail, we should see another retry attempt scheduled and the log message.

WooCommerce_status_‹_WordPress_Pinterest_—_WordPress-4

6. Do not wait an hour 🤡 to pass and force-run the action.
7. Look for the last, third attempt to get scheduled.

WooCommerce_status_‹_WordPress_Pinterest_—_WordPress-5

8. Force-run it and check no more actions scheduled and the last attempt log message written.

WooCommerce_status_‹_WordPress_Pinterest_—_WordPress-7

### Changelog entry

> Add - Failed Create Commerce Integration API call retries procedure.

### DIFF
--- a/class-pinterest-for-woocommerce.php
+++ b/class-pinterest-for-woocommerce.php
@@ -285,6 +285,7 @@ if ( ! class_exists( 'Pinterest_For_Woocommerce' ) ) :
 			add_action( 'init', array( Pinterest\Billing::class, 'schedule_event' ) );
 			add_action( 'init', array( Pinterest\AdCredits::class, 'schedule_event' ) );
 			add_action( 'init', array( Pinterest\RefreshToken::class, 'schedule_event' ) );
+			add_action( 'init', array( Pinterest\CommerceIntegration::class, 'init' ) );
 
 			// Register the marketing channel if the feature is included.
 			if ( defined( 'WC_MCM_EXISTS' ) ) {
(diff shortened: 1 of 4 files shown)


budzanowski (approved):
Looks good!

PR #1079, opened 2024-10-22
Synchronising Commerce Integration `partner_metadata` weekly.

message-dimke:

The PR adds a weekly commerce integration `partner_metadata` sync action. The action will attempt to create the integration if the one is still missing. The integration will also run through a set of retries if the API is not responding successfully, delaying the synchronization until the integration is finally created.

#### Commerce Integration Sync tries to create the integration if the one does not exist.

```php
add_filter(
	'pre_http_request',
		if (
		) {
			return array(
				'headers' => array(
				),
				'body' => json_encode(
					array(
						'code'    => 911,
					)
				),
				'response' => array(
					'code'    => 500,
				),
				'cookies'  => array(),
				'filename' => '',
			);
		}

		return $response;
	},
	10,
	3
);
```

3. Manually trigger the Commerce Integration retry `pinterest-for-woocommerce-create-commerce-integration-retry` actions so they will all run (fail to create the integration).
4. Manually trigger the `pinterest_for_woocommerce_weekly_heartbeat` action.

Since integration was not created previously, the sync action will attempt to create it. Fail and start the same commerce integration retry cycle (three attempts).

#### Commerce Integration Sync updates the `parent_metadata` if it has changed.

1. Clean up the codebase so there is no more commerce integration endpoint failure.
2. Either manually run the `pinterest_for_woocommerce_weekly_heartbeat` action to create commerce integration or reconnect to your Pinterest account.
3. Update the code that collects the current platform data, such as the extension, WP, and WC versions used by the installation.

Replace any `partner_metadata` array value with something to your taste

https://github.com/woocommerce/pinterest-for-woocommerce/blob/2ec929558cf27ab550adf8a107bbe9a06cbeeb1a/src/CommerceIntegration.php#L198-L204

4. Manually trigger the `pinterest_for_woocommerce_weekly_heartbeat` action from the Pending section of the ScheduledActions tab.
5. Observe the database `wp_options` key with `pinterest_for_woocommerce_data` name updated. 

**NOTE:** It is not possible to watch how the data goes to Pinterest API and gets saved there (under returning me the response with the updated data, I am assuming that Pinterest is saving it) without using a debugger and settings breakpoints.

### Screenshots

WooCommerce_status_‹_WordPress_Pinterest_—_WordPress-8

### Changelog entry

> Add - Commerce Integration `partner_metadata` weekly sync.

### DIFF
--- a/src/CommerceIntegration.php
+++ b/src/CommerceIntegration.php
@@ -34,6 +34,10 @@ class CommerceIntegration {
 			10,
 			1
 		);
+
+		if ( ! has_action( Heartbeat::WEEKLY, array( self::class, 'handle_sync' ) ) ) {
+			add_action( Heartbeat::WEEKLY, array( self::class, 'handle_sync' ) );
+		}
 	}
 
 	/**
@@ -44,6 +48,7 @@ class CommerceIntegration {
 	 */
 	public static function maybe_unregister_retries() {
 		as_unschedule_all_actions( 'pinterest-for-woocommerce-create-commerce-integration-retry' );
+		as_unschedule_all_actions( 'pinterest-for-woocommerce-sync-commerce-integration' );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


src/CommerceIntegration.php
```diff
@@ -119,17 +124,71 @@ public static function handle_create( $attempt = 0 ): array {
...
+		/*
+		 * If there is a commerce integration retry action, we know not to run the sync yet,
+		 * since it will also try to create the commerce integration as well.
+		 */
+		if ( as_has_scheduled_action( 'pinterest-for-woocommerce-create-commerce-integration-retry' ) ) {
+			return;
+		}
+
+		try {
+			$integration_data     = Pinterest_For_Woocommerce::get_data( 'integration_data' );
```
budzanowski, line 147: I would put this outside the `try` as it is safe from wha I see.

src/CommerceIntegration.php
```diff
@@ -119,17 +124,71 @@ public static function handle_create( $attempt = 0 ): array {
...
+			$external_business_id = $integration_data['external_business_id'] ?? '';
+
+			if ( ! $external_business_id ) {
+				self::handle_create();
+				return;
```
budzanowski: I would put this outside the `try` as it is safe from wha I see.

budzanowski (approved):
Left one small nitpick, looks great, I think that we can try it.

PR #1108, opened 2025-03-06
Replacing `get_locale()` with `WPLANG` option for feed registration.

message-dimke:

```
A merchant with a Pinterest for WooCommerce reporting continues to get emails with a "Your data source Created by Pinterest for WooCommerce has been deleted" message.
Their feed is resetting automatically a few times per day.
```

The issue is related to a feed registration process we have with Pinterest. Each 10 minutes we check if the feed parameters have changed, e.g. locale, currency

The change is related to how we obtain the locale. Since this particular issue was at one of the Canadian websites where there are two official languages and users were chaning locale from en to fr, the feed registration was detecting locale changes and re-registering the feed. Instead of using `get_local()` WP function we now take a `WPLANG` option value which is a site-wide setting and should not change frequently.

### Changelog entry

> Fix - Site locale is obtained from settings.

### DIFF
--- a/src/LocaleMapper.php
+++ b/src/LocaleMapper.php
@@ -120,6 +120,7 @@ class LocaleMapper {
 	 * @return string
 	 */
 	private static function get_wordpress_locale() {
+		// Using `WPLANG` instead `get_locale()` since the last may return inconsistent results depending on the user session language.
 		$wordpress_locale = get_option( 'WPLANG', self::PINTEREST_DEFAULT_LOCALE );
 		return str_replace( '_', '-', $wordpress_locale );
 	}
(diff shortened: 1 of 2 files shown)


shohanhossainnab:
For reference: 9425115-zen

rawdreeg (approved):
Changes look good! Thanks @message-dimke

src/LocaleMapper.php
```diff
@@ -120,7 +120,7 @@ public static function get_locale_for_api(): string {
 	 * @return string
 	 */
 	private static function get_wordpress_locale() {
```
budzanowski, line 122: Pleas adda comment why WPLANG is preferred over get_locale.

budzanowski (approved):
Looks good, pleas just add the requested comment for future devs :)

PR #1118, opened 2025-04-10
Add basic QIT

layoutd:
This PR schedules QIT tests for Pinterest for WooCommerce for Tuesdays at 4AM.

Tests also run automatically when a PR is merged into trunk and can be triggered manually if needed.

Most of the code is adapted from pdqkbG-5bb-p2.

The implementation follows this image (except for PHPStan check that it is disabled - pcShBQ-322-p2):
image


layoutd (commented):
OK for initial merge

PR #1111, opened 2025-03-18
Revert some changes

layoutd:
### Changes proposed in this Pull Request:

Revert some changes not ready for release.

### Detailed test instructions:
1. General smoke testing

### Additional details:
No changelog 

### Changelog entry

>

PR #1120, opened 2025-04-30
Disable CAPI for all merchants in update process

budzanowski:
Closes PIN4WOO-51

## Summary
- Adds an update procedure in version 1.4.18 that disables the Conversions API (CAPI) for all merchants
- Ensures merchants that already had CAPI enabled will have it disabled after update
- Implements the fix requested by PR #1111 which disabled CAPI only for new merchants

## Test plan
1. Install a previous version of the plugin
2. Enable CAPI tracking
3. Upgrade to the version with this change
4. Verify CAPI tracking has been disabled

### DIFF
--- a/src/PluginUpdate.php
+++ b/src/PluginUpdate.php
@@ -137,7 +137,7 @@ class PluginUpdate {
 			'1.4.10' => array(
 				'feed_deletion_notice_cleanup',
 			),
-			'1.4.17' => array(
+			'1.4.18' => array(
 				'disable_capi_for_all_merchants',
 			),
 		);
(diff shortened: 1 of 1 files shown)


layoutd (commented):
This looks good, thanks @budzanowski. I'll test it in a bit to confirm.

Should we also consider disabling the feature altogether with a blanket filter just to be sure?
`add_filter( 'woocommerce_pinterest_disable_conversions_capi_tracking', '__return_true' );`

budzanowski:
@layoutd we will enable the feature back soon so I just wanted to make sure that everybody starts with false in options. The filters IMO does not add any additional "security" as someone can bypass it the same as they could change the option manually.

layoutd (commented):
LGTM, setting gets disabled after update.
:shipit:

layoutd (approved):

PR #1124, opened 2025-05-23
Enable Conversions API in OAuth flow and settings UI

budzanowski:
## Summary
- Updates OAuth flow to set track_conversions_capi from features['CAPI']
- Adds a Conversions API checkbox to the settings UI 
- Ensures CAPI can be toggled while respecting the track_conversions parent setting

## Screenshots
image

## Test plan
1. Reset plugin settings and reconnect to Pinterest
2. During OAuth flow, verify CAPI setting is properly stored
3. Check that the Conversions API checkbox appears in settings and functions correctly
4. Verify CAPI is only enabled when Track Conversions is also enabled

### DIFF
--- a/assets/source/setup-guide/app/steps/SetupPins.js
+++ b/assets/source/setup-guide/app/steps/SetupPins.js
@@ -120,6 +120,38 @@ const SetupPins = ( {} ) => {
 									/>
 									<CheckboxControl
 										label={ __(
+											'Conversions API',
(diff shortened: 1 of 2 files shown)


layoutd (approved):
Nice, setting is stored correctly after onboarding, new checkbox toggles it in the database, CAPI tracking behaves as expected with both the specific Conversions API setting on/off as well as the global Track conversions setting on/off.

PR #1125, opened 2025-05-26
Add CAPI enablement modal to encourage merchants to enable Conversions API

budzanowski:
## Summary
- Adds a modal to encourage merchants to enable Pinterest Conversions API (CAPI)
- Modal appears when track_conversions is enabled but track_conversions_capi is disabled
- Provides direct enablement of CAPI with proper database persistence

## Changes Made
- **CapiEnablementModal.js**: New modal component following existing modal patterns
- **App.js**: Added modal state management and trigger logic to CatalogSyncApp
- **style.scss**: Removed unused notice styles (switched from notice to modal approach)

## Technical Details
- Uses `useSettingsDispatch(true)` for immediate database persistence
- Follows same UX patterns as existing onboarding modals
- Modal shows automatically when conditions are met (for testing - dismiss logic can be added later)
- Includes proper error handling and success notifications

## Screenshot
image

## Test Plan
1. Ensure track_conversions is enabled and track_conversions_capi is disabled
2. Navigate to catalog sync page
3. Verify modal appears automatically
4. Click "Enable Conversions API" and verify setting is saved to database
5. Verify modal closes and success notification appears
6. Check that CAPI is now properly enabled in tracking


layoutd (changes requested):
🟢 The modal displays when Conversions API tracking is disabled
🟢 Enabling Conversions API tracking with the modal successfully updates the setting

🟡 If I disable Conversions API tracking in the settings, I see the modal again (even if I previously enabled it using the modal). Do we want it to be a nag reminder if someone disables Conversion API tracking? Or should having Conversions API tracking (or enabling it) also count as having dismissed the modal?

🔴 If I dismiss the modal, no dismiss setting is saved. If I reload the catalog page with Conversions API tracking disabled (CMD+F5), I see the modal again. I don't see any indication of the modal dismissal being registered in the DB. Seems like the API request fails:
```json
{
    "code": "pinterest-for-woocommerce_user_interaction",
    "message": "Unrecognized interaction parameter",
    "data": {
        "status": 400
    }
}
```
Should this new option be handled by the backend in https://github.com/woocommerce/pinterest-for-woocommerce/blob/a4fc25c34c6411ea8df7fb9bed6b55cd8d6dee32/src/API/UserInteraction.php ?

budzanowski:
> 🟡 If I disable Conversions API tracking in the settings, I see the modal again (even if I previously enabled it using the modal). Do we want it to be a nag reminder if someone disables Conversion API tracking? Or should having Conversions API tracking (or enabling it) also count as having dismissed the modal?

I would leave it as it it. A bit annoying but it also serves as a reminder in case somebody disables it by accident.

> 🔴 If I dismiss the modal, no dismiss setting is saved. If I reload the catalog page with Conversions API tracking disabled (CMD+F5), I see the modal again. I don't see any indication of the modal dismissal being registered in the DB. Seems like the API request fails:

Fixed.

budzanowski:
CAPI modal dismissal is now part of the interactions API response:
```
{
    "ads_modal_dismissed": true,
    "ads_notice_dismissed": false,
    "capi_modal_dismissed": true
}
```

Rejecting the modal sets the proper value. The value gets preserved. I will go ahead and merge PR without waiting for followup review.

PR #1126, opened 2025-05-29
Fix: Move CAPI disable procedure from 1.4.18 to 1.4.19

budzanowski:
## Summary
- Moves the `disable_capi_for_all_merchants` update procedure from version 1.4.18 to 1.4.19
- Since 1.4.18 was already released without this procedure, it needs to run in the next release

## Test plan
- [x] Verify the update procedure is correctly set for 1.4.19
- [x] Test that the procedure runs correctly during plugin update to 1.4.19


layoutd (approved):
👍🏻 Setting is disabled on update to new version

PR #1140, opened 2025-10-09
Update product catalog sync to support product_type limits

louwie17:

Closes #PIN4WOO-55

This updates the `product_type` sync logic to limit it to 5 categories and a 1000 character limit.

1. First load the main plugin or `develop` and create a product with more then 5 categories, or update an existing product with more then 5 categories.
2. Go to your Pinterest Catalog and it should show a warning that `product_type can only contain five elements separated by >`
3. Now load this branch and make an update to the product so that the XML re-generates ( you can check the `XML feed` status here and see if it is updated: `/wp-admin/admin.php?page=wc-admin&path=%2Fpinterest%2Fcatalog`
4. It should now limit to the first 5 categories
5. Go back to the Pinterest Catalog and see if the warning disappears ( I am not sure how long this takes )

**1000 character limit test**
1. Follow the instructions above, but instead add 5 categories of 240 characters or more ( WP may have a 250 character limit ). This will go over the 1000 character limit.
2. In this branch it will remove the 5th category to stay under the 1000 character limit. 

### Changelog entry

> Fix - Issue where `product_type` include more then 5 categories if a product included more.

### DIFF
--- a/src/ProductsXmlFeed.php
+++ b/src/ProductsXmlFeed.php
@@ -311,8 +311,7 @@ class ProductsXmlFeed {
 		$original_count = count( $taxonomies );
 		if ( $original_count > self::PRODUCT_TYPE_CATEGORIES_LIMIT ) {
 			$taxonomies = array_slice( $taxonomies, 0, self::PRODUCT_TYPE_CATEGORIES_LIMIT );
-			/* translators: 1: product id, 2: original category count, 3: limit */
-			Logger::log( sprintf( esc_html__( 'Product [%1$s] has %2$d categories, limiting to first %3$d as per Pinterest requirements.', 'pinterest-for-woocommerce' ), $id, $original_count, self::PRODUCT_TYPE_CATEGORIES_LIMIT ) );
+			Logger::log( sprintf( 'Product [%1$s] has %2$d categories, limiting to first %3$d as per Pinterest requirements.', $id, $original_count, self::PRODUCT_TYPE_CATEGORIES_LIMIT ) );
 		}
 
 		// Build product_type string.
@@ -320,8 +319,7 @@ class ProductsXmlFeed {
 
 		// Ensure product_type doesn't exceed 1000 character limit.
 		if ( strlen( $product_type ) > self::PRODUCT_TYPE_CHARS_LIMIT ) {
-			/* translators: 1: product id, 2: original length, 3: limit */
-			Logger::log( sprintf( esc_html__( 'Product [%1$s] product_type length is %2$d characters, truncating to %3$d characters as per Pinterest requirements.', 'pinterest-for-woocommerce' ), $id, strlen( $product_type ), self::PRODUCT_TYPE_CHARS_LIMIT ) );
+			Logger::log( sprintf( 'Product [%1$s] product_type length is %2$d characters, truncating to %3$d characters as per Pinterest requirements.', $id, strlen( $product_type ), self::PRODUCT_TYPE_CHARS_LIMIT ) );
 
 			// Build product_type by adding taxonomies until we hit the character limit, we always include the first taxonomy.
 			$product_type = '';
--- a/src/ProductsXmlFeed.php
+++ b/src/ProductsXmlFeed.php
@@ -270,6 +270,7 @@ class ProductsXmlFeed {
 		 *
 		 * @param bool       $apply_shortcodes Shortcodes are applied if set to `true` and stripped out if set to `false`.
 		 * @param WC_Product $product          WooCommerce product object.
+		 *
 		 * phpcs:disable WooCommerce.Commenting.CommentHooks.MissingSinceComment
 		 */
 		$apply_shortcodes = apply_filters( 'pinterest_for_woocommerce_product_description_apply_shortcodes', false, $product );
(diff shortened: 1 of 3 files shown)


src/ProductsXmlFeed.php
```diff
@@ -292,7 +307,36 @@ private static function get_property_g_product_type( $product, $property ) {
 			return;
 		}
 
-		return '<' . $property . '>' . implode( ' &gt; ', $taxonomies ) . '</' . $property . '>';
+		// Limit to first 5 categories as per Pinterest requirements.
+		$original_count = count( $taxonomies );
+		if ( $original_count > self::PRODUCT_TYPE_CATEGORIES_LIMIT ) {
+			$taxonomies = array_slice( $taxonomies, 0, self::PRODUCT_TYPE_CATEGORIES_LIMIT );
+			/* translators: 1: product id, 2: original category count, 3: limit */
+			Logger::log( sprintf( esc_html__( 'Product [%1$s] has %2$d categories, limiting to first %3$d as per Pinterest requirements.', 'pinterest-for-woocommerce' ), $id, $original_count, self::PRODUCT_TYPE_CATEGORIES_LIMIT ) );
```
layoutd: Does these log messages need to use `esc_html__(`? Could it just be `__(`?

src/ProductsXmlFeed.php
```diff
@@ -292,7 +307,36 @@ private static function get_property_g_product_type( $product, $property ) {
...
+
+		// Build product_type string.
+		$product_type = implode( ' &gt; ', $taxonomies );
+
+		// Ensure product_type doesn't exceed 1000 character limit.
+		if ( strlen( $product_type ) > self::PRODUCT_TYPE_CHARS_LIMIT ) {
+			/* translators: 1: product id, 2: original length, 3: limit */
+			Logger::log( sprintf( esc_html__( 'Product [%1$s] product_type length is %2$d characters, truncating to %3$d characters as per Pinterest requirements.', 'pinterest-for-woocommerce' ), $id, strlen( $product_type ), self::PRODUCT_TYPE_CHARS_LIMIT ) );
+
+			// Build product_type by adding taxonomies until we hit the character limit, we always include the first taxonomy.
```
layoutd, line 324: > we always include the first taxonomy.

AFAICS my categories are limited to 200 chars by default. But since the first category is always included, should it be truncated if over `PRODUCT_TYPE_CHARS_LIMIT` in length?

src/ProductsXmlFeed.php
```diff
@@ -292,7 +307,36 @@ private static function get_property_g_product_type( $product, $property ) {
...
+			$taxonomies = array_slice( $taxonomies, 0, self::PRODUCT_TYPE_CATEGORIES_LIMIT );
+			/* translators: 1: product id, 2: original category count, 3: limit */
+			Logger::log( sprintf( esc_html__( 'Product [%1$s] has %2$d categories, limiting to first %3$d as per Pinterest requirements.', 'pinterest-for-woocommerce' ), $id, $original_count, self::PRODUCT_TYPE_CATEGORIES_LIMIT ) );
+		}
+
+		// Build product_type string.
+		$product_type = implode( ' &gt; ', $taxonomies );
+
+		// Ensure product_type doesn't exceed 1000 character limit.
+		if ( strlen( $product_type ) > self::PRODUCT_TYPE_CHARS_LIMIT ) {
```
layoutd, line 321: Minor: do you know if the limit includes html entities or decoded values? This `strlen` is of the HTML-encoded value so `€` becomes `&euro;` etc, which may artificially increase the length.

tests/Unit/FeedXMLGenerationTest.php
```diff
@@ -791,6 +799,166 @@ public function tearDown(): void {
...
+	 * @group feed
+	 */
+	public function testProductTypeWithFiveCategoriesNoWarning() {
+		$product_type_method = $this->getProductsXmlFeedAttributeMethod( 'g:product_type' );
+		$product             = WC_Helper_Product::create_simple_product();
+
+		/**
+		 * Mock logger object that will catch any logged messages.
+		 */
+		$mock_logger = new class() {
```
layoutd: Minor: this code is repeated multiple times, could be extracted to a `getMockLogger()`

layoutd (approved):
As mentioned, my merchant account is also kaput. But I tested the functionality by reviewing the changes in the generated XML feed file, and the testing instructions work as expected. 

- All categories after the 5th were removed successfully.
- Category names were limited to 200 characters, but including the ` &gt; ` separators put the five long categories over the 1000 character limit, and the last one was removed successfully. 
- New tests all pass correctly.

I left a few minor inline comments, and I guess final approval would depend on getting a working merchant account to confirm the changes succeed in hiding the error message. Also, confirmation around the doubt about the 1000 character limit being for the entire list of categories, or for each individual category.

src/ProductsXmlFeed.php
louwie17: > AFAICS my categories are limited to 200 chars by default. But since the first category is always included, should it be truncated if over PRODUCT_TYPE_CHARS_LIMIT in length?

I did say 'no', first I thought of just removing it and sending an empty `product_type`, but I think in this case ( although severely unlikely given the character limit ), I think its better for Pinterest to show the warning and the user to address the very long individual category name.
I felled better about auto removing any additional categories that would go over the limit.

src/ProductsXmlFeed.php
louwie17: I am not sure, but I think this would also be a very unlikely case.
I think we can re-consider if users run into this.

src/ProductsXmlFeed.php
louwie17: _Cursor_ 🤷‍♂️ 
I think just using `__(` is totally fine, although I would almost say given these are logs that we may not even want to translate them to begin with, let me see what the rest of the plugin does.

layoutd: Yeah, seems like not translated is probably fine too 😅

louwie17: Removed it: https://github.com/woocommerce/pinterest-for-woocommerce/pull/1140/commits/d4d6858408b8594f9578006780cc007f476ea8bf

tests/Unit/FeedXMLGenerationTest.php
louwie17: Makes sense, updated here: https://github.com/woocommerce/pinterest-for-woocommerce/pull/1140/commits/97fd659a509c15d1650cd2bc60d7ad6bf650ae93

src/ProductsXmlFeed.php
layoutd: > _Cursor_ 🤷‍♂️

Ironic because AI doesn't really even translations 🧠

louwie17:
> Also, confirmation around the doubt about the 1000 character limit being for the entire list of categories, or for each individual category.

@layoutd this is the total string including the separators, so the logic is correct in this PR.

louwie17:
Tested it using the feed debugger: https://nl.pinterest.com/business/catalogs/<site id>/data-sources/feed-debugger

PR #1145, opened 2025-11-17
Fix feed file cleanup on deactivation

layoutd:

Closes PIN4WOO63.

[Smoke tests](https://github.com/woocommerce/pinterest-for-woocommerce/wiki/Smoke-Tests#50-products-catalog-page) say the feed file is deleted on deactivation, but it wasn't working. This adds feed file deletion when Pinterest or WooCommerce are deactivated.

1. Connect to a Pinterest account (doesn't have to be approved).
2. Confirm feed file is generated: `/wp-content/pinterest-for-woocommerce-….xml`
3. Deactivate the Pinterest plugin.
4. Confirm feed file has been deleted.

### Additional details:
The random string appended to the feed file is changed when the plugin is activated, so feed files with that string are no longer updated.

### Changelog entry

> Fix - Remove feed file on deactivation.

### DIFF
--- a/pinterest-for-woocommerce.php
+++ b/pinterest-for-woocommerce.php
@@ -134,26 +134,26 @@ register_activation_hook(
 	}
 );
 
-// Register deactivation hook.
+/**
+ * Handle plugin deregistration.
+ */
+function pinterest_for_woocommerce_deregister() {
+	Automattic\WooCommerce\Pinterest\ProductSync::deregister();
+	Automattic\WooCommerce\Pinterest\Heartbeat::cancel_jobs();
+	Automattic\WooCommerce\Pinterest\ProductFeedStatus::deregister();
+	Pinterest_For_Woocommerce::disconnect();
+}
+
+// Register deactivation hook for this plugin.
 register_deactivation_hook(
 	PINTEREST_FOR_WOOCOMMERCE_PLUGIN_FILE,
-	function () {
-		Automattic\WooCommerce\Pinterest\ProductSync::cancel_jobs();
-		Automattic\WooCommerce\Pinterest\Heartbeat::cancel_jobs();
-		Automattic\WooCommerce\Pinterest\FeedGenerator::deregister();
-		Pinterest_For_Woocommerce::disconnect();
-	}
+	'pinterest_for_woocommerce_deregister'
 );
 
 // Register deactivation hook for WooCommerce.
 if ( defined( 'WC_PLUGIN_FILE' ) ) {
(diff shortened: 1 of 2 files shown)


pinterest-for-woocommerce.php
```diff
@@ -140,6 +140,7 @@ function () {
 	function () {
 		Automattic\WooCommerce\Pinterest\ProductSync::cancel_jobs();
 		Automattic\WooCommerce\Pinterest\Heartbeat::cancel_jobs();
+		Automattic\WooCommerce\Pinterest\FeedGenerator::deregister();
```
louwie17:
```suggestion
		Automattic\WooCommerce\Pinterest\ProductSync::deregister();
```
Should we use `ProductSync::deregister` instead? it in turns [calls the above and also calls](https://github.com/woocommerce/pinterest-for-woocommerce/blob/a0f59bb4c41753e58cb8ffa5ecadd56773ea7a9e/src/ProductSync.php#L207) `deregister` on a couple other items that may be relevant.

Suggesting this, as I did briefly see the `Successfully generated 5 minutes ago` message with the link to the original XML file, although it hadn't generated the new one yet and it was still pointing at the deleted version.

louwie17 (changes requested):
Thanks @layoutd for addressing this, this tested mostly well! Just noticed the XML feed still showing the old generated file when re-enabling Pinterest. However it did get updated relatively quickly, but left a alternative suggestion inline.

pinterest-for-woocommerce.php
layoutd: The method was `private`, but I've updated it to `public` and implemented it since it does seem more comprehensive: f256a4bcd086204bd5692e0ac44b14a47cba4b31

pinterest-for-woocommerce.php
```diff
@@ -134,24 +134,26 @@ function () {
 	}
 );
 
-// Register deactivation hook.
+/**
+ * Handle plugin deregistration.
+ */
+function pinterest_for_woocommerce_deregister() {
+	Automattic\WooCommerce\Pinterest\ProductSync::deregister();
```
louwie17, line 141: I thought you forgot `Automattic\WooCommerce\Pinterest\ProductSync::cancel_jobs();` by accident, but I see this is being triggered by the `deregister` function of `FeedGenerator` and `FeedRegistration` :)

louwie17 (approved):
Thanks for the updates @layoutd, this looks good 👍

PR #1155, opened 2026-01-27
Exclude README.md from built package

layoutd:

Closes PIN4WOO-66.

The repository development README.md file was being included in the built plugin package. This PR:
- Excludes README.md from the build process in gulpfile.js
- Updates the translation template (.pot file) with current version and timestamp
- Adds a readme to explain why the translation template is being versioned 
- Excludes that explanation README.md from the built plugin package 

1. Run `npm run build:zip` to generate the plugin package
2. Inspect the generated zip file contents
3. Verify that neither README.md is not included in the package
5. Verify that other expected files (PHP files, assets, etc.) are still included

### Changelog entry

> Fix - Exclude development README.md from built plugin package.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -188,6 +188,7 @@ gulp.task(
 					[
 						`${folder}/{assets,i18n,includes,src,vendor,views}/**/*`,
 						`${folder}/*.{php,txt,md}`,
+						`!${folder}/README.md`,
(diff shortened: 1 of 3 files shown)


chihsuan (approved):
LGTM and confirmed that the README.md files are excluded from the built zip.

PR #1187, opened 2026-05-25
Scope feed failure logging by feed

chihsuan:

Linear: PIN4WOO-86.

This builds on #1179 (unreleased) and prepares failed feed ingestion logging for future multi-feed support by:

- persisting the last logged processing result per Pinterest `feed_id` (an array keyed by feed ID) instead of a single global last value, so a failure logged for one feed no longer suppresses logging for another feed
- resolving the logged `feed_url` from the matching registered remote feed `location` (via `Feeds::get_feeds()`) instead of always using the first local feed config; logs an empty `feed_url` when no registered feed matches
- adding regression coverage for cross-feed dedup scoping and for remote feed IDs that differ from local feed config IDs

Because #1179 has not shipped, no migration is needed for the previously stored single-value dedup state.

These steps exercise the real scheduled-action code path, following the same approach as #1179. The site must be connected to Pinterest (so a feed file exists in `wp-content/uploads/` and `feed_registered` plugin data is set). Only the `processing_results` HTTP call is mocked — every other Pinterest API call in `register_feed()`, including the feeds list used to resolve `feed_url`, hits the real API.

1. Confirm preconditions — expect a non-empty Pinterest feed ID:
   ```bash
   wp eval 'echo \Pinterest_For_Woocommerce::get_data( "feed_registered" ), "\n";'
   ```
2. Drop the following mu-plugin into `wp-content/mu-plugins/pin4woo-86-test.php` (the leading `<?php` is required). It returns a stable `FAILED` processing result (a fixed `id`, so the dedup check can be verified across re-runs):
   <details><summary>pin4woo-86-test.php</summary>

   ```php
   <?php
   add_filter( 'pre_http_request', function ( $preempt, $args, $url ) {
       if ( false === strpos( $url, '/processing_results' ) ) {
           return $preempt;
       }
       return array(
           'headers'  => array( 'content-type' => 'application/json' ),
           'response' => array( 'code' => 200, 'message' => 'OK' ),
           'cookies'  => array(),
           'body'     => wp_json_encode( array(
               'items' => array(
                   array(
                       'id'                 => 'pin4woo-86-failed-1',
                       'status'             => 'FAILED',
                       'created_at'         => gmdate( 'c' ),
                       'validation_details' => array(
                           'errors'   => array( 'FETCH_ERROR' => 1 ),
                           'warnings' => array(),
                       ),
                       'product_counts'     => array( 'original' => 10, 'ingested' => 0 ),
                   ),
               ),
           ) ),
       );
   }, 10, 3 );
   ```

   </details>

3. Flush cached API responses, reset dedup state, and trigger the scheduled handler. The `processing_results` response is cached as a transient for 60s, so without flushing the handler returns the previously cached (live) result and the mock above never fires:
   ```bash
   wp transient delete --all
   wp eval '\Pinterest_For_Woocommerce::remove_data( "last_logged_processing_result_id" );'
   wp eval 'do_action( "pinterest-for-woocommerce-handle-feed-registration" );'
   ```
4. Go to `/wp-admin/admin.php?page=wc-status&tab=logs` and open the `pinterest-for-woocommerce-feed-ingestion-failure` log. Confirm the `Feed ingestion FAILED` block shows:
   - `feed_id:` matching the value from step 1
   - `processing_result_id: pin4woo-86-failed-1` (if you instead see a long numeric ID and a past `created_at`, the cache was not flushed and you are looking at live data — re-run step 3)
   - `feed_url:` matching the **registered remote feed's `location`** (the actual Pinterest catalog feed URL for this feed), not just the first local feed config
5. Confirm the dedup state is now an array keyed by feed ID (not a bare string):
   ```bash
   wp eval 'print_r( \Pinterest_For_Woocommerce::get_data( "last_logged_processing_result_id" ) );'
   ```
   Expect e.g. `Array ( [<feed_id>] => pin4woo-86-failed-1 )`.

Screenshot 2026-05-25 at 2 45 37 PM

6. Re-run the trigger to verify dedup — the same feed + same processing result must not be logged again. Do **not** flush again here, so the mocked `pin4woo-86-failed-1` result is served from its own 60s cache:
   ```bash
   wp eval 'do_action( "pinterest-for-woocommerce-handle-feed-registration" );'
   ```
   No additional `Feed ingestion FAILED` block is appended.
7. (Optional) Cross-check that the logged `feed_url` came from the registered feed `location`:
   ```bash
   wp eval '$f = \Automattic\WooCommerce\Pinterest\Feeds::get_feeds(); print_r( wp_list_pluck( $f, "location", "id" ) );'
   ```
8. Cleanup:
   ```bash
   rm wp-content/mu-plugins/pin4woo-86-test.php
   wp eval '\Pinterest_For_Woocommerce::remove_data( "last_logged_processing_result_id" );'
   ```

> **Note on multi-feed scoping:** cross-feed deduplication (a failure on feed A not suppressing feed B, and vice-versa) is hard to exercise manually on a single-feed install. It is covered by the unit tests in `tests/Unit/FeedStatusServiceTest.php` (`..._scopes_deduplication_by_feed_id` and `..._keeps_deduplication_per_feed_after_other_feed_logs`).

### Changelog entry

> Fix - Scope failed feed ingestion logging deduplication and feed URL resolution by feed.

### DIFF
--- a/src/FeedStatusService.php
+++ b/src/FeedStatusService.php
@@ -410,11 +410,24 @@ class FeedStatusService {
 	 * @return string
 	 */
 	private static function get_feed_url_for_feed_id( string $feed_id ): string {
-		foreach ( Feeds::get_feeds() as $feed ) {
+		$feeds = Feeds::get_feeds();
+
+		foreach ( $feeds as $feed ) {
 			if ( ( $feed['id'] ?? '' ) === $feed_id ) {
 				return $feed['location'] ?? '';
 			}
 		}
+
+		// Distinguish "no registered feeds returned" (e.g. feeds API unavailable) from
+		// "feeds returned but none match" (e.g. a stale or remote-only feed ID) so the
+		// empty feed_url in the failure log can be diagnosed.
+		Logger::log(
+			empty( $feeds )
+				? "Could not resolve feed_url: no registered feeds returned for feed_id={$feed_id}"
+				: "Could not resolve feed_url: no registered feed matched feed_id={$feed_id}",
+			'warning',
+			'feed-ingestion-failure'
+		);
 
 		return '';
 	}
--- a/src/FeedStatusService.php
+++ b/src/FeedStatusService.php
@@ -374,6 +374,9 @@ class FeedStatusService {
 		}
 
 		$feed_url = self::get_feed_url_for_feed_id( $feed_id );
+		if ( '' === $feed_url ) {
+			$feed_url = '(unresolved)';
+		}
 
 		Logger::log(
 			sprintf(
@@ -431,7 +434,11 @@ class FeedStatusService {
 	}
 
 	/**
-	 * Get the local feed URL for a Pinterest feed ID.
+	 * Get the registered feed URL (location) for a Pinterest feed ID.
+	 *
+	 * The feed ID is the remote Pinterest feed ID, so it is resolved against the
+	 * registered remote feeds. Returns an empty string when no remote feed matches
+	 * (e.g. the feeds API is unavailable).
 	 *
 	 * @param string $feed_id Pinterest feed ID.
 	 * @return string
@@ -440,14 +447,6 @@ class FeedStatusService {
 		foreach ( Feeds::get_feeds() as $feed ) {
 			if ( ( $feed['id'] ?? '' ) === $feed_id ) {
 				return $feed['location'] ?? '';
-			}
-		}
-
-		$configs = LocalFeedConfigs::get_instance()->get_configurations();
-
-		foreach ( $configs as $config ) {
-			if ( ( $config['feed_id'] ?? '' ) === $feed_id ) {
-				return $config['feed_url'] ?? '';
 			}
 		}
 
--- a/src/FeedStatusService.php
+++ b/src/FeedStatusService.php
@@ -377,9 +377,6 @@ class FeedStatusService {
 		}
 
 		$feed_url = self::get_feed_url_for_feed_id( $feed_id );
-		if ( '' === $feed_url ) {
-			$feed_url = '(unresolved)';
-		}
 
 		Logger::log(
 			sprintf(
--- a/src/FeedStatusService.php
+++ b/src/FeedStatusService.php
@@ -364,7 +364,10 @@ class FeedStatusService {
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


src/FeedStatusService.php
```diff
@@ -394,7 +389,69 @@ public static function log_failed_processing_result( string $feed_id, array $pro
...
+			}
+		}
+
+		$configs = LocalFeedConfigs::get_instance()->get_configurations();
+
+		foreach ( $configs as $config ) {
+			if ( ( $config['feed_id'] ?? '' ) === $feed_id ) {
+				return $config['feed_url'] ?? '';
+			}
+		}
```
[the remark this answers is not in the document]
chihsuan: Addressed in `594bbe75`: the local config fallback was removed because the value being matched here is the remote Pinterest feed ID, while local configs store local feed file IDs. The log now resolves the URL only from registered remote feed `location` values and logs `(unresolved)` when no remote feed matches.

src/FeedStatusService.php
```diff
@@ -394,7 +392,31 @@ public static function log_failed_processing_result( string $feed_id, array $pro
...
+	 * Get the registered feed URL (location) for a Pinterest feed ID.
+	 *
+	 * The feed ID is the remote Pinterest feed ID, so it is resolved against the
+	 * registered remote feeds. Returns an empty string when no remote feed matches
+	 * (e.g. the feeds API is unavailable).
+	 *
+	 * @param string $feed_id Pinterest feed ID.
+	 * @return string
+	 */
+	private static function get_feed_url_for_feed_id( string $feed_id ): string {
```
layoutd, line 412: Nit: could make sense use something other than `''` for to clarify the result when no feed is found. E.g., is `Feeds::get_feeds() === []` the same as feeds existing but none found matching the the provided ID?

layoutd (approved):
LGTM 👍🏻  All tests passed correctly
Left one small comment

src/FeedStatusService.php
chihsuan: Good call! Rather than swap in a sentinel value, I kept `feed_url` empty but added a `warning` log in `get_feed_url_for_feed_id()` that separates the two cases you mentioned — no feeds returned at all (API unavailable) vs. feeds returned but none matching the ID (stale/remote-only feed). That way the empty value is still diagnosable from the logs without muddying the field itself.

PR #1196, opened 2026-07-23
Prevent overlapping feed generation runs from corrupting the product feed

louwie17:
Fixes PIN4WOO-88 (follow-up to PIN4WOO-87 / #1166).

## Problem

Feed generation is a chained Action Scheduler job in which every chain action shares one cursor (`feed_last_queued_item_id`). Two generation runs can overlap because:

1. The framework's `is_running()` only sees its own `chain_start`/`chain_batch` hooks — the plugin's daily `pinterest-for-woocommerce-start-feed-generation` action is invisible to it, so after any dead run every product edit immediately schedules a new start.
2. After a batch dies with a timeout fatal, Action Scheduler marks it failed *before* firing `action_scheduler_unexpected_shutdown`; until the plugin reschedules the retry there are zero pending/in-progress chain actions, so the "already running" guard in `handle_start_action()` passes for a concurrently claimed start.

An overlapping run's batch #1 resets the shared cursor to 0, so runs rewind each other. On 1.4.27 this produces duplicate items in the feed, truncated feeds silently renamed to final and marked `generated`, and false "maximum batch limit of 1000 reached" circuit-breaker trips (see the wp.org thread "XML Feed Generation never ends" — a ~37k-item store capped at exactly 100,000 items).

## Fix

**1. Cycle identity.** When a start action executes, it mints a cycle ID (`wp_generate_uuid4()`), persists it as the current cycle and stamps it into the job args, which the framework propagates verbatim through the whole chain (`chain_start → chain_batch → … → chain_end`). Every chain handler validates its args' cycle ID against the current one and, when stale, aborts quietly — no processing, no successor action, no error status, no touching the cursor or the batch-size throttling state; just a log line. A stale `chain_end` in particular does **not** rename the tmp file nor set status `generated`. The timeout-retry path (`handle_unexpected_shutdown`) also skips stale actions, so a retry from a superseded cycle neither runs nor shrinks the current cycle's batch size.

Two cache-correctness details proved essential under load (the first stress run failed without them):

- The cycle ID lives in a **dedicated option** (`pinterest_for_woocommerce_feed_generation_cycle_id`), not inside the shared `pinterest_for_woocommerce_data` array — the shared array is rewritten wholesale by concurrent processes and would clobber the ID with a stale copy. It is **read with a direct `$wpdb` query**: Action Scheduler runners are long-lived requests whose `alloptions` cache predates a concurrent supersession, so `get_option()` (and the plugin's `get_data( …, $force )`, which only bypasses the plugin's static cache) would keep serving each process its own cycle as "current".
- The **feed cursor read** (`get_last_batch_id()`) also goes **directly to the database**: the previous batch may have committed the cursor from another process, and that write is invisible to the current request's options caches — a stale read here re-fetches already-written products and duplicates them.
- The **feed dirty flag also moved to a dedicated option** (`pinterest_for_woocommerce_feed_dirty`). `mark_feed_dirty()` runs on every product edit in requests whose options caches are stale; saving the flag through the shared data option rewrote the whole array from that stale snapshot and could rewind the feed cursor mid-generation (reproduced in the second stress run: one edit rolled the cursor back 700 IDs and ~540KB of items were appended twice). With a dedicated option, product edits never touch the shared array at all.

**2. Hard concurrency cap of one active cycle.** `handle_start_action()` first checks whether the current cycle is still genuinely alive — i.e. a pending or in-progress `chain_batch`/`chain_end` action carries the *current* cycle ID. If alive, the start does not begin a new cycle (and does not truncate the tmp file): it sets the dirty flag and returns, so the active run finishes and `handle_end()` triggers a single follow-up regeneration that coalesces all edits made in the meantime. Only when the current cycle is dead or absent does a start mint a new ID and take over. `mark_feed_dirty()` additionally debounces the scheduling side with a time-aware check (a start pending within the next 5 minutes suppresses rescheduling — the naive check would always be truthy because the daily recurring start is effectively always pending). Layering: the debounce is best-effort de-noising, the liveness check stops needless takeovers of a healthy run, and if a start still slips through the post-timeout blind window, part 1's supersession guarantees the old cycle's actions abort on their next execution — under every path at most one cycle actually writes.

**3. Hygiene.** `handle_error()` now deletes the `-tmp.xml` files (new `FeedFileOperations::delete_temporary_feed_files()`), including on the circuit-breaker path which deliberately schedules no retry, so failed/aborted cycles no longer leave multi-hundred-MB partials on disk.

## Updates since the original description (post-review, Aug 18-19)

- **Start serialization is now a real lock.** Cycle starts take an atomic lock in a dedicated option (`pinterest_for_woocommerce_feed_generation_start_lock`) before minting a cycle ID, so two concurrently claimed start actions can no longer both pass the liveness check. A start that cannot take the lock defers, same as deferring to a live cycle.
- **The feed cursor moved to its own option** (`pinterest_for_woocommerce_feed_last_queued_item_id`), out of the shared plugin data array, for the same clobbering reason as the cycle ID and dirty flag. The value is lazily migrated from the shared option on first read, so upgrades need no migration step.
- **The liveness check pages through all scheduled feed actions** instead of inspecting only the first page, so a live cycle cannot be missed behind a page of stale actions.

## Known limitations

- A batch already mid-execution at the moment a new cycle truncates the tmp file can still append its buffer once before it next validates; the residue is bounded to a single batch and is corrected by the next full regeneration.
- Actions scheduled by previous plugin versions carry no cycle ID. They keep running as long as no cycle ID has ever been minted (upgrade compatibility); after the first post-upgrade start they are treated as stale and abort, and the feed is refreshed by that same run.
- In the rare case where a stale pending batch keeps `is_running()` truthy after the current cycle already completed, a product edit only sets the dirty flag and regeneration waits for the next daily start.

## Testing instructions

Unit tests: `vendor/bin/phpunit --filter FeedGeneratorTest` (15 new tests covering cycle-ID propagation, stale batch/end aborts, start deferral vs. takeover, the start lock, cursor option and lazy migration, liveness pagination, stale timeout handling, and tmp cleanup).

Stress repro (wp-env, ~2k products + ~5.2k variations, harness that injects timeout fatals and widens the post-fatal race window — available as a ready-to-use plugin with a wp-admin control page: https://gist.github.com/louwie17/76292fe79b85de637298d331e2f53043 ):

Test with `develop` & this branch to see the difference.
1. Arm `pint_repro_timeout_every=10` and `pint_repro_widen_window=3`.
2. Run 3 parallel web-context Action Scheduler runner loops (`admin-ajax.php?action=pint_repro_run`) while editing a product every ~2 s for ~6 minutes; then disarm and let a clean cycle finish.
3. Before this fix: cursor regressions and a truncated final feed appear within minutes. With this fix: the trace shows no cursor regressions outside legitimate new-cycle starts, stale actions abort with "superseded generation cycle" log lines, and the final feed returns to the healthy size with no duplicate `<g:id>` entries.

### Changelog entry

> Fix - Prevent overlapping feed generation runs from corrupting the product feed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Dsb33QpwXpKeQe17c65ZEo

### DIFF
--- a/src/FeedGenerator.php
+++ b/src/FeedGenerator.php
@@ -53,10 +53,20 @@ class FeedGenerator extends AbstractChainedJob {
 	const OPTION_FEED_DIRTY = 'pinterest_for_woocommerce_feed_dirty';
 
 	/**
+	 * The option used as an atomic lock while a generation cycle is started.
+	 */
+	const OPTION_START_LOCK = 'pinterest_for_woocommerce_feed_generation_start_lock';
+
+	/**
 	 * How soon (in seconds) an already pending start action must fire for
 	 * mark_feed_dirty() to skip scheduling another immediate start.
 	 */
 	const START_DEBOUNCE_WINDOW = 5 * MINUTE_IN_SECONDS;
+
+	/**
+	 * Maximum lifetime for a start lock left behind by an interrupted request.
+	 */
+	const START_LOCK_TTL = 5 * MINUTE_IN_SECONDS;
 
 	/**
 	 * The time in seconds to wait after a failed feed generation attempt,
@@ -380,23 +390,34 @@ class FeedGenerator extends AbstractChainedJob {
 	 * @throws Throwable Related to creating an empty feed temp file and populating the header possible issues.
 	 */
 	public function handle_start_action( array $args ) {
-		if ( $this->is_current_cycle_alive() ) {
+		$start_lock = $this->acquire_start_lock();
+		if ( '' === $start_lock ) {
 			update_option( self::OPTION_FEED_DIRTY, 1, false );
-			self::log( __( 'Feed generation is already running. Marked the feed dirty to regenerate when the current cycle finishes.', 'pinterest-for-woocommerce' ) );
+			self::log( __( 'Another feed generation start is in progress. Marked the feed dirty to regenerate afterward.', 'pinterest-for-woocommerce' ) );
 			return;
 		}
 
-		// Mint the new cycle before truncating the temporary files: from this moment
-		// any still-scheduled action from an older cycle self-terminates.
-		$cycle_id = wp_generate_uuid4();
-		update_option( self::OPTION_CYCLE_ID, $cycle_id, false );
-		$args[ self::ARG_CYCLE_ID ] = $cycle_id;
+		try {
+			if ( $this->is_current_cycle_alive() ) {
+				update_option( self::OPTION_FEED_DIRTY, 1, false );
+				self::log( __( 'Feed generation is already running. Marked the feed dirty to regenerate when the current cycle finishes.', 'pinterest-for-woocommerce' ) );
+				return;
+			}
 
-		/* translators: feed generation cycle ID */
-		self::log( sprintf( __( 'Starting feed generation cycle `%s`.', 'pinterest-for-woocommerce' ), $cycle_id ) );
+			// Mint the new cycle before truncating the temporary files: from this moment
+			// any still-scheduled action from an older cycle self-terminates.
+			$cycle_id = wp_generate_uuid4();
+			update_option( self::OPTION_CYCLE_ID, $cycle_id, false );
+			$args[ self::ARG_CYCLE_ID ] = $cycle_id;
 
-		$this->handle_start();
-		$this->queue_batch( 1, $args );
+			/* translators: feed generation cycle ID */
+			self::log( sprintf( __( 'Starting feed generation cycle `%s`.', 'pinterest-for-woocommerce' ), $cycle_id ) );
+
+			$this->handle_start();
+			$this->queue_batch( 1, $args );
+		} finally {
+			$this->release_start_lock( $start_lock );
+		}
 	}
 
 	/**
@@ -836,6 +857,7 @@ class FeedGenerator extends AbstractChainedJob {
 		as_unschedule_all_actions( self::ACTION_START_FEED_GENERATOR, array(), PINTEREST_FOR_WOOCOMMERCE_PREFIX );
 		delete_option( self::OPTION_CYCLE_ID );
 		delete_option( self::OPTION_FEED_DIRTY );
+		delete_option( self::OPTION_START_LOCK );
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


src/FeedGenerator.php
```diff
@@ -160,6 +184,23 @@ public function handle_unexpected_shutdown( int $action_id, ?array $error ) {
 			return;
 		}
 
+		// Do not resurrect a superseded cycle, and do not let its timeout shrink the
+		// current cycle's batch size throttling state.
+		$job_args = ( isset( $args[1] ) && is_array( $args[1] ) ) ? $args[1] : array();
+		if ( $this->is_stale_cycle( $job_args ) ) {
```
chihsuan, line 205: Would it be worth mirroring this guard in `handle_failed_execution()` too? 

The timeout path checks ownership, but the exception path goes directly to `handle_error()`. If an old batch is superseded while running and then throws, `handle_error()` deletes the current cycle’s temporary files and marks its status as failed.
 
Because later batches append to the missing file, they recreate it without the XML header, and the current `chain_end` can still publish it as generated.

src/FeedGenerator.php
```diff
@@ -319,6 +360,38 @@ protected function handle_start() {
...
+		if ( $this->is_current_cycle_alive() ) {
+			update_option( self::OPTION_FEED_DIRTY, 1, false );
+			self::log( __( 'Feed generation is already running. Marked the feed dirty to regenerate when the current cycle finishes.', 'pinterest-for-woocommerce' ) );
+			return;
+		}
+
+		// Mint the new cycle before truncating the temporary files: from this moment
+		// any still-scheduled action from an older cycle self-terminates.
+		$cycle_id = wp_generate_uuid4();
+		update_option( self::OPTION_CYCLE_ID, $cycle_id, false );
```
chihsuan: I think there is still an ownership race between concurrent start actions.

`is_current_cycle_alive()` and `update_option()` are separate operations, so two starts can both decide that the previous cycle is dead. There is also a window after one start stores its cycle ID but before it queues the first batch: another start sees no live batch/end action and can take ownership.

The first start never checks ownership again before `handle_start()` truncates the shared temporary files. If that happens after the winning cycle has written a batch, the final feed can be well-formed but silently omit that batch while still being published as generated.

src/FeedGenerator.php
```diff
@@ -736,16 +870,30 @@ protected function get_batch_size(): int {
...
-		return (int) Pinterest_For_Woocommerce::get_data( 'feed_last_queued_item_id' );
+
+		global $wpdb;
+
+		$value    = $wpdb->get_var( // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
+			$wpdb->prepare(
+				"SELECT option_value FROM {$wpdb->options} WHERE option_name = %s LIMIT 1",
+				PINTEREST_FOR_WOOCOMMERCE_DATA_NAME
+			)
+		);
```
chihsuan: The cursor read now bypasses the options cache, but the corresponding write still goes through `save_data()`, which rewrites the entire shared `pinterest_for_woocommerce_data` option.

This leaves the cursor vulnerable to the same lost-update problem that motivated moving feed_dirty into a dedicated option. For example, `FeedRegistration::handle_feed_registration()` runs every 10 minutes and unconditionally calls `save_data()` twice while clearing merchant errors. A concurrent registration write can restore an older cursor snapshot, or a cursor write can clobber unrelated registration data.

Would it make sense to move `feed_last_queued_item_id` into its own `non-autoloaded` option as well, with direct reads and writes?

src/FeedGenerator.php
```diff
@@ -758,6 +906,94 @@ protected function set_last_batch_id( int $id ): void {
...
+			$this->get_action_full_name( self::CHAIN_BATCH ),
+			$this->get_action_full_name( self::CHAIN_END ),
+		);
+
+		foreach ( $hooks as $hook ) {
+			$actions = (array) $this->action_scheduler->search(
+				array(
+					'hook'     => $hook,
+					'status'   => array( ActionSchedulerInterface::STATUS_PENDING, ActionSchedulerInterface::STATUS_RUNNING ),
+					'per_page' => 50,
```
chihsuan: This search only inspects the first 50 matching actions, and Action Scheduler defaults to ascending date order. That means it examines the oldest actions first, while the current cycle’s batch is normally one of the newest. 

I wonder if we could paginate through all matching actions? Ordering by `DESC` would reduce the immediate risk, but pagination would avoid silently depending on queue size and ordering.

src/FeedGenerator.php
```diff
@@ -570,17 +686,29 @@ public function mark_feed_dirty(): void {
...
+	 * request after this request's options caches were primed.
+	 *
 	 * @since 1.0.10
 	 * @return bool Indicates if feed is dirty or not.
 	 */
 	public function feed_is_dirty(): bool {
-		return (bool) Pinterest_For_Woocommerce()::get_data( 'feed_dirty' );
+		global $wpdb;
+
+		$value = $wpdb->get_var( // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
```
chihsuan: Optional cleanup: the same uncached `wp_options` lookup and PHPCS suppressions now appear in three places.

Would a small private helper such as `get_uncached_option_value( string $option_name )` make the cache-bypass behavior easier to audit and keep the direct-query justification in one place? Each caller could still handle its own casting or unserialization.

chihsuan (commented):
Thanks for the thorough investigation and stress testing here! @louwie17  cycle identity looks like a solid direction.

I found a few remaining ownership gaps where stale or concurrent work could still affect the active cycle’s files or cursor. I’ve left inline suggestions.

As follow-up work, I think it may be also worth isolating temporary files per cycle and making batch writes idempotent. That would remove much of the shared-file coordination this PR currently needs and make retries safer by design.

src/FeedGenerator.php
layoutd: Good catch. Fixed in 64b22518. `handle_failed_execution()` now validates the action’s cycle before invoking shared error cleanup, with regression coverage confirming the active cycle remains untouched.

src/FeedGenerator.php
layoutd: Agreed. With 74d86c8, starts now acquire an atomic, expiring database lock before checking liveness and hold it through cycle creation, temporary-file preparation, and first-batch enqueue.

src/FeedGenerator.php
layoutd: Cursor reads and writes changed to use a dedicated non-autoloaded option with direct database access. I also added lazy migration for in-flight legacy chains. 91b5dd32.

src/FeedGenerator.php
layoutd: Good point, fixed in 748c110d. The liveness check now paginates through all matching actions, so it no longer depends on the first 50 results or their ordering.

src/FeedGenerator.php
layoutd: Good suggestion. 57ed7fd1 centralizes the uncached option lookup and its direct-query justification in a private helper, while callers retain their own casting.

layoutd:
@chihsuan I think I addressed all your comments last week, but forgot to request re-review. Would appreciate another green light

src/FeedGenerator.php
```diff
@@ -319,6 +382,49 @@ protected function handle_start() {
...
+	public function handle_start_action( array $args ) {
+		$start_lock = $this->acquire_start_lock();
+		if ( '' === $start_lock ) {
+			update_option( self::OPTION_FEED_DIRTY, 1, false );
+			self::log( __( 'Another feed generation start is in progress. Marked the feed dirty to regenerate afterward.', 'pinterest-for-woocommerce' ) );
+			return;
+		}
+
+		try {
+			if ( $this->is_current_cycle_alive() ) {
```
chihsuan, line 406: nit: Should `start_generation()` gate on this too? `is_running()` counts a superseded pending batch, so a daily start could return early.

chihsuan (approved):
Thanks @layoutd The lock and the dedicated options addressed everything I raised earlier. 👍

I left a nit note but not going to block this. Thanks!

PR #1202, opened 2026-08-24
Fix scheduled sale pricing in product feeds

jimjasson:

This PR prevents future and expired scheduled sale prices from being exported as active `sale_price` values. To detect sale price schedule, it uses WooCommerce’s schedule-aware `is_on_sale()` state for simple products and variations.

Moreover, it adds a `sale_price_effective_date` for active sales with defined start and end dates in UTC ISO 8601 format, as described in [Pinterest documentation](https://help.pinterest.com/en/business/article/before-you-get-started-with-catalogs).
 
Finally, it adds regression coverage for future, active, expired, partial, and variation sale schedules.

Closes PIN4WOO-92

1. Create a Simple Product with a Sale price.
2. Save it.
3. Go to Marketing > Pinterest and wait for a new XML file to be generated. run `wp action-scheduler run --group=pinterest-for-woocommerce --force` to run the AS actions to generate the feed.
4. Ensure that the product shows up with a `<sale_price>` in the XML. 
5. Re-edit the product and set a sale date in the future. 
6. Ensure that the product shows up without a `<sale_price>` in the XML.
7. Re-edit the product and set a sale date that covers the present date.
8. Ensure that the product shows up with a `<sale_price>` in the XML, and with a `<sale_price_effective_date>`.

### Changelog entry

> Fix - Fixed scheduled sale pricing in product feeds.

### DIFF
--- a/src/ProductsXmlFeed.php
+++ b/src/ProductsXmlFeed.php
@@ -42,6 +42,7 @@ class ProductsXmlFeed {
 		'g:availability',
 		'g:price',
 		'sale_price',
+		'sale_price_effective_date',
 		'g:mpn',
 		'g:brand',
 		'g:tax',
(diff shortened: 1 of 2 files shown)


layoutd (approved):
👍🏻 LGTM, thanks for fixing this.

Using WooCommerce’s schedule-aware `is_on_sale()` guard before emitting `sale_price` looks like the right approach here, and the new `sale_price_effective_date` handling is scoped correctly to active sales with both start/end dates defined.

The added regression coverage looks good too.

PR #1207, opened 2026-08-25
Fix PageVisit tracking on full-page cache hits

jimjasson:

Previously, the PageVisit `event_id` was generated in PHP and embedded in the rendered HTML. Full-page caches reused that ID across visitors, while cache hits skipped PHP entirely and therefore never dispatched the matching Conversions API event.

This PR:

- Generates the PageVisit `event_id` in the browser for every page view.
- Uses the same runtime ID for the Pinterest Tag and CAPI event.
- Sends PageVisit CAPI events through a background browser beacon, including on cache hits.
- Prevents synchronous PageVisit CAPI dispatch during HTML rendering, avoiding duplicate events on cache misses.
- Preserves the original storefront URL and the visitor’s current IP address and User-Agent.
- Rebuilds product data server-side from the constrained `product_id`.
- Respects tracking settings, consent filters, and crawler detection.
- Validates the event ID and requires the event source URL to belong to the store.
- Intentionally does not use a nonce for the public tracking beacon. Anonymous nonces embedded in cached HTML eventually expire and do not provide meaningful authorization because every logged-out visitor receives the same nonce. The endpoint performs no privileged action or site-state mutation.
- Adds regression tests for runtime ID generation, nonce-free CAPI dispatch, invalid event IDs, source URL preservation, Tag-only tracking, and duplicate prevention.

Closes PIN4WOO-94

#### Prerequisites

1. Connect Pinterest and enable Pinterest Tag tracking.
2. Enable Conversions API tracking.
3. Enable debug logging.
4. Enable WP Fastest Cache or another full-page cache.

#### Validate a cache MISS

1. Delete the full-page cache.
2. Open a fresh incognito window on `about:blank`.
3. Open Chrome DevTools → Network.
4. Enable **Preserve log** and leave **Disable cache** unchecked.
5. Visit a product page.
6. Select the document request ("Doc" filter).
8. Open its **Response** tab and search for: `WP Fastest Cache file was created`
9. Record the cache creation timestamp.
10. Filter Network requests for `admin-ajax.php` with action `pinterest_for_woocommerce_page_visit`
11. Record the conversion ID. 

#### Validate a cache HIT
1. Close the incognito window and open a new one. 
2. Open Chrome DevTools → Network.
3. Enable **Preserve log** and leave **Disable cache** unchecked.
4. Visit a product page.
5. Select the document request ("Doc" filter).
6. Open its **Response** tab and search for: `WP Fastest Cache file was created`
7. Ensure that it is the same timestamp as before. 
8. Filter Network requests for `admin-ajax.php` with action `pinterest_for_woocommerce_page_visit`
9. Ensure that a new conversion ID was created. 

### Changelog entry

> Fix - Fix PageVisit tracking on full-page cache hits

### DIFF
--- a/src/Tracking/Conversions.php
+++ b/src/Tracking/Conversions.php
@@ -31,16 +31,29 @@ class Conversions extends Tracker {
 	 */
 	private const CLICK_ID_SESSION_KEY = 'pinterest_for_woocommerce_click_id';
 
-	/** @var User $user User data object. Data for Conversions API. */
+	/**
+	 * User data for the Conversions API.
+	 *
+	 * @var User
+	 */
 	private $user;
+
+	/**
+	 * URL where the event occurred.
+	 *
+	 * @var string
+	 */
+	private $event_source_url;
 
 	/**
 	 * Pinterest Conversions API class constructor.
 	 *
-	 * @param User $user User data object to hold ip address and agent string.
+	 * @param User   $user             User data object to hold ip address and agent string.
+	 * @param string $event_source_url Optional URL where the event occurred.
 	 */
-	public function __construct( User $user ) {
-		$this->user = $user;
+	public function __construct( User $user, string $event_source_url = '' ) {
+		$this->user             = $user;
+		$this->event_source_url = $event_source_url;
 	}
 
 	/**
@@ -115,11 +128,13 @@ class Conversions extends Tracker {
 	private function get_default_data( string $event_name ) {
 		global $wp;
 
+		$event_source_url = $this->event_source_url ? $this->event_source_url : home_url( $wp->request );
+
 		$data = array(
 			'event_name'       => $event_name,
 			'action_source'    => 'web',
 			'event_time'       => time(),
-			'event_source_url' => home_url( $wp->request ),
+			'event_source_url' => $event_source_url,
 			'partner_name'     => 'ss-woocommerce',
 			'user_data'        => array(
 				'client_ip_address' => $this->user->get_client_ip_address(),
--- a/src/Tracking/Conversions.php
+++ b/src/Tracking/Conversions.php
@@ -153,7 +153,7 @@ class Conversions extends Tracker {
 			$data['user_data']['external_id'] = array( $external_id );
 		}
 
-		$click_id = self::maybe_get_click_id();
+		$click_id = self::maybe_get_click_id( $this->event_source_url );
 		if ( false !== $click_id ) {
 			$data['user_data']['click_id'] = $click_id;
 		}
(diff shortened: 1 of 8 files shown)


src/Tracking/Conversions.php
```diff
@@ -108,11 +121,13 @@ public function prepare_request_data( string $event_name, Data $data ) {
 	private function get_default_data( string $event_name ) {
 		global $wp;
 
+		$event_source_url = $this->event_source_url ? $this->event_source_url : home_url( $wp->request );
```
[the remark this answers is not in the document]
jimjasson: Resolved: https://github.com/woocommerce/pinterest-for-woocommerce/pull/1207/commits/8f4202a66abda44bc339cbf0ae8119c65424e992

src/Tracking/PageVisit.php
```diff
@@ -0,0 +1,183 @@
...
+			return;
+		}
+
+		$source_url = isset( $_POST['event_source_url'] ) ? esc_url_raw( wp_unslash( $_POST['event_source_url'] ) ) : ''; // phpcs:ignore WordPress.Security.NonceVerification.Missing
+		$source_url = static::validate_source_url( $source_url );
+		if ( ! $source_url ) {
+			return;
+		}
+
+		$product_id = isset( $_POST['product_id'] ) ? absint( $_POST['product_id'] ) : 0; // phpcs:ignore WordPress.Security.NonceVerification.Missing
```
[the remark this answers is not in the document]
jimjasson: Resolved: https://github.com/woocommerce/pinterest-for-woocommerce/pull/1207/commits/8f4202a66abda44bc339cbf0ae8119c65424e992

src/Tracking/PageVisit.php
```diff
@@ -0,0 +1,191 @@
...
+		$source_url = esc_url_raw( wp_unslash( $source_url_raw ) );
+		$source_url = static::validate_source_url( $source_url );
+		if ( ! $source_url ) {
+			return;
+		}
+
+		$product_id = absint( wp_unslash( $product_id_raw ) );
+		$data       = static::get_event_data( $event_id, $product_id );
+		$user       = new User( \WC_Geolocation::get_ip_address(), wc_get_user_agent() );
+		$tracker    = new Conversions( $user, $source_url );
```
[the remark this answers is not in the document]
jimjasson: Resolved: https://github.com/woocommerce/pinterest-for-woocommerce/pull/1207/commits/2a914cb3deb45b1bbed93c2bd9f48a9a8f2df0b5

src/Tracking/PageVisit.php
```diff
@@ -0,0 +1,188 @@
...
+	public static function get_tag_event_code( array $data ) {
+		unset( $data['event_id'] );
+
+		$event_data = wp_json_encode( (object) $data, JSON_HEX_TAG | JSON_UNESCAPED_SLASHES );
+		$event_data = $event_data ? $event_data : '{}';
+		$capi_code  = '';
+
+		if ( Pinterest_For_Woocommerce()::get_setting( 'track_conversions_capi' ) ) {
+			$capi_code = sprintf(
+				'var requestData=new FormData();requestData.append("action",%1$s);requestData.append("event_id",eventId);requestData.append("event_source_url",window.location.href);var beaconSent=navigator.sendBeacon&&navigator.sendBeacon(%2$s,requestData);if(!beaconSent&&window.fetch){window.fetch(%2$s,{method:"POST",body:requestData,credentials:"same-origin",keepalive:true});}',
```
layoutd, line 64: How will this preserve the `epik` click ID added in #1203?

This beacon posts to `admin-ajax.php`, so the landing page’s `?epik=...` value will not be in the AJAX request’s `$_GET`. The original URL is sent as `event_source_url`, but the server does not currently read `epik` from it.

In my manual test, I’d expect a first visit to `/product/example/?epik=test-click-id` to include that value as `user_data.click_id` in the PageVisit CAPI event. Could we cover that combined flow before merging?

layoutd: I tested this with #1203 in a fresh browser session using `?epik=finaltestvalue`. The conversion log preserved it in `event_source_url`, but `user_data.click_id` was missing:
```
event_source_url: .../product/beanie/?epik=finaltestvalue

user_data: { client_ip_address, client_user_agent }
```
Since the beacon request goes to `admin-ajax.php`, #1203 apparently cannot read the landing page’s epik from `$_GET`, and neither the cookie nor session supplied it in this test.

Could we make sure the beacon preserves this value as `user_data.click_id` and add a regression test for the combined flow?

layoutd (changes requested):
Thanks @jimjasson, I tested both cache misses and cache hits and confirmed that each visit gets a fresh event ID, while the Tag and CAPI events use the same ID. The overall approach makes sense for full-page caches, and rebuilding product data from the page URL is a good safeguard.

I found one issue when testing this together with #1203. On a fresh visit with `?epik=finaltestvalue`, the conversion log preserved the value in event_source_url, but `user_data.click_id` was missing. Since the beacon posts to `admin-ajax.php`, #1203 cannot read the landing page’s `epik` from `$_GET`, and neither the cookie nor session supplied it in this test.

Could we preserve the click ID in the beacon flow and add a regression test covering #1203 and #1207 together? Other than this attribution issue, the cache-hit fix worked well in my testing.

src/Tracking/PageVisit.php
jimjasson: Thanks for the feedback @layoutd! 

Looks like you have a clear resolution in mind. Could you please go ahead and implement the last piece needed here? Thank you!

layoutd: Done, will test and approve this afternoon.

layoutd (approved):
Retested with #1203 included. Cache hit visit sends a fresh browser-generated event_id via the beacon, and a landing visit with `?epik=chromevalue1` now carries `user_data.click_id` in the PageVisit CAPI event. Invalid event IDs and off-site source URLs are dropped. Approving and merging.

PR #1208, opened 2026-08-25
Fix AEM configuration for auto-created tags

### LINKED ISSUE #1206: Fix tags being created with missing Automatic Enhanced Match (AEM) config
### Describe the bug:

When the WooCommerce setup wizard auto-creates a Pinterest tag, Automatic Enhanced Match (AEM) is enabled for every parameter except `external_id`, which was added most recently for AEM. This is caused by the tag creation request omitting the `aem_external_id_enabled` field, so external ID stays off for every merchant who onboards through the plugin. External ID is one of the highest-value match signals, so this measurably reduces conversion match rates for WooCommerce merchants.

### Steps to reproduce:
1. Follow this doc to set up tag via WooCommerce: https://docs.google.com/document/d/1PrWM7GZrhcuNUkqGHvY5lE_DRPAXOQ06ousCE4_SkC0/edit?tab=t.0#heading=h.tw2u2lvsqc6g
2. Completing the flow should automatically create a Pinterest tag.
3. Wait a day for Pinterest tag to show up in Ads Manager > Conversions > [Tag manager page](http://ads.pinterest.com/conversions/tag).
4. Open the tag's Enhanced match settings and check the Automatic Enhanced Match parameters; every parameter is ON except External ID, which is OFF.

Image
Image

### Expected behavior:
All AEM params, including external_id, should be turned ON by default.

### Actual behavior:
All AEM params, except external_id, are turned ON.

### Proposed solution:

In `APIV5::create_tag()` (`src/API/APIV5.php`), the `POST ad_accounts/{ad_account_id}/conversion_tags` payload lists each AEM parameter explicitly but has no entry for external ID. The v5 create-conversion-tag endpoint accepts `aem_external_id_enabled`, so adding it should be enough:

  ```php
  array(
      'name'                    => $tag_name,
      'aem_enabled'             => true,
      'md_frequency'            => 1,
      'aem_fnln_enabled'        => true,
      'aem_ph_enabled'          => true,
      'aem_ge_enabled'          => true,
      'aem_db_enabled'          => true,
      'aem_loc_enabled'         => true,
      'aem_external_id_enabled' => true, // added
  )
```

Separately, the current payload sends ae_loc_enabled, but the documented field name is aem_loc_enabled; the ae_loc_enabled key is not recognized by the endpoint and is dropped, so location AEM is not actually being set by the plugin today (it appears enabled only because of the server-side default). Worth fixing in the same change; the snippet above includes the corrected key.

Code pointer: https://github.com/woocommerce/pinterest-for-woocommerce/blob/465f5085fb1e80e1557476c19f9d99e87c5155c6/src/API/APIV5.php#L266-L282

### PR DESCRIPTION
**Pull request by jimjasson**

Closes #1206 

Fix the Automatic Enhanced Match configuration sent when Pinterest for WooCommerce automatically creates a conversion tag.

This PR:

- Adds `aem_external_id_enabled` to enable External ID matching.
- Corrects `ae_loc_enabled` to the documented `aem_loc_enabled` field.
- Updates the conversion-tag response documentation.
- Adds regression coverage that verifies the outgoing tag-creation request and the REST onboarding flow.

1. On a new store, install Pinterest for Woo from this branch.
2. Connect Pinterest and select to create a new account/tag combination, if you already have an account.
3. Once connection is completed and a new tag is created, go to the Tag Manager at: `https://ads.pinterest.com/advertiser/<ADS_ID>/conversions/tag/`.
4. Edit the tag and confirm that all the options are enabled.

### Changelog entry

>  Fix - Fixed AEM configuration for auto-created tags

### DIFF
--- a/src/API/APIV5.php
+++ b/src/API/APIV5.php
@@ -178,6 +178,7 @@ class APIV5 extends Base {
 	 *              @type ?bool $aem_ge_enabled     Whether Automatic Enhanced Match gender is enabled.
 	 *              @type ?bool $aem_db_enabled     Whether Automatic Enhanced Match birthdate is enabled.
 	 *              @type ?bool $aem_loc_enabled    Whether Automatic Enhanced Match location is enabled.
+	 *              @type ?bool $aem_external_id_enabled Whether Automatic Enhanced Match external ID is enabled.
 	 *          }
 	 *      }
 	 * }
@@ -221,6 +222,7 @@ class APIV5 extends Base {
 	 *          @type bool    $aem_ge_enabled      Whether Automatic Enhanced Match gender is enabled.
 	 *          @type bool    $aem_db_enabled      Whether Automatic Enhanced Match birthdate is enabled.
 	 *          @type bool    $aem_loc_enabled     Whether Automatic Enhanced Match location is enabled.
+	 *          @type bool    $aem_external_id_enabled Whether Automatic Enhanced Match external ID is enabled.
 	 *      }
 	 * }
 	 * @throws PinterestApiException Throws 500 exception in case of unexpected error.
@@ -259,6 +261,7 @@ class APIV5 extends Base {
 	 *          @type ?bool $aem_ge_enabled     Whether Automatic Enhanced Match gender is enabled.
 	 *          @type ?bool $aem_db_enabled     Whether Automatic Enhanced Match birthdate is enabled.
 	 *          @type ?bool $aem_loc_enabled    Whether Automatic Enhanced Match location is enabled.
+	 *          @type ?bool $aem_external_id_enabled Whether Automatic Enhanced Match external ID is enabled.
 	 *      }
 	 * }
 	 * @throws PinterestApiException|Exception Throws 500 exception.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by sangwo — APPROVED** [review]
lgtm, thank you for the quick fix!

**Review by layoutd — COMMENTED** [review]
Thanks, this looks good, but I'm waiting on this step for final confirmation:

> Wait a day for Pinterest tag to show up in Ads Manager > Conversions > [Tag manager page](http://ads.pinterest.com/conversions/tag).

**Comment by jimjasson** [comment]
@layoutd Any updates on this? Did it show up in Ads Manager > Conversions > [Tag manager page](http://ads.pinterest.com/conversions/tag)?

**Comment by layoutd** [comment]
(Was stuck in draft) @jimjasson Yes, it's there. Location (City, State, Zip Code and Country) is enabled on it. External ID came back false though. I pulled the tag via the API to make sure it wasn't just the UI lagging, and `configs.aem_external_id_enabled` is false there too.

One caveat on all that: I had debug logging off when the tag was actually created, so I don't have the create request and response captured, only the state afterwards. Going to do one clean run with logging on from the start so I can see the payload and the resulting configs side by side. Will confirm here once I have it.

Worth noting either way that v5 only exposes GET and POST on conversion_tags, no PATCH or PUT, so create is our only chance to set the AEM config. Happy for this to go in as-is regardless. The location fix alone is worth it, and the External ID param is a harmless no-op if Pinterest carries on ignoring it.

**Comment by layoutd** [comment]
image

Clean run done with debug logging on from the start. External ID now comes back enabled, so my earlier result was from a tag created before the fix. Disregard that.

Create request body as logged:
```json
{"name":"Auto-created by Pinterest for WooCommerce","aem_enabled":true,"md_frequency":1,"aem_fnln_enabled":true,"aem_ph_enabled":true,"aem_ge_enabled":true,"aem_db_enabled":true,"aem_loc_enabled":true,"aem_external_id_enabled":true}
```
Response (200) configs on the new tag:
```
aem_enabled: true
md_frequency: 1
aem_fnln_enabled: true
aem_ph_enabled: true
aem_ge_enabled: true
aem_db_enabled: true
aem_loc_enabled: true
aem_external_id_enabled: true
```

Tag Manager shows the same: all AEM toggles on, including External ID. Approving and merging.

PR #1209, opened 2026-08-25
Fix stale feed cleanup deleting manual catalog feeds on the store domain

### LINKED ISSUE #1205: Stale-feed cleanup deletes manual catalog feeds hosted on the WooCommerce site domain
### Describe the bug:

The stale-feed cleanup can delete manually configured Pinterest catalog feeds that are hosted on the same domain as the WooCommerce store.

After selecting or creating the current WooCommerce feed, `maybe_delete_stale_feeds_for_merchant()` iterates over every other feed. It classifies a feed as WooCommerce-owned when its location begins with `get_site_url()` and sends a catalog feed `DELETE` request.

The site URL is not a feed-ownership boundary. Merchants can host manual CSV, TSV, or XML feeds elsewhere on the same domain. Those feeds satisfy the current condition even though this plugin did not create them.

Relevant code:

[Recurring feed registration](<https://github.com/woocommerce/pinterest-for-woocommerce/blob/0f293a6d266d9a2d340905c90c542d05f16dff78/src/FeedRegistration.php#L61-L65>)

[Cleanup invocation after feed selection or creation](<https://github.com/woocommerce/pinterest-for-woocommerce/blob/0f293a6d266d9a2d340905c90c542d05f16dff78/src/FeedRegistration.php#L145-L160>)

[Same-domain classification and deletion](<https://github.com/woocommerce/pinterest-for-woocommerce/blob/0f293a6d266d9a2d340905c90c542d05f16dff78/src/FeedRegistration.php#L191-L216>)

[DELETE request sent to the Pinterest Catalogs API](<https://github.com/woocommerce/pinterest-for-woocommerce/blob/0f293a6d266d9a2d340905c90c542d05f16dff78/src/API/APIV5.php#L540-L544>)

### Steps to reproduce:

1. Connect Pinterest for WooCommerce and let it create its feed, for example `https://store.example.com/wp-content/uploads/pinterest-for-woocommerce-abc123.xml`.
2. Add a separate manual Pinterest catalog feed hosted on the same site, for example `https://store.example.com/catalogs/manual.csv`.
3. Let the recurring feed-registration action run, or invoke the registration handler.
4. List the merchant's catalog feeds.

### Expected behavior:

The plugin preserves catalog feeds it did not create. Stale-feed cleanup deletes only stale feeds that are demonstrably owned by Pinterest for WooCommerce.

### Actual behavior:

The current WooCommerce feed remains, but the same-domain manual feed receives a catalog feed DELETE request. On Pinterest, the manual feed becomes disabled/deleted.

The disconnect path invokes the same cleanup with an empty preserved feed ID, so it can also delete all same-domain feeds:

* [Disconnect cleanup](<https://github.com/woocommerce/pinterest-for-woocommerce/blob/0f293a6d266d9a2d340905c90c542d05f16dff78/class-pinterest-for-woocommerce.php#L829-L833>)

### Additional details:

recognize the plugin-owned path pattern:

/pinterest-for-woocommerce-.xml

so have a check for this pattern?

PanosSynetos:

Closes #1205.
Closes PIN4WOO-98

`FeedRegistration::maybe_delete_stale_feeds_for_merchant()` treated every catalog feed whose `location` starts with `get_site_url()` as plugin-owned and sent a catalog feed `DELETE` for it:

```php
// Delete the feed if it belongs to the current website.
$is_woocommerce_feed = 0 === strpos( $feed['location'], get_site_url() );
```

The site URL is not an ownership boundary. This plugin only ever generates one shape of feed file — `{uploads_baseurl}/pinterest-for-woocommerce-{id}.xml`, written by `LocalFeedConfigs` — so any data source the merchant configured by hand and happened to host on their own domain (a CSV in `/wp-content/uploads/catalogs/`, a feed served by another plugin, a second install in a subdirectory) matched the check and was deleted from Pinterest.

**Changes**

- `Feeds::is_plugin_generated_feed_location()` — new public helper extracted from the ownership half of `does_feed_match()`. Returns true only for locations under the uploads base URL carrying the `pinterest-for-woocommerce-` prefix, scheme-normalised through the existing `normalize_feed_location_url()`. `does_feed_match()` now delegates to it, so both call sites share one definition of ownership.
- `maybe_delete_stale_feeds_for_merchant()` uses that helper instead of the site-URL prefix, and no longer assumes `id` / `location` are present in the API payload.

The country and locale equality checks at the top of `does_feed_match()` are deliberately **not** part of the helper: stale plugin feeds are usually stale precisely because the merchant's country or locale changed, so folding those in would leave real stale feeds behind.

This also closes the three collateral paths that ran the same faulty check with an empty preserved feed ID — disconnect (`class-pinterest-for-woocommerce.php:831`), plugin deactivation (`pinterest-for-woocommerce.php:141-145`) and uninstall (`uninstall.php:17`). Deactivating the plugin was enough to delete a merchant's manual feeds while the access token was still valid.

**Automated**

1. `./bin/install-wp-tests.sh wordpress_test root root 127.0.0.1` (once).
2. `vendor/bin/phpunit --filter FeedRegistrationTest`

Five tests in `tests/Unit/FeedRegistrationTest.php` fake `GET catalogs/feeds` and record every `DELETE catalogs/feeds/{id}`:

| Test | Covers |
| --- | --- |
| `test_stale_feed_cleanup_keeps_manual_feeds_hosted_on_the_site_domain` | Registration path — the stale plugin feed is deleted, the manual same-domain feed is not. |
| `test_disconnect_cleanup_keeps_manual_feeds_hosted_on_the_site_domain` | Disconnect path, cleanup called with an empty preserved feed ID. |
| `test_plugin_deactivation_keeps_manual_feeds_hosted_on_the_site_domain` | Drives the real `pinterest_for_woocommerce_deregister()` entry point with a connected business account. |
| `test_stale_feed_cleanup_does_not_match_on_a_partial_site_url_prefix` | A sibling domain (`shop.testing.com` vs a site URL of `shop.test`) and a subdirectory install no longer match. |
| `test_stale_feed_cleanup_skips_feeds_without_an_id` | A feed entry with a missing or empty `id` is skipped instead of triggering a `DELETE` with an empty ID. |

They were added red in the first two commits of this branch ([CI run](https://github.com/woocommerce/pinterest-for-woocommerce/actions/runs/32861223216) — `Tests: 291, Failures: 2` before the deactivation and prefix cases landed) and pass with the fix. Full suite: 293 tests, 607 assertions, green. `vendor/bin/phpcs` clean on the changed files.

**Manual, against the live Pinterest API**

1. On a store with a public URL, connect the plugin and let it register its feed (`{uploads}/pinterest-for-woocommerce-XXXXXX.xml`).
2. Drop a hand-written CSV catalog at `wp-content/uploads/catalogs/manual-feed.csv` on the same domain, and add it in Pinterest under **Catalogs → Data sources → Add a data source**.
3. Confirm both are registered:

```bash
wp eval 'Automattic\WooCommerce\Pinterest\Feeds::invalidate_feeds_cache();
foreach ( Automattic\WooCommerce\Pinterest\Feeds::get_feeds() as $f ) {
  printf( "%s | %s | %s\n", $f["id"], $f["format"] ?? "-", $f["location"] ?? "-" );
}'
```

4. Trigger the cleanup (or wait for the recurring `pinterest-for-woocommerce-handle-feed-registration` action):

```bash
wp eval 'Automattic\WooCommerce\Pinterest\FeedRegistration::maybe_delete_stale_feeds_for_merchant(
  Pinterest_For_Woocommerce::get_data( "feed_registered" )
);'
```

5. Re-run step 3. **Before this fix** the manual CSV was gone from Pinterest — reproduced end to end against the production API, the manual data source deleted while the plugin's own feed was kept. **With this fix** both feeds are still listed.
6. Copy the plugin's feed file to `{uploads}/pinterest-for-woocommerce-STALE1.xml`, register it as a second data source, and repeat steps 4–5: it *is* deleted, while the manual CSV survives. This proves the fix narrows ownership rather than disabling cleanup.

> Note: `Feeds::get_feeds()` caches for 60 seconds and the Pinterest UI lags behind the API, so call `invalidate_feeds_cache()` before re-reading rather than trusting the dashboard.

### Additional details:

The scheme normalisation inherited from `normalize_feed_location_url()` also covers stores whose registered feed URL was stored over `http` and later moved to `https`, which the old bare `strpos()` on `get_site_url()` silently skipped.

### Changelog entry

> Fix - Stale feed cleanup no longer deletes manually configured Pinterest catalog data sources hosted on the store domain.


src/FeedRegistration.php
```diff
@@ -198,15 +202,21 @@ public static function maybe_delete_stale_feeds_for_merchant( string $feed_id )
...
-			$is_woocommerce_feed = 0 === strpos( $feed['location'], get_site_url() );
+			/*
+			 * Delete the feed only if this plugin generated the feed file it points to.
+			 * Merchants may host their own data sources on the store domain, so the site
+			 * URL alone does not establish feed ownership.
+			 */
+			$is_woocommerce_feed = Feeds::is_plugin_generated_feed_location( $feed['location'] ?? '' );
 			if ( $is_woocommerce_feed ) {
-				Feeds::delete_feed( $feed['id'] );
+				Feeds::delete_feed( $registered_feed_id );
```
layoutd, line 219: What happens if Pinterest returns a feed with a location but no ID?

For example:

```php
array(
    'location' => 'https://example.com/wp-content/uploads/pinterest-for-woocommerce-Ab1cD2.xml',
)
```

`is_plugin_generated_feed_location()` would return true, but we would then call `delete_feed( '' )`.

Could we defensively skip entries whose ID is missing or empty, and add a test for this case?

src/Feeds.php
```diff
@@ -354,10 +354,30 @@ private static function does_feed_match( array $feed ): bool {
...
+	public static function is_plugin_generated_feed_location( string $location ): bool {
+		if ( '' === $location ) {
+			return false;
+		}
+
 		$upload_base_url = self::normalize_feed_location_url( wp_get_upload_dir()['baseurl'] );
 		$feed_location   = trailingslashit( $upload_base_url ) . PINTEREST_FOR_WOOCOMMERCE_LOG_PREFIX . '-';
 
-		return 0 === strpos( self::normalize_feed_location_url( $feed['location'] ?? '' ), $feed_location );
+		return 0 === strpos( self::normalize_feed_location_url( $location ), $feed_location );
```
layoutd, line 385: Could a merchant manually create a file whose name starts with the same prefix?

For example: `/wp-content/uploads/pinterest-for-woocommerce-manual.csv`

This check would treat it as plugin-generated. I do not think this needs to block the current fix, since URL/file naming can only be a heuristic when recovering historical or orphaned feeds, but it may be worth documenting as a limitation or considering a more robust ownership determination strategy, maybe based on the Pinterest feed IDs created by the plugin.

layoutd (approved):
🚢  Thanks @PanosSynetos! This looks like the right fix. The old same-domain check was clearly too broad, and sharing the ownership check with the existing feed-matching code is a good improvement. The regression coverage is helpful too.

I left one small defensive comment about skipping a feed with a missing ID, so we do not issue a malformed delete request and create unnecessary log noise.

I also considered the remaining edge case of a manually created feed deliberately using the plugin filename prefix. I do not think that should block this fix. For older or orphaned feeds, the URL and filename are only our best guess. Longer term, the safest option would be to store the Pinterest feed IDs created by the plugin and use those to identify feeds it owns.

Overall, this is focused, safer than the existing behavior, and looks good to merge.

src/FeedRegistration.php
layoutd: Good catch. Entries with a missing or empty `id` are now skipped before the ownership check in 01beae51, and `test_stale_feed_cleanup_skips_feeds_without_an_id` covers both the missing key and the empty string. The fake API in the test now records DELETE requests with an empty ID too, so the test fails against the previous code.

src/Feeds.php
layoutd: Agreed that this stays a heuristic for now. Documented it as a known limitation on `is_plugin_generated_feed_location()` in e1b0ccf9, including the suggestion to record the feed IDs the plugin creates as the longer-term ownership check.

PR #1193, opened 2026-07-20
Fix keyboard accessibility of the Pinterest Save button on product images

mukeshpanchal27:
### Changes proposed in this Pull Request:

Closes https://linear.app/a8c/issue/WOOA11Y-256/a11y-element-is-not-keyboard-operable-in-div-with-visibilityhidden, https://linear.app/a8c/issue/WOOA11Y-242/a11y-element-is-not-keyboard-operable-span-instead-of-button-wooext

The Pinterest Save button sits in `.pinterest-for-woocommerce-image-wrapper`, which used `visibility: hidden` until hover. That kept the link out of keyboard focus and made it unusable without a mouse (and harder for screen reader users).

This change:
- Replaces `visibility: hidden` / `visible` with `pointer-events: none` / `auto` so the control stays in the accessibility/focus tree while still avoiding accidental clicks when visually hidden.
- Adds `:focus-within` alongside `:hover` so the Save button becomes visible and interactive when reached via keyboard.

Touch devices without hover are unchanged (`@media (hover: none)` still shows the button).

> [!NOTE]
> https://resident-tern-saxophone.jurassic.ninja/wp-admin
> demo / 3u7ef3oCX0ZpDYrcRCN7

#### Shop page
https://github.com/user-attachments/assets/61bc7328-685c-4e96-83a6-22e3d264af2a

#### Product page
https://github.com/user-attachments/assets/5ffeacd9-bec6-4714-965d-9b5a2c7b1814

### Detailed test instructions:

1. Pull the latest changes for this branch.
2. Build assets: `npm run build`.
3. Open a product page, e.g. https://resident-tern-saxophone.jurassic.ninja/product/blouse/
4. Use `Tab` to move focus until the `Save` button is selected.
5. Confirm the Save button:
   - Can receive keyboard focus
   - Becomes visible when focused (via `:focus-within`)
6. Open a Shop page, e.g. https://resident-tern-saxophone.jurassic.ninja/shop/
7. Confirm point number 4 and 5

### Changelog entry

> Fix - Make the Pinterest Save button reachable and usable with keyboard navigation by removing `visibility: hidden` from the image wrapper on shop and product page.

PR #1211, opened 2026-08-27
Fix keyboard accessibility for Pinterest Save button on products with no featured image

mukeshpanchal27:

Closes https://linear.app/a8c/issue/WOOA11Y-242/a11y-element-is-not-keyboard-operable-span-instead-of-button-wooext

Follow-up to #1193.

When a product has no featured image, Pinterest’s Save widget does not upgrade our `<a data-pin-do="buttonPin">` to a pin link with `data-pin-log="button_pinit"`. It renders a bookmarklet control (`span` with `data-pin-log="button_pinit_bookmarklet"`).

The Save button a11y script only handled the `<a>` / `button_pinit` case, so on those products the control was not keyboard-focusable, the screen-reader label was not attached, and Enter did not open the Save dialog.

This change:
- Treats `span[data-pin-log]` as the Save control when no `<a>` is present
- Accepts `data-pin-log="button_pinit_bookmarklet"` as a ready state
- Adds `role="button"`, `tabindex="0"`, and `aria-haspopup="dialog"` on that span
- Activates the control on Enter/Spacebar (Pinterest only listens for click)

> [!NOTE]
> https://panda-of-aphids.jurassic.ninja/wp-admin
> demo / hDrIgiroQjRy2LWJTBCH

### Screen Recording:

https://github.com/user-attachments/assets/47caf7b1-6a89-4dce-b14e-0252adacb09d

1. Open a shop/product page with with no featured image.
2. Tab to a product card; confirm the Save button becomes visible on keyboard focus.
3. Tab until the Save control is focused. 
4. Press Enter/Spacebar. Confirm the Pinterest Save dialog/window opens (same as a mouse click).

### Changelog entry

> Fix - Make the Pinterest Save button keyboard-accessible on products with no featured image.


petenelson (approved):

qasumitbagthariya:
**QA Update: QA Passed. ✅**
---


* Verified the Save button becomes visible when the product card receives keyboard focus.
* Confirmed the .screen-reader-text span is correctly placed inside the Pinterest pin <span>.
* Verified the Save control is visible and announced correctly with the product name and Pinterest action.
* Confirmed pressing Enter opens the Pinterest Save dialog/window as expected, matching mouse-click behavior.

Everything is working as expected. 


🖥 **Testing Environment** 

<Details>

- WordPress: 7.0.4
- Theme: Twenty Twenty-Five 1.4
- WooCommerce - 11.0.1
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome
- OS: macOS 15.3.2
- Branch: fix/WOOA11Y-242

</Details>

qasumitbagthariya (approved):

amberhinds:
Hello, I tested this on the demo above and there are some issues with the implementation.

1. This is adding "dialog popup" to the links when there is a featured image set.
Screenshot 2026-09-01 at 4 53 37 PM

`aria-haspopup` should not be on the element in this context.

2. When a featured image is not set it still has link semantics and a new window warning.
Screenshot 2026-09-01 at 4 54 40 PM

In this context:
- The element should be a `<button>` or have `role="button"`. If using the `role` attribute to make it a button JS must also be added allowing it to be clicked with the spacebar. [See MDN button role docs.](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/button_role)
- "(opens in a new window)" should not be present.

mukeshpanchal27:
@amberhinds Thanks for the review! The issue has been addressed, and the PR is ready for your review. Could you please take another look when you have a moment?

Demo site - https://panda-of-aphids.jurassic.ninja/shop/

amberhinds:
Hi @mukeshpanchal27, Thank you! The version on  https://panda-of-aphids.jurassic.ninja/shop/ fixes this issue.

mukeshpanchal27:
@layoutd over to you for review.

assets/source/js/pinterest-for-woocommerce-save-button.js
```diff
@@ -32,20 +52,31 @@ document.addEventListener( 'DOMContentLoaded', function () {
...
+							'button_pinit_bookmarklet' );
 
 				if ( isPinterestReady && srSpan ) {
-					// Move the span inside the processed <a> tag.
-					pinLink.appendChild( srSpan );
-
-					// Add aria-haspopup to <a> tag.
-					pinLink.setAttribute( 'aria-haspopup', 'dialog' );
+					if ( pinLink.tagName.toLowerCase() === 'span' ) {
+						pinLink.setAttribute( 'aria-haspopup', 'dialog' );
```
layoutd, line 86: This moved into the `span` branch, so the `<a>` case no longer gets `aria-haspopup="dialog"`. Verified on a product with a featured image: `null` on this branch, `"dialog"` on `develop`. Setting it before the if keeps both paths covered.

```suggestion
					pinLink.setAttribute( 'aria-haspopup', 'dialog' );
					if ( pinLink.tagName.toLowerCase() === 'span' ) {
```

assets/source/js/pinterest-for-woocommerce-save-button.js
```diff
@@ -32,20 +52,31 @@ document.addEventListener( 'DOMContentLoaded', function () {
...
+						pinLink.setAttribute( 'role', 'button' );
+						pinLink.setAttribute( 'tabindex', '0' );
+
+						// Remove the screen reader text from wrapper tag.
+						wrapper.removeChild( srSpan );
```
layoutd: Removing the screen reader text leaves the span announced as just "Save", so in a grid every product reads the same. Could we carry the name over as `aria-label` before removing it? The "(opens in a new window)" suffix does not apply here since the bookmarklet opens an in-page dialog, so something like:

```suggestion
						pinLink.setAttribute( 'aria-label', srSpan.textContent.replace( /\s*\(opens in a new window\)\s*$/, '' ) );
						wrapper.removeChild( srSpan );
```

Not a blocker on its own, but worth doing while we are here.

layoutd (changes requested):
Tested at 834ff0c on TT25 with a product that has no featured image, plus a product with one for comparison.

The bookmarklet path works as described: once pinit.js swaps in `span[data-pin-log="button_pinit_bookmarklet"]` the control gets `role="button"`, `tabindex="0"` and `aria-haspopup="dialog"`, and both Enter and Space open the Pinterest "Choose a Pin to save" overlay without scrolling the page.

Requesting changes for the `aria-haspopup` regression on the `<a>` path (inline comment). The accessible name on the span is a smaller follow-up, also inline.

Note for testers: the script defers processing through `requestAnimationFrame`, so nothing gets labeled while the tab is in the background. Keep the tab visible while Pinterest renders.

assets/source/js/pinterest-for-woocommerce-save-button.js
mukeshpanchal27: @layoutd Check amber's first feedback her https://github.com/woocommerce/pinterest-for-woocommerce/pull/1211#issuecomment-5500993042. She suggest to remove the `aria-haspopup` from the `<span>`

assets/source/js/pinterest-for-woocommerce-save-button.js
mukeshpanchal27: @layoutd Does this works in other languages?

mukeshpanchal27: @layoutd In https://github.com/woocommerce/pinterest-for-woocommerce/pull/1211/commits/5177aa203ef63c42d01c0b6d2661519b05827a5f address the issue.

assets/source/js/pinterest-for-woocommerce-save-button.js
layoutd: Thanks for pointing that out, you and Amber are correct.

layoutd (approved):
Tested this out and it's looking good. Both paths behave as called for: the bookmarklet span gets `role="button"`, `tabindex`, and `aria-haspopup="dialog",` and Enter/Space opens the in-page Pinterest grid with no `window.open`.

One nit that doesn't block approval: the bookmarklet path still has no test coverage, the only test change here is the flipped `aria-haspopup` assertion. I may add a follow-up with a `markAsBookmarklet()` case for the role/tabindex/hint-removal plus a keydown test.

layoutd:
I'm going to go ahead and merge this one, and the followup testing PR too

mukeshpanchal27:
@layoutd do you know what version number of the plugin this might be released in?

layoutd:
Almost certainly 1.5.0, this week.

PR #1216, opened 2026-09-09
Tidy up OAuth callback state handling

layoutd:
Closes PIN4WOO-103

Small robustness pass on the OAuth callback's state check. The comparison is now type-consistent, the callback behaves predictably when no connect flow is in progress, and the one-time state is cleared once a callback has used it so it is not accepted again. No change to the normal connect experience.

1. Run `vendor/bin/phpunit --filter AuthTest` and confirm it passes.
2. Start a Pinterest connect flow and confirm it completes and lands on the settings page as before.
3. Reconnect after disconnecting and confirm that also works.

### Changelog entry

> Tweak - Tidy up the Pinterest OAuth callback state handling.

### DIFF
--- a/src/API/Auth.php
+++ b/src/API/Auth.php
@@ -46,14 +46,20 @@ class Auth extends VendorAPI {
 	 * @return boolean
 	 */
 	public function permissions_check( WP_REST_Request $request ) {
-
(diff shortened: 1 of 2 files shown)


louwie17 (approved):
Thanks for this, tested it and code looks good 👍

PR #1217, opened 2026-09-09
Redirect OAuth callback rejections to the settings page

layoutd:

Restores the settings-page redirect on a rejected OAuth callback. The `rest_pre_serve_request` hook that produced it was dropped in the 2023 auth refactor, so rejected requests returned the raw REST error (`rest_forbidden` JSON) instead of sending the merchant back to onboarding with a message. This reinstates the redirect for both rejection paths. Follow-up to the OAuth callback state-handling PR that this one is based on.

1. Run `vendor/bin/phpunit --filter AuthTest` and confirm it passes.
2. Visit `/wp-json/pinterest/v1/oauth/callback` with no valid state in a browser. Confirm you land on the Pinterest onboarding screen with an error notice, not a raw JSON error.
3. Complete a normal connect flow and confirm it still succeeds.

### Changelog entry

> Fix - Send failed Pinterest OAuth callbacks back to the settings page instead of showing a raw REST error.

### DIFF
--- a/src/API/Auth.php
+++ b/src/API/Auth.php
@@ -52,26 +52,13 @@ class Auth extends VendorAPI {
 		 * Fail closed when no state has been issued, and only accept a string that matches it exactly.
 		 */
 		$expected = get_transient( \PINTEREST_FOR_WOOCOMMERCE_CONNECT_NONCE );
+		$state    = $request->get_param( 'state' );
 
-		if ( ! is_string( $expected ) || '' === $expected ) {
-			return $this->reject();
+		if ( is_string( $expected ) && '' !== $expected && is_string( $state ) && hash_equals( $expected, $state ) ) {
+			return true;
 		}
 
-		$state = $request->get_param( 'state' );
-
-		if ( ! is_string( $state ) || ! hash_equals( $expected, $state ) ) {
-			return $this->reject();
-		}
-
-		return true;
-	}
-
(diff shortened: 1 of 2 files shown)


src/API/Auth.php
```diff
@@ -54,12 +54,27 @@ public function permissions_check( WP_REST_Request $request ) {
...
+
+	/**
+	 * Rejects the request, sending the merchant back to the settings page instead of a raw REST error.
+	 *
+	 * @return bool Always false.
+	 */
+	private function reject(): bool {
+		add_filter( 'rest_pre_serve_request', array( $this, 'redirect_to_settings_page' ), 10, 3 );
+
+		return false;
```
louwie17, line 64: It seems a bit overkill to me to create a new method for this alone, I did rather have the filter inline I think.

louwie17 (commented):
Tested well, just uncertain about creating a brand new function just to call this hook. I did rather keep that inline.

src/API/Auth.php
layoutd: Fair enough, inlined it. Collapsed the two rejection branches into one so the `add_filter` isn't duplicated. Thanks for testing. b8f0ce7c

louwie17 (approved):
LGTM 🚀 Thanks for the update this looks good and tested well!

PR #1220, opened 2026-09-14
Updated Jetpack Device Detection to v3.4.6

jimjasson:

This PR updates the Jetpack Device Detection package to v3.4.6, which recognizes Pinterest's documented crawler user agents as bots: https://github.com/Automattic/jetpack-device-detection/compare/v3.4.5...v3.4.6

Closes PIN4WOO-95

### Changelog entry

> Fix - Detect and bypass conversion events triggered by Pinterest's documented crawler user agents.


composer.lock
```diff
@@ -70,23 +70,23 @@
         },
         {
             "name": "automattic/jetpack-device-detection",
-            "version": "v3.4.2",
+            "version": "v3.4.6",
```
layoutd, line 73: Does it make sense to bump `composer.json` to `^3.4.6` too? As is, `^3.4` still allows 3.4.2, so nothing actually requires the version with the fix.

layoutd (approved):
LGTM 🚀 Good choice to get this fixed upstream! Tested on a local store with CAPI on. Before the bump, a PageVisit beacon with the Pinterestbot UA went out to Pinterest. With 3.4.6 in place, `Pinterestbot/1.0` and `Pinterest/0.2` get skipped, and a regular Chrome UA still goes through.

Nothing blocking, just a few things:

- Could we add the Pinterest UAs to `CrawlerDetectorTest`? The whole fix lives in the lock file, so a bad lock merge could undo it without anything failing. Something like this in `crawler_user_agents()`:

  ```php
  'Pinterestbot' => array( 'Mozilla/5.0 (compatible; Pinterestbot/1.0; +http://www.pinterest.com/bot.html)' ),
  'Pinterest 0.2' => array( 'Pinterest/0.2 (+http://www.pinterest.com/bot.html)' ),
  ```

  A Pinterest in-app browser UA in `human_user_agents()` would be good too, so nobody broadens the match to plain `pinterest` later and drops real shoppers.

- PR description: the lock moves from 3.4.2, not 3.4.5, so the compare link in the description only shows part of it; not particularly important since it's just phpcs comment cleanup.
- More of a question: if another plugin loads an older device-detection through its own Composer autoloader, its copy of the class could win, and Pinterestbot would get through again on that site. I wonder if it's worth a `pinterestbot` fallback in `PROGRAMMATIC_CLIENT_REGEX`, or overkill?

layoutd:
Pushed `eb412b4f` with the two things from my review, hope you don't mind me doing it directly (happy to revert if you'd rather take them yourself):

- `composer.json` now requires `^3.4.6` instead of `^3.4`. As it was, a `composer update` could have pulled 3.4.2 back in and quietly dropped the fix, since it only lived in the lock file. `composer update --lock` refreshed the lock hash; no package versions moved.
- Added `Pinterestbot/1.0` and `Pinterest/0.2` to `crawler_user_agents()`, plus a Pinterest iOS in-app browser UA to `human_user_agents()` so nobody later broadens the match to plain `pinterest` and drops real shoppers.

Full suite green (344 tests), phpcs clean.

I left the `PROGRAMMATIC_CLIENT_REGEX` fallback alone, that's a behavior change to our own detection rather than part of the bump, so it's better off as its own issue if we want it.

